Issue #5, opened 2018-05-23
Suggestion to reorganize the spec


thekaveman:
Below, I suggest a (fairly minor) reorganization of the way this spec is presented. I will follow-up with a PR to introduce these changes, if one is warranted, after there's been a chance for some discussion.

## Currently

From the README:

> Currently, there are two implementations avaliable. Version `v0.1` defines a data specification for providers to implement that the City can query in realtime, while `v0.2` implements an API that providers can query for similar information and additional realtime parking, cost and trip permission information.
>...
> The specification will be versioned using Git tags

Given that this all still very early, the current demarcation into *components* `v0.1` and `v0.2` makes sense. I think what my proposal boils down to is re-labeling these components to provide for a little more clarity and maybe even more flexibility in the future. 

## Suggestion

  - Rename *Version `v0.1`* to the **Provider spec** (or similar)
  - Rename *Version `v0.2`* to the **Agency spec** (or similar)
  - Use Git tags / [semvar](https://semver.org/) to version these component specs accordingly

### Accessibility

The first goal in this reorg is to make explicit the different parties responsible for the different components of the spec. Using a name for the component that mirrors what data the component describes/contains, helps a newcomer jump in more easily.

This is consistent with how both GBFS and GTFS/RT name their feed components; both of which were specifically called out as inspiration for this spec. 

### Versioning

The second goal is to allow for independent versioning of these different components. Imagine for example a new field being required in trip data coming from the **Provider** -- the field could be added, version updated, without necessarily affecting the **Agency** side of the spec. 

This naming scheme also eliminates possible confusion between the *name* of the component and the *version* of the component. In the current spec, we might have `v2.0` of `v0.1`, and `v1.0` of `v0.2` -- this could get confusing quickly. 

### Future Growth

Finally, this naming scheme would allow for adding entirely new components, again more-or-less independently of the other components (versioning, naming, implementation, etc.) Maybe this isn't a concern right now, but at least in theory it would be *easier* to do so in the future.

hunterowens:
I like renaming the specs and splitting them. Please include (formerly v0.1 and v0.2 in the pull request). 

For future growth this additionally makes lots of sense. Could even imagine future splits. 

Tagging @LADOTBikeshare and @movementai for more comments.

hunterowens:
#7  is merged, had to make a quick update about the LADOT guidelines in the repository. Closing issue.

PR #7, opened 2018-05-24
Reorganization of the spec

thekaveman:
This is a proposed reorganization following up on #5 

I am totally open to suggestions, renames, etc. (I used **provider** and **agency** but other options include **operator** and **municipality** for example). 

There were no *substantive* content changes introduced. Some language was restructured, or edited for brevity or generalization. Some additional minor normalization of field names crept in as well.

I opted to organize components using *folders*, with a `README.md` file in each that contains that portion of the spec. My thinking was this would allow for additional files per component (e.g. an example `.json` data file). Another option is just a `provider.md` and `agency.md` file in the root.


hunterowens:
This LGTM. @LADOTBikeshare @toddapetersen @thecorconian @xxmarcelxx?

thekaveman:
An even simpler option is the single `mobility-data-specification.md` file we had before, but with these renames.

hunterowens:
Merging

PR #14, opened 2018-06-03
Suggested API additions

jfh01:
I made three suggested additions to the provider API:

- trip/accuracy: For both technical (GPS accuracy limitations) and privacy (deliberate blurring of location information), the start/end point information may have limited accuracy. This flag allows data consumers to understand the limitations of the data reported. This presumes blurring will by done by a fixed radius. We may need to expand this to support blurring by geographic area (e.g. census block group). 

- trip/parking_verification: Many dockless providers request a photo of the parked vehicle from the user to verify correct parking. This field could be used to provide access to the photos to agencies for spot checks or other monitoring.

- availability/allowed_placement: A boolean to indicate whether the provider believes the placement was in compliance with rules. Could be used by agencies to analyze rate and location of misplacements by users (e.g. out of service area), and to compare providers on their success at avoiding misplacement.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -36,7 +36,7 @@ Response:
 | `trip_distance` | Integer | Required | Trip Distance, in Meters | 
 | `start_point` | Point | Required | | 
 | `end_point` | Point | Required | | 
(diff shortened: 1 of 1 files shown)


xxmarcelxx:
@jfh01 these are great suggestions - we are going to merge them into the spec.  grazie!

PR #21, opened 2018-07-02
New Routes Implementation.  fixes #19

### LINKED ISSUE #19: Suggestion: Replace start_point,end_point, start_time, end_time, route, and sample rate with route containing timestamp
First and last points in route are start and end; all points in route are marked with timestamps.

hunterowens:
This  change replaces `start_point`, `end_point` and associated field with `routes`, a new datatype which is a `geojson` list of points to represent the route a device took during a trip.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -34,17 +34,53 @@ Response:
 | `trip_id` | UUID | Required | a unique ID for each trip | 
 | `trip_duration` | Integer | Required | Time, in Seconds | 
 | `trip_distance` | Integer | Required | Trip Distance, in Meters | 
-| `start_point` | Point | Required | Must be in WGS 84 (EPSG:4326) standard GPS projection | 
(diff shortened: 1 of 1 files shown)


hunterowens:
@dmklinger @thekaveman does this meet your needs?

thekaveman:
This leaves `start_time` and `end_time` as part of the `Trip` structure, where we assume this data would also be captured as part of the first/last points in the new `Route` object, correct?

In #19, @dmklinger suggests removing these time fields. I can see the benefit of not having to dive into the geoJSON to get these two values, but they may be duplicative. Were they intentionally left as part of the `Trip` object? Perhaps these represent when the vehicle was locked/unlocked by the user, vs. when the vehicle actually started moving?

hunterowens:
@thekaveman / @dmklinger do we have resolution on this?

PR #28, opened 2018-08-08
Removed HTML tags

toddapetersen:
Carry over from Swagger

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -250,21 +250,21 @@ Body:
 
 ## Enum Definitions 
 
-<a name="vehicle_type"></a> For `vehicle_type`, options are:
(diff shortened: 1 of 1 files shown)


thekaveman:
These were used by the jump links in the tables that reference the enums - if the anchors need to be removed, the (now broken) jump links should also be removed, yes?

E.g. under [`remove-vehicle`](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/master/agency#remove-vehicle) the link pointing to *Reason* is broken.

toddapetersen:
Fixed.   Will issue pull request with some additional changes I'm working on.   Thanks @thekaveman

PR #30, opened 2018-08-09
Consistency in underscores

sethherr:
It doesn't make sense to have all the responses use snake case (underscores) with endpoints with dashes.

Partially because consistency but also because most web frameworks use underscores. The only one I'm aware of that uses dashes is Django, and [that seems somewhat contentious](https://medium.com/@raiderrobert/django-url-names-should-be-underscore-separated-7815e3d4e7c2)

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -8,24 +8,24 @@ This specification contains a collection of RESTful APIs used to specify the dig
 
 ## Table of Contents
 
-* [register-vehicle](#register-vehicle)
-* [remove-vehicle](#remove-vehicle)
-* [service-vehicle](#service-vehicle)
-* [report-maintenance](#report-maintenance)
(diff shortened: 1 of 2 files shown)


sethherr:
I just switched to deregister vehicle, since that change was merged in #31 

Personally, I don't think it makes sense to have the header be "deregister" when the endpoint is "remove" though

hunterowens:
I'm used to dashes, but I also used to write stuff in Django so maybe that's why. Requesting @toddapetersen to review

hunterowens:
@sethherr we're gonna go ahead and use underscores. 

We've made the appropriate changes in the provider spec, however, Agency is still behind. Would you mind rebasing and pushing that to us and we can merge in?

PR #32, opened 2018-08-10
Change device_id to VIN

toddapetersen:
Suggesting a change to the identifier for a vehicle away from a UUID and using the operators VIN identifier instead.   The thought here is that each bike manufacturer already has a unique identifier for their vehicle embedded in their code (QR code or otherwise).  It is inherent to the business.   Let's use that number so they could be correlated to real bikes on the street if it ever became necessary.

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -39,12 +39,8 @@ Body:
 | `vehicle_mfgr` | Enum | Required | Vehicle Manufacturer |
 | `vehicle_model` | Enum | Required | Vehicle Model |
 | `vin` | String | Required | Vehicle Identification Number assigned by Manufacturer or Operator |
+| `device_id` | UUID | Required | UUID geneated by requestor to uniquely identify the device. Must be consistent with that physical device. If UUID is not unique, API will return a status code not equal to `200` | 
 
-Response:
-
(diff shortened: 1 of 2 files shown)


hunterowens:
This is good, but creates some weird issues with indexing in a merged database and could degrade analytics/query performance. 

For example, if say, both `company A` and `company B` used sequential integer IDs, ie, {1,2}, we could have a trips database where we have to have a joint index, and that is slower than a uuid index. 

Additionally, if `company A` used a UUID and `company B` uses int/longs, that becomes a pain to index and load into a DB. 

Tagging @dmklinger and @hannah-ross for more thoughts.

jannasmith:
could we stipulate that provide us with a UUID that is based on the VIN - for example, a 4-digit operator number we assign preceding the VIN to create the entire identifier. then we could use that to identify devices on the street as well, and it wouldn't create conflicts on our side. Would that be a reasonable ask of the operators? Also would we need to require that all UUIDs/VINs be the same number of characters or anything like that?

toddapetersen:
What about taking the the provider_id and the vehicle_id and concatenate them to make a unique key for the table and query.

hunterowens:
@toddapetersen does this look good?

Issue #29, opened 2018-08-09
Reconcile changes made to Provider to Agency


toddapetersen:
Hunter and David,  You guys mentioned in our last meeting that there were changes you made during implementation of the Provider side that need to be made to the Agency side.   Could you make those changes before we start building?

hunterowens:
Tagging @dmklinger

hunterowens:
@toddapetersen timeline you're looking for. 

I think the major changes off the top of my head are 

* the new `route` structure
* WGS:84 Compliance 
* anything else?

toddapetersen:
I'd like to get it in by Tuesday next week.   Would also like to discuss some other changes we're considering to make the Agency side a little simpler for bikes.    I'll be in Monday if you have time to get together.

hunterowens:
sounds good. I will take a stab at drafting today but let's meet on Monday to hash out.

hunterowens:
I believe we actually only had to move stuff from Agency into Provider. Opened PR #34.

PR #34, opened 2018-08-17
adding enums back into provider, matching to agency

hunterowens:
* The Device_Type enum only made it into provider. This matches it into agency.


hunterowens:
#29

Issue #37, opened 2018-08-21
Device ID / UUIDs cannot be reassigned


hunterowens:
Clarify that they cannot reassign these numbers. Per @jannasmith

toddapetersen:
John and I have been down this road several times.   Our going in position should be a single UUID for a vehicle (no-reassignment).    We'll let the companies decide what makes a scooter a scooter, bike a bike (Frame, battery, etc).   Most likely only minor maintenance will be performed due to the low cost of these devices relative to labor and parts.    As we progress toward higher value assets like Autonomous vehicles, this policy will become more important and we will get more specific about what makes car a car, etc.    Make sense?

jannasmith:
Yes - this is mostly in response to some insight Jose and I gained from a chat with the folks from Seattle yesterday and the data they've already collected, that all the connected and tracking technology seems to be housed in the GPS/lock device. The concern is when say, a vehicle's frame is damaged, and the undamaged GPS/lock device is placed on a different vehicle, the operator may transfer the ID # to the replacement frame. Do we track it using the same UUID as the previous vehicle frame, or have them register it as a new vehicle? Basically Seattle's advice was whatever strategy we decide, to be clear about it up-front so operators are all doing the same thing.

jannasmith:
I want to get input from Jose just in case before finalizing this (if he anticipates needing to track the individual bikes/scooters for maintenance issues or anything else) @LADOTBikeshare, let me know if this needs to be resolved urgently though

jannasmith:
OK cool we can agree to referring to the GPS/Lock Device as the "Device" to which the UUID will be assigned.

PR #40, opened 2018-09-05
[WIP] Provider cleanup

thekaveman:


thekaveman:
Added the `[WIP]` designation to indicate that I'm still making sure this PR is good to go... stand by.

thekaveman:
@hunterowens @black-tea: `trips` have both the `provider_id` and `provider_name`, as well as a `vehicle_id` and `vin` -- one identifier each for machines and humans. 

Should `status_changes` mirror this approach? Currently it has only the "for humans" identifiers.

hunterowens:
Yes, let's duplicate.

thekaveman:
@hunterowens done. I left the field name as `device_id`, based on the discussion in #37, and the conclusion that this is really the identifier for the GPS device, no matter what "vehicle frame" it is attached to. I added a note under *General Information* to reflect this conclusion. 

So this PR should fix #37.

PR #42, opened 2018-09-10
Add Sem-Var notes:

hunterowens:


thekaveman:
I suggest the initial version be tagged at `0.1.0` vs. `0.0.1`, but it's a minor difference.

PR #43, opened 2018-09-10
Some additional Provider cleanup

thekaveman:
Mostly just for consistency's sake.

  - cleanup enum tables
  - vin => vehicle_id
  - reason => event_type_reason
  - location => event_location

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -83,10 +83,10 @@ Data: `{ "trips": [] }`, an array of objects with the following structure
 
 | Field | Type    | Required/Optional | Comments |
 | ----- | -------- | ----------------- | ----- |
-| `provider_name` | String | Required | The public-facing name of the Provider |
 | `provider_id` | UUID | Required | A UUID for the Provider, unique within MDS |
+| `provider_name` | String | Required | The public-facing name of the Provider |
 | `device_id` | UUID | Required | A unique device ID in UUID format |
-| `vin` | String | Required | Vehicle Identification Number assigned by Provider |
+| `vehicle_id` | String | Required | Vehicle Identification Number assigned by Provider |
 | `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
 | `propulsion_type` | Enum | Required | see [propulsion types](#propulsion-types) table |
 | `trip_id` | UUID | Required | a unique ID for each trip |
@@ -102,19 +102,19 @@ Data: `{ "trips": [] }`, an array of objects with the following structure
 
 ### Vehicle Types
 
-| vehicle_type | description |
-|-------|-------------|
-| bicycle | |
-| scooter | |
-| recumbent | |
+| `vehicle_type` |
+|--------------|
+| bicycle      |
+| scooter      |
+| adaptive     |
 
 ### Propulsion Types
 
-| propulsion_type | description |
-|-------|-------------|
-| human | |
-| electric | |
-| combustion | |
+| `propulsion_type` |
+|-----------------|
+| human           |
+| electric        |
+| combustion      |
 
 ### Routes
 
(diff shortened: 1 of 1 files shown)


provider/README.md
```diff
@@ -83,10 +83,10 @@ Data: `{ "trips": [] }`, an array of objects with the following structure
 
 | Field | Type    | Required/Optional | Comments |
 | ----- | -------- | ----------------- | ----- |
-| `provider_name` | String | Required | The public-facing name of the Provider |
 | `provider_id` | UUID | Required | A UUID for the Provider, unique within MDS |
+| `provider_name` | String | Required | The public-facing name of the Provider |
 | `device_id` | UUID | Required | A unique device ID in UUID format |
-| `vin` | String | Required | Vehicle Identification Number assigned by Provider |
+| `vehicle_id` | String | Required | Vehicle Identification Number assigned by Provider |
```
hunterowens: for vehicle ID or VIN, we should probably add something like `must match the number/serial on physical device"

provider/README.md
```diff
@@ -102,19 +102,19 @@ Data: `{ "trips": [] }`, an array of objects with the following structure
...
-| vehicle_type | description |
-|-------|-------------|
-| bicycle | |
-| scooter | |
-| recumbent | |
+| `vehicle_type` |
+|--------------|
+| bicycle      |
+| scooter      |
+| adaptive     |
```
hunterowens: I like replacing recumbent with adaptive. @LADOTBikeshare can you thing of any other vehicle types we should start supporting?

provider/README.md
```diff
@@ -168,22 +168,22 @@ Data: `{ "status_changes": [] }`, an array of objects with the following structu
...
 | Field | Type | Required/Optional | Comments |
 | ----- | ---- | ----------------- | ----- |
-| `provider_name` | String | Required | The public-facing name of the Provider |
 | `provider_id` | UUID | Required | A UUID for the Provider, unique within MDS |
-| `vin` | String | Required | Vehicle Identification Number assigned by Provider |
+| `provider_name` | String | Required | The public-facing name of the Provider |
 | `device_id` | UUID | Required | A unique device ID in UUID format |
+| `vehicle_id` | String | Required | Vehicle Identification Number assigned by Provider |
 | `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
 | `propulsion_type` | Enum | Required | see [propulsion types](#propulsion-types) table |
```
hunterowens: Shouldn't we add an array here?

provider/README.md
toddapetersen: Technically, Adaptive bikes come in many shapes and sizes based on the type of injury they are accommodating.  I think we need to stay as specific as possible.

provider/README.md
toddapetersen: A VIN is a 17 character string.  Not sure why it would need an array?   https://www.thezebra.com/insurance-news/2130/vehicle-identification-number-101/

provider/README.md
toddapetersen: Yes, I agree.   Go ahead and change.

provider/README.md
thekaveman: It might be useful to add that `Description` column back and populate with some clear distinctions, regardless of what keyword we choose.

provider/README.md
thekaveman: @hunterowens did you mean to comment on the `propulsion_type` field, which we discussed might be more than one type?

provider/README.md
thekaveman: Another option is to just support `bicycle` and `scooter` for now, until companies deploy additional vehicle types at which point they can submit a PR 😄

provider/README.md
thekaveman: dbb2f4a

provider/README.md
hunterowens: @thekaveman Let's go with the latter option then.

thekaveman: 2a98c22

provider/README.md
thekaveman: 20755c6

PR #48, opened 2018-09-12
Adding electric_assist as a propulsion_type.

aickin:
Just a small suggestion that it might be useful to distinguish between completely electric vehicles (like Lime and Bird Scooters) and pedal assist vehicles that combine human and electric power (like Jump bikes).

I added `electric_assist` to the `propulsion_type` enum to fix this.

I understand if you want to call both of these kinds of vehicles `electric`, but I think the distinction could be useful!

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -112,6 +112,7 @@ Data: `{ "trips": [] }`, an array of objects with the following structure
 | `propulsion_type` |
 |-----------------|
 | human           |
+| electric_assist |
 | electric        |
 | combustion      |
 


thekaveman:
Thanks for this feedback @aickin! You are definitely correct in that it is an important distinction, and one that we care about.

The good news is, `propulsion_type` is actually meant to be an array field, meaning a provider could give:

```json
{
    "propulsion_type": ["human", "electric"]
}
```

to cover the "electric-assist" scenario.

@hunterowens @toddapetersen thoughts?

aickin:
Hmmmm... I just noticed that `propulsion_type` is an array, which maybe implies that pedal-assist should be ['human', 'electric']. I'd argue that ['human', electric'] should be a bike that can be pedaled OR where you can use a throttle, which is genuinely different than pedal assist. But I'm interested to hear other perspectives.

aickin:
@thekaveman Thanks for the explanation; our comments collided :)

To give a concrete example, how should we encode something like [a GenZe 200 bike](https://www.genze.com/e-bikes/?gclid=EAIaIQobChMInvjQ3rC03QIVCIN-Ch2aIgVqEAAYAiAAEgLS9vD_BwE#--SIL-350-SKD), which has "Throttle and Pedal Assist Ride Modes"?

In the current spec, you could encode it as `["human", "electric"]`, but then it's indistinguishable from bikes that only have pedal assist (like JUMP bikes). Thoughts?

hunterowens:
@aickin I think the best way to capture that would be to add `electric_assist` as recommended in this PR, as that bike could be coded `["electric_assist", "electric"]` to suggest the two modes of operation compared to jump's `["human", "electric"]`.

PR #52, opened 2018-09-12
Consider taking out trip_duration

aickin:
I noticed that the trip schema has `trip_duration`, `start_time`, and `end_time`. I may be misunderstanding what these fields mean, but it seems like they may be redundant, as the trip duration should just be end time minus start time.

In this PR, I suggest removing `trip_duration`, although you could remove `end_time` (or even `start_time`, but that would be... strange :) ).

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -90,7 +90,6 @@ Data: `{ "trips": [] }`, an array of objects with the following structure
 | `vehicle_type` | Enum | Required | See [vehicle types](#vehicle-types) table |
 | `propulsion_type` | Enum[] | Required | See [propulsion types](#propulsion-types) table; allows multiple values |
 | `trip_id` | UUID | Required | A unique ID for each trip |
-| `trip_duration` | Integer | Required | Time, in Seconds |
 | `trip_distance` | Integer | Required | Trip Distance, in Meters |
 | `route` | Route | Required | See detail below |
 | `accuracy` | Integer | Required | The approximate level of accuracy, in meters, of `Points` within `route` |


hunterowens:
One question that was raised today is how to capture "time to start". 

I think a better way of doing this would be suggesting that 

`trip_duration` represents the time from "users selects transit on phone" to "user completes trip" 
`start_time` represents the time when device was unlocked. 
`end_time` represents the time when the user completed the trip. 

This way, we can accurately compare MDS trips to other modes of transportation. This difference also captures important info for rideshare/micro transit.

ezheidtmann:
I think the `trip_duration` definition could be clearer still. For example, if I schedule a TNC ride for tomorrow, that shouldn't be the start of `trip_duration`.

If LADOT is looking for information on how long the asset (scooter, bike, AV, TNC driver) is dedicated to the task of transporting the individual, then the spec should require specifically that information. For example, `reserved_time` in addition to `start_time` and `end_time`. But you could probably derive this information from events (status changes) associated with the device.

If LADOT is trying to quantify waiting time, that's likely to be tough because even providers don't know how much time a rider spent trying to find a scooter & walk to it.

hunterowens:
On the LADOT roadmap is to support Taxi data in MDS vs the current format, which involves physical disks (including, I've heard, floppy drives at some point). 

Given that, I think "reserved_time" as a replacement for "trip_duration" makes a lot of sense, but I might be missing some nuance here.

jfh01:
As I've thought about this more, I think the right way to approach this is by removing end_time, while keeping start time and duration.

Not all cities are comfortable holding highly precise data due to privacy / FOIA concerns. One potential privacy measure for these cities would be to round times to the nearest 15 minute or one hour block. If you do that with start and end times, you lose the ability to see how long trips actually took.

As a proposal:

`start_time` -- timestamp for when the user first initiated the trip request
`reserved_duration` -- seconds between start_time and when the user unlocked/entered the vehicle (0 if the user did not reserve the vehicle in advance of unlocking)
`travel_duration` -- seconds from when the user unlocked/entered the vehicle until they locked/exited the vehicle

For LA, start_time would be precise. Other cities may round it, but would still be able to see exactly how long trips and reservations took.

Issue #46, opened 2018-09-11
Consider using JSON Schema


ian-r-rose:
The mobility spec already provides some nice narrative description of the inputs and outputs for the REST API. However, there is still some room for ambiguity, and it would be useful to be able to test providers/agencies against regressions and compliance with the spec.

One commonly-used approach to this problem is to use a [JSON schema](https://json-schema.org). This allows you to describe JSON data structures in a well-typed fashion. This allows for
* validation
* regression testing
* a source of truth for the spec to guard against different interpretations of the narrative.
* Easier tracking of spec versions

For instance, a schema for the response body of the `/deregister-vehicle` might look something like:
```json
{
  "properties": {
    "unique_id": {
      "type": "string"
    },
    "device_id": {
      "type": "string"
    },
    "reason_code": {
      "type": "string",
      "enum": ["rebalancing", "maintenance"]
    }
  },
  "type": "object",
  "additionalProperties": false
}
```
High-quality JSON-schema validators should exist in most web-friendly languages.

thekaveman:
I think this is a great idea.

I will start putting together a PR for `provider` using JSON-schema so we can explore this idea further.

thekaveman:
I started putting something together over here: https://github.com/CityofSantaMonica/mobility-data-specification/blob/json-schema/provider/schema.json

Take a look at let me know if this seems like the right direction. I generated this using some sample data I cooked up and the tool at https://www.jsonschema.net/

ian-r-rose:
@thekaveman that's awesome. Can you open a PR to allow for commenting?

hunterowens:
This would be a breaking change, tagging as a 0.2.0 milestone.

hunterowens:
#53 is merged. closing.

PR #53, opened 2018-09-12
Implementing JSON Schema for Provider

thekaveman:
This is for #46 and more of a test of the idea than fully-baked implementation.

There is plenty of room for improvement:

  - Define the top-level envelope structure once:
    ```json
    {
        "version": "...",
        "data": {},
        "links": {}
    }
    ```
    and re-use it across different parts of the overall MDS schema. 

  - Define the common objects (`route`, `vehicle_type`, `propulsion_type`, etc.) once and reference

  - Reference (existing?) geoJSON schemas if applicable

I'm sure there are others.


provider/schema.json
```diff
@@ -0,0 +1,494 @@
...
+                                "description": "The actual cost, in cents, paid by the customer of the mobility as a service provider",
+                                "default": 0,
+                                "examples": [
+                                    520
+                                ]
+                            }
+                        }
+                    }
+                },
+                "status_changes": {
```
ian-r-rose: My understanding of the narrative description is that the JSON payloads are different between whether they hit the `trips` or `status_changes` endpoints. So having the JSON schema describe both of them at once seems like it is not the right thing. Instead, I would think that there would be a separate schema for each of the endpoints (maybe both deriving from a common base schema?).

Or am I misunderstanding things?

thekaveman: No, it's one of the things I meant to put on my list above... definitely needs to be cleaned up. 

Reading about [combining schemas](https://json-schema.org/understanding-json-schema/reference/combining.html) got me thinking that something along those line might work. Just haven't thought about how to do it yet.

provider/schema.json
```diff
@@ -0,0 +1,494 @@
...
+    ],
+    "properties": {
+        "version": {
+            "$id": "#/properties/version",
+            "type": "string",
+            "title": "The MDS Provider version this data represents",
+            "examples": [
+                "0.1.0"
+            ],
+            "pattern": "^(.*)$"
```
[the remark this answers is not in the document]
thekaveman: Good point, I just left the default there but SemVer is well-defined enough that we could easily tighten it. 

I don't know about hard-coding the `minorVersion` in a pattern like `^0\.1\.[0-9]+$`, but maybe with some kind of generation tool it wouldn't be that big of a deal.

ian-r-rose:
I think CityofSantaMonica/mobility-data-specification#1 is close to ready. @thekaveman and I were uncertain about whether it was best to include some of the tooling (validation + schema generation scripts) in this repo, or an associated one. Also happy to open that PR against this repo to supersede this one.

thekaveman:
@hunterowens I thought this was good for `0.1.x` since it is more of a clarification/expression of the existing spec than a change? I'd like to work from this as I'm building out the ingestion framework if possible (since we didn't have clarification on some field types for example). 

Outstanding concerns to address for this to be ready:

 - [ ] `event_location` is a GeoJSON `Feature` #74 
 - [ ] `timestamps` are float Unix epoch seconds (#87 ) unless anyone has other thoughts?

ian-r-rose:
Also, are you using `dev` to track feature changes, while `master` is the official spec? If so, we should reopen this PR to be against `dev`.

hunterowens:
@ian-r-rose going forward that is the case. I just changed this PR to be against `dev` rather than `master`.

thekaveman:
I'm going to rebase this on `dev`, stand by...

thekaveman:
Ok, I think we still need the schema to address the following: 

 - [ ] the latest clarifications from #94

(I'm going to go through it again and potentially add to this list, feel free to do so anyone else @hunterowens @ian-r-rose)

generate_schema/generate_provider_schema.py
```diff
@@ -0,0 +1,123 @@
...
+
+    1. Only allow GeoJSON Point features.
+    2. Those Point features *must* include a `timestamp` number property.
+    """
+    # Get the canonical Feature schema
+    f = requests.get("http://geojson.org/schema/Feature.json")
+    feature = f.json()
+    # Modify some metadata
+    feature.pop("$schema")
+    feature["$id"] = "#/definitions/FeatureMDS"
```
ian-r-rose: I wonder if we should call these something like `PointFeatureMDS` and `RouteFeatureCollectionMDS`, in order to leave the door open for other kinds of features and feature collections.

thekaveman: Good idea

thekaveman: Changes incoming, just turning it into a bit of a larger refactor for the future proofing

generate_schema/provider/status_changes.json
```diff
@@ -0,0 +1,157 @@
...
+                  "user_drop_off",
+                  "rebalance_drop_off",
+                  "maintenance_drop_off",
+                  "user_pick_up",
+                  "maintenance",
+                  "low_battery",
+                  "service_end",
+                  "rebalance_pick_up",
+                  "maintenance_pick_up"
+                ]
```
hunterowens: Do we need to do anything to link `event_type` and `event_type_reason` 

ie, you can't have a `removed` event with `user_drop_off` as reason.

thekaveman: We [talked about that](https://github.com/CityofSantaMonica/mobility-data-specification/pull/1#discussion_r218934410) over in the original PR but decided to hold off for now. It *may* be possible but it is complex and we need to think on it more.

hunterowens: ok.

PR #54, opened 2018-09-12
Add a note about array Types to Provider

sethherr:


thekaveman:
Unfortunately this changes the intention of those fields, @sethherr. These are meant to be *arrays*, e.g. a vehicle may have more than one `propulsion_type`. See the related discussion on #48.

sethherr:
Ahh. That's an important distinction, that Type declaration wasn't clear to me.

thekaveman:
Indeed there is room for ambiguity. See our discussion on #46 and #53, where we are exploring JSON-Schema to more clearly define the spec in a machine-readable/machine-verifiable format.

sethherr:
I put a note about it being an array in the description, I think this would have clarified it for me?

I think JSON-Schema could be great, this is more of a quick fix...

PR #56, opened 2018-09-12
Correct JSON API pagination

### LINKED ISSUE #49: Pagination doesn't work the same here as in JSON API
In the provider API definition, it says that "pagination must comply with the JSON API specification", but the JSON API spec says:

> Pagination links MUST appear in the `links` object that corresponds to a collection. To paginate the primary data, supply pagination links in the top-level `links` object.

This spec puts the `first`, `last`, `prev`, `next` URLs at the top level of the response, rather than inside in a `links` object. I feel like either this spec should use the `links` object or not say that paginations complies with JSON API.

thekaveman:
Fixes #49

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -52,10 +52,12 @@ The following keys must be used for pagination links:
             //etc.
         }]
     },
-    "first": "https://...",
-    "last": "https://...",
-    "prev": "https://...",
-    "next": "https://..."
+    "links": {
+        "first": "https://...",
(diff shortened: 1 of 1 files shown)


aickin:
This looks good!

aickin:
Actually, one extra piece of feedback just occurred to me. I think it'd be useful to explicitly note whether or not MDS complies to the JSON API spec as a whole, or just to the pagination section. But this PR is good as is!

PR #64, opened 2018-09-13
Add detail to query params for providers

asadowns:
For trips and status_changes break query params
	out into their own section.

Add more detail around how to specify time
	and location based parameters.

For location this is just a suggestion.

There may be url length concerns when passing in large polygons
	in query params.

A better approach might be to pass in service_area id or
	use a location filter based on the token and
	filter data further after retrieval

Comments welcome.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -113,7 +113,8 @@ The trips API should allow querying trips with a combination of query parameters
 * `start_location`
 * `end_location`
 
-All of these query params will use the *Type* listed above with the exception of `start_location` and `end_location` which can be provided as stringified GeoJson Polygons.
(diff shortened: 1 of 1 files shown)


hunterowens:
@asadowns any thoughts on a POST based implementation? It was raised as an issue in #78

oderby:
Currently the documentation for the [`/service_areas` endpoint](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/master/agency#service_areas) of the Agency API states that it's a GET request with a Body containing `provider_id` and `service_id`. Per discussion in #78 - is that something you feel like fixing here @asadowns @thekaveman or would you rather I open a separate PR?

thekaveman:
@oderby you are absolutely correct, good catch. I think opening a separate PR for that fix makes sense.

`provider` and `agency` are kind of evolving a different pace right now but eventually as more clarity is realized, the two should align pretty well.

asadowns:
@oderby I'll update my PR here.

Issue #58, opened 2018-09-12
Flat file implementations of provider API


jfh01:
I am concerned that many cities may not have the technical capacity to consume a dynamic, query-based API. 

The /trips and /events endpoints both lend themselves to a flat file format. While LA will clearly want to require a dynamic API, it may be worth broadening MDS to allow for other cities to request the same data in flat file format (e.g. monthly/weekly CSVs).

Are LADOT folks amenable to including some guidance on "flat file MDS implementations" in the spec? I suspect this would make MDS more appealing to other cities.

Happy to do put in pull request of people think this is a good idea.

ezheidtmann:
I can't speak to the relevance of a flat file implementation specifically, but I strongly agree that there is value to having a well-defined object format (i.e. events, trips, routes) independent of any specific interface. Maybe we can start by breaking things down into object types and interfaces?

jfh01:
This makes a ton of sense to me. 

@hunterowens: Thoughts on this? Related discussion in #60 as well.

Idea would be to:

- Define data types that can be used as returns from API calls OR shared via static/flat file
- Define API endpoints as query mechanisms which return typed data

LA would require that all providers implement the API, but other cities could adopt the data component of the spec while using something other than an API to move data.

bhandzo:
I think a flat file standard would be 💯 as many cities won't have the ability to consume API endpoints but will want to access the data. Main concern here is some standardized data chunking or ability to break what will certainly be very large data sets up to be manageable.

migurski:
Strongly agree with this. I’d like to see simplifications to the Provider API that allow it to be expressed fully as a flat file interface instead of an additional component to the spec. This would position the provider API as an easily-consumable data feed accessible to smaller cities with limited IT or engineering staff, but available data analysts familiar with spreadsheet tools.

A sample [trips request and response](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/master/provider/README.md#trips) might look like this:

```
> GET /trips
> Accept: text/csv

< HTTP/1.1 200 OK
< Content-Type: text/csv
< 
< provider_id, provider_name, device_id,…
```

Parameters for the `/trips` endpoint are unspecified, but might include date and bounding box arguments that can be manipulated in a web browser.

This change would conflict with https://github.com/CityOfLosAngeles/mobility-data-specification/issues/46 which pulls the API away from common skillsets present in cities and toward those typically found in software engineering orgs.

[...]

Issue #51, opened 2018-09-12
Consider moving accuracy to properties in the route.


aickin:
When a GPS device moves through an environment, especially an urban environment, the accuracy of the signal can vary tremendously. This means it's possible that some points on a GPS route may have an accuracy of a meter while others points in the route may have accuracy of 10 or 30 meters or whatever. In this kind of case, it's not entirely clear what to send in the `accuracy` field for a trip, since it varies so much.

I'd suggest that maybe it would be better to move the `accuracy` field into the `route` GeoJSON, like this:

```
"route": {
    "type": "FeatureCollection",
    "features": [{
        "type": "Feature",
        "properties": {
            "timestamp": 1529968782.421409,
            "accuracy": 1,
        },
        "geometry": {
            "type": "Point",
            "coordinates": [
                -118.46710503101347,
                33.9909333514159
            ]
        }
    },
    {
        "type": "Feature",
        "properties": {
            "timestamp": 1531007628.3774529
            "accuracy": 10,
        },
        "geometry": {
            "type": "Point",
            "coordinates": [
                -118.464851975441,
                33.990366257735
            ]
        }
    }]
}
```

ezheidtmann:
Strongly agree; accuracy information should always be associated with a point location.

In our experience with smartphone chips, GPS location accuracy is communicated in two ways.

 * [Android](https://developer.android.com/reference/android/location/Location#getAccuracy()) - `accuracy` - "radius of 68% confidence" (although our experience shows this definition is not rigorous)
 * [iOS](https://developer.apple.com/documentation/corelocation/cllocation/1423599-horizontalaccuracy) - `horizontalAccuracy` - "radius of uncertainty for the location"

In the general case, [it doesn't appear that accuracy can be reliably computed from standard NMEA sentences](https://www.gps-forums.com/threads/estimating-accuracy-from-raw-nmea-data.46273/) but I bet most stacks can produce a radius of uncertainty.

Of course, the real location isn't always within the radius, even when the radius is very small. And both smartphone platforms have additional layers of location processing that can add errors. So we always track which platform produced a location so we can process differently if needed.

---

So I'd suggest using the field `horizontalAccuracy` for clarity, and perhaps including a GPS platform enum in the route or trip description.

```
"route": {
    "type": "FeatureCollection",
    "features": [{
        "type": "Feature",
        "properties": {
            "timestamp": 1529968782.421409,
            "horizontalAccuracy": 1,
        },
[...]

PR #68, opened 2018-09-14
Discussion: provider flat-file representation

thekaveman:
There has been a lot of interest in this over at #58. Here's one starting point to hopefully further the discussion. 

The major changes of note:

  - the geoJSON `route` has gone away in favor of the flattened `trip_point` (including the move of accuracy as proposed on #51)

  - in the spirit of making the spec more friendly and accessible (as with flat files), this changes the date/time formatting to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) as originally discussed by @monolyst way back in #8 

As part of addressing the desire to have a well-defined object format, `.csv` files in the `models/` directory describe the fields, types, requirements, etc. for each (rather than tables in the README). This is still a little hand-wavy in my opinion, but at least its machine-readable 🤷‍♂️. Not sure if/how this effort can intersect with #46 and its current proposal #53


ezheidtmann:
Interesting. Just a half-baked thought here, but what if we support the CSV use-case by providing an open-source transform tool that produces CSV based on JSON input?

I like the idea of machine-readable schema, but perhaps we should use an established schema language like JSON Schema. Unless the CSV schema format you use is a standard already?

migurski:
@ezheidtmann A transform tool could temporarily bridge this gap. I’d like to see it officially supported as a client-side tool on an MDS spec page, maybe found inside this repo? It would require that any MDS Provider API endpoint be fully CORS-compliant.

hunterowens:
👍 to the idea of a transformation tool rather than a CSV endpoint. @ezheidtmann @migurski.

ezheidtmann:
Great -- that tool can be developed when a city needs it.

thekaveman:
It'll still be here for folks to comment on... but I'm closing this PR as we look towards different approaches.

PR #72, opened 2018-09-14
Adding providers.csv. fixes #39

### LINKED ISSUE #39: Discussion: create a central registry of Providers?
Similarly to how GBFS maintains a central [`systems.csv`](https://github.com/NABSA/gbfs/blob/master/systems.csv) file, should MDS maintain a `providers.csv` (or similar)?

MDS regulators (*municipalities*) need a way to uniquely identify each of the different `provider` data feeds. On the other hand, a given `provider` may have a different MDS feed for many different municipalities.

Rather than forcing each `provider` to register with each municipality separately (to obtain a unique `provider_id`), the idea is to maintain a single, central source of truth. 

An example schema:

| provider_id | provider_name | url |
| ------------------|-------------|-------|
| MDS-wide unique provider identifier | human friendly company name | Company website URL |

Does it make sense to have the `provider` API URL as part of this registry? If that is the case, the API needs to support queries for a particular municipality (e.g. by geo-fenced area).

hunterowens:
This pull request adds an empty provider csv file for each operator to add in a PR. 

Each company should provide their name, a UUID ID, url, and their MDS url in a following PR.

### DIFF
--- a/providers.csv
+++ b/providers.csv
@@ -1,0 +1,1 @@
+provider_name, provider_id, url, mds_api_url


hunterowens:
Yeah. We should probably add that to documentation somewhere.

Issue #60, opened 2018-09-12
How should access to the 'provider' API be authenticated?


jfh01:
Given that the provider API is intended for use by cities (and their vendors/patners) and NOT the general public, how should providers implement access control on their API endpoints? 

Simple option would be for LADOT to issue a shared secret at the time when providers are granted an API.

Longer term, probably need a more cryptographic or account-oriented solution, since cities may want the ability to authorize 3rd parties (e.g. vendors) to access MDS data on their behalf.

aickin:
I’d argue that it might be a good choice to remove transport protocol completely from this spec, or at least put it into a separate optional spec. 

For example, you could imagine a provider and agency agreeing that the provider just emails a large data dump of all trips every week or that the provider FTPs the data to a central server. Authentication would be figured out differently in each of those situations.

thekaveman:
@aickin that's an interesting thought, I can see how that kind of setup could be more useful to certain organizations that maybe aren't quite prepared to deal with API calls and AWS databases etc. 

That being said... do we have any of those agencies listening in, that want to put forth their thoughts on this matter? Given that this is still early days for MDS, I would hesitate to add too many "options" for how to move this data around etc. until we've at least established the baseline spec and its utility. Unless we have a willing partner agency that wants to try a different transport protocol approach?

jfh01:
I proposed a static file option in #58. Given the technical capabilities of LA and a few other cities, and the likelihood of vendors emerging to help cities consume real-time APIs, it seems wise to have a game plan on this. But for many cities, something far less sophisticated is likely the best option.

jfh01:
@thekaveman Good question. My guess is that the cities least equipped to deal with a full API are also the least likely to be hanging out in a GitHub issues list. Might need to figure out a way to do some outreach to them.

hunterowens:
@LADOTBikeshare has been maintain a list of all the cities we've reach out to with email address, and there is also NACTO.

asadowns:
I would suggest a [JWT](https://jwt.io/) that is issued by the provider to LADOT (or other agencies). This provides a lot of customization from the provider's perspective if there is other information they want to encode in the token and allows the providers to authorize accounts with well-defined permissions for accessing only MDS resources.

From the agency's perspective it is also relatively easy to issue new tokens and track token usage to prevent improper access or sharing. We also have envisioned agencies being able to generate their own tokens with bounded from this master token for use by their employees or vendors which seems like a win.

[...]

PR #81, opened 2018-09-19
Add auth.md to document provider authentication

asadowns:
Provide information on authentication with JWT and minimal payload.

Update provider/README.md to reference auth.md

Refs Issue [#60](https://github.com/CityOfLosAngeles/mobility-data-specification/issues/60)

### DIFF
--- a/provider/auth.md
+++ b/provider/auth.md
@@ -1,0 +1,11 @@
+# Authorization
+
+Authorization for **provider** API should be provided via a JSON Web Token or [JWT](https://jwt.io/introduction/).
+
+JWTs provide a safe, secure way to verify the identity of an **agency** and provide access to MDS resources without providing access to other, potentially sensitive data.
+
+> JSON Web Token (JWT) is an open standard ([RFC 7519](https://tools.ietf.org/html/rfc7519)) that defines a compact and self-contained way for securely transmitting information between parties as a JSON object. This information can be verified and trusted because it is digitally signed. JWTs can be signed using a secret (with the HMAC algorithm) or a public/private key pair using RSA or ECDSA.
+
+**Providers** may include any metadata in the JWT they wish that helps to route, log, permission, or debug **agency** requests leaving their internal implementation flexible.
+
+JWT provides a helpful [debugger](https://jwt.io/#debugger). for testing your token and verifying security.
--- a/provider/auth.md
+++ b/provider/auth.md
@@ -1,17 +1,25 @@
 # Authorization
 
+MDS `providers` **SHALL** provide authorization for API endpoints via a bearer token based auth system. 
 
-MDS APIs SHALL provide authorization for **provider** API should be provided via a bearer token based auth system. 
+For example, the `Authorization` header sent as part of an HTTP request:
 
-For example, `Authorization: Bearer <token>` sent in the header of the request. More info on how to document [Bearer Auth](https://swagger.io/docs/specification/authentication/bearer-authentication/). 
+```
+GET /trips HTTP/1.1
+Host: api.provider.co
+Authorization: Bearer <token> 
+```
 
+More info on how to document [Bearer Auth in swagger](https://swagger.io/docs/specification/authentication/bearer-authentication/)
 
-RECOMMENDED : Use JSON Web Tokens or [JWT](https://jwt.io/introduction/) to generate tec
+## JSON Web Tokens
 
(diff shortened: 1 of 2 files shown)


migurski:
I’m not convinced that JWT should be a part of the MDS spec. From the standpoint of a mobility provider, JWT might be a great solution to the problem of token generation. From the perspective of an API consumer, it doesn’t seem relevant how a particular bearer token was generated as long as it can be used to gain access to a resource.

noonhub:
I somewhat agree with @migurski here. However, if we do not standardize on the format and how to retrieve the token, then I think the provider should be allowed to enforce any bearer token based authentication scheme.

Is the current proposal assuming that the provider generates this JWT and issues it to the agency manually? What is the TLL of the tokens?

If we do want to standardize on the format and interface to generate these tokens, OAuth 2.0 client_credentials grant type as laid out in [RFC 7523](https://tools.ietf.org/html/rfc7523) would be a good candidate for that standard.

I am in favor of leaving it up to the provider, but this means that the agency may need to implement some flow to retrieve a bearer token for each provider.

thekaveman:
Regardless of the particular implementation / token format, I think it's useful to have a single standard baked into MDS. Agencies aren't going to want to implement N different flows for N different providers. 

And on the flip side, when the `agency` API takes off, we'll hopefully have a corresponding flow for Providers which is straightforward for any/all Agencies they interact with.

migurski:
Not sure I agree @thekaveman. Even if JWT is chosen as a format it doesn’t specify the agency/provider token flow at all. One way to implement a token flow is via OAuth which places the onus on providers to develop a 3rd party developer key infrastructure. Even Github offers [personal access tokens](https://github.com/settings/tokens) as an external-to-OAuth ad hoc way to get tokens.

Specifying Authorization header bearer tokens in MDS makes sense, but the means by which those tokens are generated and shared seems like scope overreach at this stage.

asadowns:
@migurski and @noonhub I think your guys points are valid. I agree with @noonhub that ultimately as long as a bearer token is provided it doesn't make a huge difference what the format is.

All that said, JWT does seem like a good standard to cement around for provider and agency API. And specifying an auth method does have the benefit of providing clear instructions for providers as to what they should provide which helps avoid confusion. 

We haven't discussed any sort of refresh/login flow or endpoints so right now the TTL on the tokens we would provide to the city is one year. In order not to add the extra burden of a login/refresh flow to the agencies it does seem worthwhile to set a long TTL.

I don't necessary think it's overreach to provide info on an authorization method. But, I think it's very reasonable to make additional PRs to add more authorization methods. Thoughts, @thekaveman ?

thekaveman:
> Specifying Authorization header bearer tokens in MDS makes sense, but the means by which those tokens are generated and shared seems like scope overreach at this stage

I think this is what I tried (meant? wanted?) to say, but you said it much clearer @migurski 😄

I guess I mixed concepts - what I'm concerned about (as an agency that will be consuming provider data from 4 providers to start) is having a single type of auth method. I don't want to send a username/password to one, `Authorization: <token>` to another, SMS 2-Factor for another, blood sample for yet another... So any kind of clarification around *that* aspect is what I am looking for, and I think you said it correctly above. 

For the use-cases currently outlined (straight auth, plus tying geographic areas to specific agencies) I think JWT is compelling. But it does seem like more of a provider implementation detail than something specifically needed in the spec (and ditto for TTL) -- at least at this point, still so early in the game.

migurski:
++ @thekaveman, that’s great. Bearer tokens are a clean, simple way to do this. JWT could be offered as a suggestion on how to generate good ones.

asadowns:
Sounds like we're all in agreement on bearer tokens. Please feel free to merge / close / or request revisions to PR based on what you think belongs in the spec at this point. Thanks!

provider/auth.md
```diff
@@ -0,0 +1,11 @@
+# Authorization
+
+Authorization for **provider** API should be provided via a JSON Web Token or [JWT](https://jwt.io/introduction/).
```
oderby: Perhaps pivot this document away from discussing JWTs and towards specifying that authorization for the API will be communicated using Bearer Tokens, specified in the `Authorization` header, as detailed in [IETF RFC 6750](https://tools.ietf.org/html/rfc6750). Then a section can discuss how the spec does not dictate how those tokens are generated and transmitted to consumers, suggesting JWT as a common approach?

Issue #78, opened 2018-09-18
Don't use GET bodies for query parameters


oderby:
The current MDS specification relies on using the request body in GET requests as the mechanism for specifying the resource being requested. This is generally considered to be an anti-pattern (see discussion [here](https://stackoverflow.com/a/983458)). More importantly, it makes caching such requests impossible, as caches operate based on the request URI only.

Fixing this is simple - either move the request parameters from the body to the query string or don't use GET for those requests. The fix depends on the intent and desired semantics for the endpoint. Right now, I'd propose the following:
* Move the Provider API endpoints (`/trips` and `/status_changes`) to `POST` requests. These queries will need to accept complex parameters (specifically thinking about the geographic shapes for narrowing the query)
* Change the Agency API `/service_areas` endpoint to accept a query param `provider_id` and support requesting specific service areas by id. e.g. `GET /service_areas?provider_id=XXX` would return all the active service areas for a provider and `GET /service_areas/:id` (where `:id` is the id of a specific service area) would return the data for a specific service area

asadowns:
I would say moving the request parameters into a query string is probably the correct approach here.
I don't believe there is the need for complex geometric shapes if the base geometry for the particular MDS client is provided as part of the token based auth. 
Using a `POST` to retrieve data is often considered to be not semantic.

I would recommend keeping `/trips` and `/status_changes` as GET and clarifying around query parameters. I provided a potential approach in [#64](https://github.com/CityOfLosAngeles/mobility-data-specification/pull/64).

On the `agency` side I think it may be a bit early to deciding if service_areas need support for specific `provider_id` or an additional `/:id` route. I think both of these are good ideas to keep in mind moving forward but I don't think there should necessarily be an update now without a use case.

Let me know what you think!

thekaveman:
I think part of the confusion is, these query parameters aren't spelled out anywhere in the spec *yet*. But I also don't see where it is stated that MDS is expecting GET request bodies? That sounds crazy, I totally agree it is a bad idea @oderby!

I also tend to agree with @asadowns and @migurski here, these are very clearly GET requests (read-only, should be idempotent given the same parameters, etc.). @asadowns did some initial work on defining the query params over on #64 - can we move this discussion to that proposal so we can comment on some actual changes to the language in the spec?

(BTW I also really like the suggestion of tying the geo-area to the auth token (see #60) from @asadowns, thus obviating the need for that data on every request.)

[...]

PR #85, opened 2018-09-20
Clarify that /service_areas endpoint takes query parameters.

oderby:
Per discussion in #78, all endpoints for GET requests should only accept
request parameters via the query string in the URL. While #64 clarifies
this for the Provider API, this updates the only GET endpoint in the
Agency API.

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -194,12 +194,12 @@ Gets the list of service areas available to the provider.
 
 Endpoint: `/service_areas`  
 Method: `GET`  
-Body:
+Query Parameters:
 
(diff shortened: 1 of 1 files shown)


asadowns:
Looks good @oderby . I think if we merge in this and #64 it handles both the provider and agency query params based on our discussions.

oderby:
@thekaveman yeah, would love clarification on what is returned if no `service_area_id` is specified in the request. From my reading of the spec, it seems as though the plan is for there to be potentially multiple "active service areas", so I think the change is in line with that.

PR #86, opened 2018-09-20
Minor cleanup to formatting of unique_id in Agency API.

oderby:
Just some minor cleanup I noticed while working on #85

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -29,7 +29,7 @@ Body:
 
 | Field | Type     | Required/Optional | Other |
 | ----- | -------- | ----------------- | ----- |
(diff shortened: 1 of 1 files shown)


thekaveman:
Hopefully @toddapetersen and/or @hunterowens can weigh in on this, since I haven't been as involved in the agency discussions. 

It seems to me a further cleanup is to rename `unique_id` in agency to `device_id`, to mirror what provider calls it. Really the descriptions of the fields that are common to provider and agency should be the same as well, perhaps that is an additional cleanup (to use the descriptions from provider, which I think have been refined a bit more).

asadowns:
👍 for converting `unique_id` to `device_id` for consistency with provider API. Similar descriptive text around what the id represents would be helpful as well.

PR #90, opened 2018-09-26
add Provider entry for Lime

c-fang:


providers.csv
```diff
@@ -1 +1,2 @@
 provider_name, provider_id, url, mds_api_url
+Lime, 63f13c48-34ff-49d2-aca7-cf6a5b6171c3, li.me, data.lime.bike/api/partners/v1/mds/los_angeles
```
thekaveman: Is this `mds_api_url` the same for Santa Monica? @hunterowens weigh in, but I think we're looking for a generic URL here, where geo-filtering is done via query. Or is `los_angeles` meant to be more general and regional.

hunterowens: ^ should be more general I believe. 

Ie, the base url for MDS at Lime is  `data.lime.bike/api/partners/v1/mds/` I think? 

@c-fang how do you plan on supporting multiple cities on MDS? Will clarify for future revisions of spec.

c-fang: `data.lime.bike/api/partners/v1/mds` would be the base URL, and the next fragment in the URL is the city identifier. `los_angeles` is our internal city ID for "the city of Los Angeles," and we'll use other city IDs for other cities. Should I remove the city ID from `providers.csv`?

hunterowens: @c-fang yep, since this is a global listing.

PR #93, opened 2018-09-27
Clarifying timestamp format

### LINKED ISSUE #87: Unix Timestamp clarification in Provider API
Time fields like `start_time`, `end_time`, and route `timestamp` currently specify their type as `Unix Timestamp`. An example from route `timestamp` is `1531007628.3774529`. 

This leaves it open to interpretation and confusion what the correct format is. 

GBFS spec specifies for instance
> Integer POSIX timestamp

and provides an example `1434054678` or epoch seconds.

So using the example from route it should be either `1531007628` for epoch seconds. Or, if greater precision is required epoch milliseconds would be `531007628377`.

In either case it would be good to clarify what the preferred format for providing time is. If you let me know your preference, I'll put out a PR.

thekaveman:
Closes #87

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -73,6 +73,13 @@ References to geographic datatypes (Point, MultiPolygon, etc.) imply coordinates
 
 [Top][toc]
 
+### Timestamps
+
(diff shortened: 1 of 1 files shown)


hunterowens:
@thekaveman approved!

PR #103, opened 2018-10-01
Make `event_type_reason` conditional on `event_type`.

### LINKED ISSUE #97: Improve JSON Schema for event types and reasons
The spec defines these enums in somewhat of a hierarchy, e.g. each `event_type_reason` applies only to a specific `event_type`. 

However the current schemas don't enforce this hierarchy. This should be addressed going forward. 

Earlier discussion [here](https://github.com/CityofSantaMonica/mobility-data-specification/pull/1#discussion_r218648544) and [here](https://github.com/CityOfLosAngeles/mobility-data-specification/pull/53#discussion_r221112531).

ian-r-rose:
Fixes #97. No breaking changes, just a stricter schema matching event types and event type reasons.
Thanks to @dyakovlev for getting the ball rolling on a solution.

### DIFF
--- a/generate_schema/generate_provider_schema.py
+++ b/generate_schema/generate_provider_schema.py
@@ -5,6 +5,7 @@ When run, makes standalone schemas for the MDS provider API.
 """
 
 import json
+import jsonschema
 import requests
 
 POINT = "Point"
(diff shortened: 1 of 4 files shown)


hunterowens:
yeah, this seems like a `0.2.1` change. will add milestone.

thekaveman:
I think we're good to merge this, yes?

hunterowens:
LGTM

Issue #84, opened 2018-09-19
Expected QPS, latency, and data retention


noonhub:
Providing frequent and quick access to an ever growing set of data will obviously put pressure on provider's infrastructure, especially for new or smaller providers which may not have as sophisticated technology stacks. I think that we should establish explicit limits and expectations around QPS, latency, and data retention.

QPS/Load
 * What kind of QPS should providers expect on these APIs? 
 * As a provider, I expect we will want to enforce rate limits, what are acceptable limits?

Latency
 * What is an acceptable latency?
 * Should we consider asynchronous implementations for providers that cannot meet latency expectations easily?

Data Retention
 * What is the expected data retention? Indefinite?

oderby:
> * What is the expected data retention? Indefinite?

Judging from the material released from last week's workshop, it seems like the thought is about 1 year. (On slide 11 of [MDS-Presentation-Provider-Agency.pdf](https://drive.google.com/file/d/1-5nl8wH5aJcd5XcPHi1eARn-N1O1w8uH/view) it states "API must ... Contain and store 1 year of Data")

Regardless, the specification as documented here on Github is ambiguous and should be clarified. Additionally, it would be nice to agree on response code served from Provider APIs when the request is for a time range outside of the required range (and clarity on how to handle requests for which the provider only has partial information).

hunterowens:
Ref for #105, we should have an error code for "data no longer stored" 

On one hand, would love to standardize, on the other than, different jurisdictions == different retention schedules.

jsierles:
Is there a document we can refer to that clearly states the retention policy and expected response times for historical data in the Provider API?

Given the requirement of the new agency API, we're wondering if it makes sense to have a long retention period like one year. As providers grow, the arbitrary one-year cutoff can become increasingly difficult to support in real time.

For practical purposes today, 'd to know if currently we can set a shorter retention cut-off while this spec evolves and providers work on long term data warehousing strategies.

schnuerle:
The SLA/latency and other expectations from an agency could be added to the #646 Policy Requirements feature in an upcoming release, and provider specific details like historic data retention could be added to #628 in the future.

pierre-bouffort:
Maybe this issue is outdated? For SLAs related to latency, data retention etc... On our end what we do at Blue Systems is that we track those metrics (latency, data retention, etc...). And then cities/operators get to monitor that and look at the metrics and choose what they do with it. I am not sure the standard would need to include this?

PR #107, opened 2018-10-01
Clarify time parameter querying for status changes

black-tea:


black-tea:
ah, I see. Then I have follow up questions:

1) If start_time and end_time are referring to filters on time, shouldn't we clarify what the start_time and end_time are filtering on for the `trips` endpoint? Right now these filters have the same names as the fileds in `trips` Eg. Are we filtering on start_time or end_time? 

2) Do we want to add clarification on which comparison operators should be supported? Eg. do we automatically assume all of (these)[https://helpcenter.veeam.com/docs/backup/rest/query_comparison.html?ver=95] are supported?

thekaveman:
@black-tea I agree, more clarification for the query params sections for both Trips and Status Changes would be useful. 

Maybe something along the lines of:

> The `start_time` parameter filters for events occurring at or after the given time. The `end_time` parameter filters for events occurring at or before the given time. Using both filters for events within the range (including the bounds).

This language obviates the need for comparison operators, which I think are a bit overkill at this point, and leaves it to the submitter of the query to craft the range they want.

All of this is related to the issues raised in #84 around QPS and the question of *how far back in time* can one query a Provider API?

black-tea:
@thekaveman what do you think about the proposed change? For `trips` endpoint, `start_time` is a query on `start_time`, and `end_time` is a query on `end_time`. For `events`, both are queries on `event_time`.

PR #116, opened 2018-10-03
[Agency] API is REST-like, not RESTful

HenriJ:
I don't want to go into any bike-shedding, and it does not matter if an API is RESTful or REST-like. That said, it is nice to use the correct term to define the API, and this one is clearly REST-like (otherwise we would have a "/vehicles" resource that we could PUT and DELETE, not "register_vehicle" and "deregister_vehicle")

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -1,6 +1,6 @@
 # Mobility Data Specification: **Agency**
 
-This specification contains a collection of RESTful APIs used to specify the digital relationship between *mobility as a service* providers and the agencies that regulate them.
+This specification contains a collection of REST-like APIs used to specify the digital relationship between *mobility as a service* providers and the agencies that regulate them.
 
 * Authors: LADOT
 * Date: 10 Aug 2018


hunterowens:
@toddapetersen do we want to rework agnecy to be true restful or rest-like as is?

toddapetersen:
Yes.

HenriJ:
@toddapetersen "yes" meaning "rework to be true restful" ?
I have a little bit of spare time time in the next days, I could work on it with my team if you are open to it.

HenriJ:
Hi @toddapetersen , any news on that one ?
It would also help us if you could have a look on https://github.com/CityOfLosAngeles/mobility-data-specification/issues/148

lionel-panhaleux:
Hi @toddapetersen, @hunterowens. I submitted a PR for a more RESTful API and can provide adjustments on it if needed.

hunterowens:
@HenriJ Let's close this, in light of merging #194 which is inspired by this work.

PR #106, opened 2018-10-01
Add membership type field to the provider API

cttengsfmta:
Add membership type field to the provider API. We presume that more values could be added to the list of membership types.

PR #76, opened 2018-09-15
Access method

mattwigway:
Add an `access_method` to the trips file which details how a user unlocked a vehicle (see #71). I included all the ways I have ever unlocked a shared vehicle, as well as a few I've heard talked about (cash, call center, etc).

This will be particularly useful for understanding the use of pilots that allow people to access dockless mobility without smartphones.

PR #123, opened 2018-10-05
Add membership type field

cttengsfmta:
Add membership type field to the provider API. We presume that more values could be added to the list of membership types. (Reopening #106, as requested)

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -131,6 +131,7 @@ Response: See the [`trips` schema][trips-schema] for the expected format.
 | `parking_verification_url` | String | Optional | A URL to a photo (or other evidence) of proper vehicle parking |
 | `standard_cost` | Integer | Optional | The cost, in cents, that it would cost to perform that trip in the standard operation of the System |
 | `actual_cost` | Integer | Optional | The actual cost, in cents, paid by the customer of the *mobility as a service* provider |
+| `membership_type` | Enum | Optional | See [membership types](#membership-types) table |
 
 ### Trips Query Parameters
 
@@ -167,6 +168,15 @@ Gets all trips within that bounding-box where any point inside the `route` is in
 | electric_assist |
 | electric        |
 | combustion      |
+
+### Membership Types
+
+| `membership_type` |
+|--------------|
+| subscriber             |
+| subscriber_low_income  |
+| single_ride            |
+| single_ride_low_income |
 
 ### Routes
 


hunterowens:
@cttengsfmta this looks good. We will also want to change the corresponding `JSONSchema` definitions for our [validator](https://github.com/CityOfLosAngeles/mds-validator) codebases. Do you want to take a crack at that or should I?

hunterowens:
I wonder how similar #126 and #76 are.

cttengsfmta:
I don't see a way to add a new Enum in `generate_provider_schema.py`

alexdemisch:
To summarize the discussion from the 4/11 call, the current proposal would be to add the membership_type field, starting with the enum values proposed earlier.  This would be an optional field.  While this field does provide information related to the user of the trip, it is still generalized and not PII.  There are concerns that an enumerated list of values may not be able to capture quickly changing business models that may introduce new membership types, so the enum values would need to be actively managed.  An experimental prefix (e.g., “ex-“) could be used to allow new values not explicitly added to the list of enumerated values.

lionel-panhaleux:
To summarize discussion from 4/18 :
- The specification states that a specific `Content-Type` should be used for responses, cf. https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/provider#versioning

This means an `OPTIONS` request can be used for version negociation using the `Accept` header, cf. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept . Example: 

Ask for options, stating that version 0.2 is preferred to version 0.3 (using the q parameter):
```
OPTIONS /trips/ HTTP/1.1 
Host: provider.example.com 
Accept: application/vnd.mds.provider+json;version=0.2,application/vnd.mds.provider+json;version=0.3;q=0.9
```

Answer: Only version 0.3 available
```
HTTP/1.1 200 OK
Content-Type: mds.provider+json;version=0.3
```

hunterowens:
moving this to 0.4.1 as an optional field.

hunterowens:
@dirkdk to rewrite based on discussion on 1/16 into a separate reporting / equity notes.

schnuerle:
Would like to revisit this idea for adding this info directly to the Provider API, in addition to potentially adding to a Metrics API #486.

schnuerle:
This is being discussed as part of a subset of a proposed Metrics API with aggregated (not trip level) information about special groups in #569.

johnclary:
I'm strongly opposed to adding attributes to trip records that describe the users taking trips.

alexdemisch:
We've discussed some of this already in some of the Provider WG calls on Metrics, but I'd consider this specific proposal to add additional elements to the detailed trip data to be abandoned in favor of the aggregate "special groups" reporting in the Metrics #569.  

The only reason it would've be collected at a detailed level would be to roll it up anyway, so #569 is a better way to do this.

schnuerle:
Closing this in favor of work on #569.

PR #124, opened 2018-10-05
Standardize line breaks

cttengsfmta:
We were using the undocumented two-trailing-spaces method to do line breaks in some places. Line breaks were missing in other places. Trailing spaces are not handled well in some editors, and Git likes to whine about trailing whitespace. Replace with backslashes and add missing. Backslashes are also non-standard and undocumented, unfortunately.


hunterowens:
If this makes it easier to view /edit on Windows, SGTM.

PR #125, opened 2018-10-06
Fixing times to be more descriptive.

hunterowens:
In #52, we realized that trip_start, end and time aren't the best way to describe time. 

This pull request replaces trip_start, trip_duration, and trip_end with `start_time`, `reserved_duration` and `travel_duration` approach. 

this will be a breaking change.


hunterowens:
closing this PR in favor of the solution proposed by @babldev in #139

PR #135, opened 2018-10-11
Fixing newlines again

thekaveman:
And removing a duplicate line of text. And some wording cleanup while I'm at it.


hunterowens:
@thekaveman looks good to me. 

We should also build a markdown and JS/ Python linter as part of the PR process for this repo.

Issue #133, opened 2018-10-11
define times for query params 


hunterowens:
query param times should be defined. To mirror spec, probably unix timestamp

thekaveman:
I think we already specify this for the query params as:

> All of these query params will use the Type listed above with the exception of `bbox`, which is the bounding-box....

There might still be some work to cleanup the times, per #104

asadowns:
Adding two cents here. Are you open to using either Unix timestamp in seconds or unix timestamp in milliseconds, Python seems like an outlier in having a default time format that allows decimals and it results in much less pretty JSON formatting in most viewers (no scientific notation). Would you be open to maybe converting to unix timestamp in either seconds or milliseconds as an Int?

thekaveman:
@asadowns check out the issue I linked above, I think we're open to that, someone just needs to submit the PR!

asadowns:
Sure. I'll try and get to it this weekend.

Issue #104, opened 2018-10-01
Timestamp format


cttengsfmta:
Currently defined as "...the format returned by Python's time.time() function."  [The Python 3 spec](https://docs.python.org/3/library/time.html#time.time) says this function returns a float. [The precision and representation of a float in Python are platform-dependent](https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex).  Platform-dependent behavior in a specification that is likely to be implemented in many different platforms will lead to incompatible values and inconsistent results.

hunterowens:
TIL @cttengsfmta, that's a nice catch. 

Do you have a proposed switch? I can't imagine lack of second level-accuracy will head hurt in the provider API but let me know if you have a use case that would require more standardization.

cttengsfmta:
We think we should use the number of milliseconds since January 1, 1970 00:00:00 UTC.

Milliseconds because some implementation use seconds, and others use milliseconds. This has already caused confusion in the GBFS project:
https://github.com/NABSA/gbfs/issues/9

Also, many implementations that use seconds suffer from [the year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem).

Requiring milliseconds eliminates the confusion and makes it impossible to use a problematic implementation.

cttengsfmta:
We don't know whether you want to specify the handling of leap seconds or not. We would suggest leaving that up to implementors since a 37 second difference either way is not likely to be significant.

thekaveman:
Sounds great @cttengsfmta. 

The GBFS issue you linked to seemed to be around whether or not to use a more human-readable datetime format vs. POSIX timestamps. You aren't suggesting anything along those lines, are you?

This is more about moving away from `float` seconds and towards `bigint` milliseconds, right?

cttengsfmta:
Correct. Textual timestamp formats have other problems.

Issue #128, opened 2018-10-10
update JSON Schema to handle optional fields


hunterowens:
Running list of JSON validator fixes 

* Currently, the parking_verification_field in trips is required, not optional, causing my validator to fail. 
* Change `0.{version_number}.{anynumber}` regex for `version` to `{anynumber}.{anynumber}.{anynumber}-{anystring_or_null}` to more close match semvar

ian-r-rose:
On the second point: I agree that the last field should be `{anynumber}-{anystring-or-null}`. However, I think the first two should be kept the same: if an instance with major version `0.1.0` tries to validate against a schema of version `0.2.0`, then it should fail based on the version number alone. My understanding of semver is that, pre-1.0, every minor version is effectively a major version.

thekaveman:
I think the major concern @hunterowens raises with point two is, having to update that regex every time there is a new point release is one more thing to remember to do. And it would be a bad thing to forget.

hunterowens:
Additional thing to fix is that `actual_cost` and `real_cost` are optional, but current fail if null.

ian-r-rose:
Should `accuracy`, `trip_duration`, and `trip_distance` be integers? My ontological sense thinks they should be floating point numbers, but there may be some context that I am not aware of.

thekaveman:
@hunterowens:

> Currently, the parking_verification_field in trips is required, not optional, causing my validator to fail.

That field is not in the list of [required properties in the Trips schema](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/master/provider/trips.json#L232):

```json
 "trips": {
    "$id": "#/properties/data/properties/trips",
    "type": "array",
    "title": "The Trips Schema",
    "items": {
        "$id": "#/properties/data/properties/trips/items",
        "type": "object",
        "title": "The Trip Schema",
        "required": [
            "provider_name",
            "provider_id",
            "device_id",
            "vehicle_id",
            "vehicle_type",
            "propulsion_type",
            "trip_id",
            "trip_duration",
            "trip_distance",
            "route",
            "accuracy",
            "start_time",
            "end_time"
        ],
    }
}
```

However that property does have a [validation regex pattern](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/master/provider/trips.json#L350) that forces `HTTPS`, is that maybe the issue?

```json
"parking_verification_url": {
[...]

PR #138, opened 2018-10-13
Update timestamp desc to  unix timestamp milliseconds

asadowns:
Per #104 and #133 all timestamp references are unix timestamp in milliseconds.

Provide link with different language implementations.

Also, update all provider sample data to be in unix timestamps in milliseconds.

No changes made to agency.

Addresses: #104 and #133

### DIFF
--- a/generate_schema/common.json
+++ b/generate_schema/common.json
@@ -97,10 +97,10 @@
     },
     "timestamp": {
       "$id": "#/definitions/timestamp",
-      "title": "Integer seconds since Unix epoch",
+      "title": "Integer milliseconds since Unix epoch",
       "type": "number",
-      "multiple_of": 1.0, 
-      "minimum": 1483228800 
+      "multipleOf": 1.0, 
+      "minimum": 0 
     }
   }
 }
--- a/generate_schema/common.json
+++ b/generate_schema/common.json
@@ -97,9 +97,10 @@
     },
     "timestamp": {
       "$id": "#/definitions/timestamp",
-      "title": "Floating-point seconds since Unix epoch",
+      "title": "Integer seconds since Unix epoch",
       "type": "number",
-      "minimum": 0.0
+      "multiple_of": 1.0, 
+      "minimum": 1483228800 
     }
   }
 }
--- a/provider/status_changes.json
+++ b/provider/status_changes.json
@@ -146,10 +146,10 @@
     },
     "timestamp": {
       "$id": "#/definitions/timestamp",
-      "title": "Integer seconds since Unix epoch",
+      "title": "Integer milliseconds since Unix epoch",
       "type": "number",
-      "multiple_of": 1.0,
-      "minimum": 1483228800
+      "multipleOf": 1.0,
+      "minimum": 0
     },
     "vehicle_type": {
       "$id": "#/definitions/vehicle_type",
--- a/provider/status_changes.json
+++ b/provider/status_changes.json
@@ -47,12 +47,19 @@
       "properties": {
         "type": {
           "type": "string",
-          "emum": [
+          "enum": [
             "Feature"
           ]
         },
         "properties": {
-          "type": "object",
+          "oneOf": [
+            {
+              "type": "null"
+            },
+            {
+              "type": "object"
+            }
+          ],
           "required": [
             "timestamp"
           ],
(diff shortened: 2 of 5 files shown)


hunterowens:
in postgres 

```
CREATE TABLE IF NOT EXISTS test (

float_field float, 
int_field int
);

INSERT INTO test (float_field, int_field) VALUES (2.0, 2.0); 
``` 

runs successfully and if you pass `2.0,2.3` it will automatically round floats -> ints. 

So at least in the python/postgres DB world this in non-breaking. I don't think that will be universal but right now that is most provider ingest stacks so I am okay making the change. 

We should also update the JSONSchema, right @ian-r-rose?

ian-r-rose:
Yes, I think it should be updated.

I think that, strictly speaking, changing the type from a float to an integer would be considered a breaking change. Python is pretty lax in this regard, and Javascript doesn't even have the concept of integers, but if anybody is developing tooling in JVM-based or C-allied languages there could be some updates required.

As long as we are discussing changing data types, is there a reason the `trip_duration`, `trip_distance`, and `accuracy` are defined as integers rather than floats?

hunterowens:
@ian-r-rose can you move that discussion to #128. I'm tracking all JSON schema errors over there.

thekaveman:
So we're saying this is breaking then, and should be in `0.3.0`?

hunterowens:
@thekaveman 

I've been wrestling with this issue for a few days, and learning a whole bunch about how weird python and the concept of time can get. 

That being said, this seems like an important change, and I'm unaware of anybody building MDS ingest tooling in a language that doesn't silently downcast floats->ints, so I think we should tag this in 0.2.1. 

I'll update the milestone and the JSON schema to get this PR ready to merge.

hunterowens:
@asadowns can you give me permission to push to this branch? there are some changes to the JSON schema needed to support this PR. [Follow these steps](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

asadowns:
@hunterowens . Looks like it is already clicked but you should be good to go now.
image

hunterowens:
found the issue, was trying to push the @adownsbird  to the @asadowns branch. fixed now.

provider/status_changes.json
```diff
@@ -139,9 +146,10 @@
     },
     "timestamp": {
       "$id": "#/definitions/timestamp",
-      "title": "Floating-point seconds since Unix epoch",
+      "title": "Integer seconds since Unix epoch",
       "type": "number",
-      "minimum": 0.0
+      "multiple_of": 1.0,
+      "minimum": 1483228800
```
thekaveman: See comment above.

provider/trips.json
```diff
@@ -47,12 +47,19 @@
       "properties": {
         "type": {
           "type": "string",
-          "emum": [
+          "enum": [
             "Feature"
           ]
         },
         "properties": {
-          "type": "object",
```
thekaveman: Use array for `type`, see comment above

provider/trips.json
```diff
@@ -170,9 +177,10 @@
     },
     "timestamp": {
       "$id": "#/definitions/timestamp",
-      "title": "Floating-point seconds since Unix epoch",
+      "title": "Integer seconds since Unix epoch",
       "type": "number",
-      "minimum": 0.0
+      "multiple_of": 1.0,
```
thekaveman: `multipleOf`

provider/trips.json
```diff
@@ -170,9 +177,10 @@
...
+      "minimum": 1483228800
```
thekaveman: See comment above.

generate_schema/common.json
```diff
@@ -97,9 +97,10 @@
     },
     "timestamp": {
       "$id": "#/definitions/timestamp",
-      "title": "Floating-point seconds since Unix epoch",
+      "title": "Integer seconds since Unix epoch",
       "type": "number",
-      "minimum": 0.0
+      "multiple_of": 1.0, 
+      "minimum": 1483228800
```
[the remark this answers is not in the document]
hunterowens: that fair, I'm gonna say let's not enforce via schema but some sort of "truth-yness" validator.

hunterowens:
@thekaveman changes made.

thekaveman:
Closing in favor of #146

PR #139, opened 2018-10-13
Suggestion to improve time based filtering

babldev:
Opening this PR as a conversation starter about a possibly cleaner way to specify time bounds, or at least learn more about the nature of the queries mobility providers will be receiving.

**Issue**
I imagine cities will be querying the `/trips` Mobility Provider APIs on a regular frequency to pull data for the hour.

**Query 1:** 1pm - 2pm
**Query 2:** 2pm - 3pm
**Query 3:** 3pm - 4pm

Currently, this would be done via the `start_time` and `end_time` query parameters. However, this will result in duplicate rows across responses. For example, a ride from 1:45-2:15pm will show up in both Query 1 and Query 2. If cities are ingesting this data, this will create more work to deduplicate.

**Suggestion**
- Replace `start_time` + `end_time` for `/trips` with `min_end_time` and `max_end_time`. The start time will be in inclusive, the max will be exclusive.
- Modify `/status_changes` `end_time` parameter to indicate that it is exclusive and not inclusive for that time.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -143,8 +143,8 @@ The trips API should allow querying trips with a combination of query parameters
 
 * `device_id`
 * `vehicle_id`
-* `start_time`: filters for trips where `start_time` occurs at or after the given time
-* `end_time`: filters for trips where `end_time` occurs at or before the given time
+* `min_end_time`: filters for trips where `end_time` occurs at or after the given time
+* `max_end_time`: filters for trips where `end_time` occurs before the given time
 * `bbox`
 
 All of these query params will use the *Type* listed above with the exception of `bbox`, which is the bounding-box.
@@ -246,7 +246,7 @@ Data: `{ "status_changes": [] }`, an array of objects with the following structu
 The status_changes API should allow querying status changes with a combination of query parameters.
 
 * `start_time`: filters for status changes where `event_time` occurs at or after the given time
-* `end_time`: filters for status changes where `event_time` occurs at or before the given time
+* `end_time`: filters for status changes where `event_time` occurs before the given time
 * `bbox`: filters for status changes where `event_location` is within defined bounding-box. The order is definied as: southwest longitude, southwest latitude, northeast longitude, northeast latitude (separated by commas). 
 
 Example:


babldev:
@hunterowens @thekaveman

black-tea:
Hi @babldev  - I think this amendment makes a lot of sense. I had put together this initial proposal when I was thinking about modifiers, but without them I can see how the current spec would either exclude certain trips or bring them in multiple times. 

@hunterowens @thekaveman I do think this is important to address sooner than later, since providers are currently building in the time-based querying. Would this be able to be part of 0.2.1?

hunterowens:
@black-tea given that this change would be breaking, I would tag it as a requirement for a `0.3.0`

hunterowens:
That being said, I think this is a solid proposal, as it saves implementations from needing upsert logic. I've approved and thanks for the PR @babldev

babldev:
Thanks for the review!

thekaveman:
@oderby:

> So if I understand correctly, this would fundamentally change the behavior of the /trips endpoint, such that time-based queries no long return in-progress trips (trips that, as of the time of the query, have a start_time but no end_time)?

I don't agree that this is the current behavior -- Provider is all about things that *already* occurred, a Trip in progress by definition hasn't occurred (fully) yet. Without an `end_time` (and an `end_location` for that matter) the Trip wouldn't pass schema validation. Can you help me see where the spec implies that the `/trips` endpoint allows returning in-progress Trips?

This PR raises some interesting issues, since Trips don't necessarily start and end in the same hour. But again, since Provider is all about *historical* data, I agree that pinning time definitions to `end_time` -- meaning very definitely the Trip has ended -- is a great idea.

oderby:
@thekaveman ah, you are correct. I was off in my own world it seems in understanding the existing behavior of the `/trips` endpoint - I had missed the "historical" wording there. My apologies. I agree that being consistent and only dealing with completed trips makes sense. Perhaps we should update the wording to explicitly state the `/trips` endpoint should only return completed trips?

Nevertheless, I'd encourage still trying to find a way to avoid unnecessarily breaking endpoints between minor version changes. It seems here that could be easily achieved. Thoughts?

kyle-nycdot:
I'd like to bring up a subtle consideration, but first a quick introduction: as the username suggests, I'm Kyle (Yakal-Kremski), data analyst at NYC DOT Bike Share.

For what it's worth, populating trips only when they have completed results in behavior where relatively simple queries, like total trip volumes in a given time interval, may return different results depending on when they are requested. A trip starting is, in a sense, its own event, and that data is stored to be attributed to the trip entry when it is generated - before the trip has ended. Even though very long trips, or ones which never end because of the vehicle being disappeared or destroyed somehow, are rare, they do still happen and can affect accuracy of periodic reporting, for example.

Looking forward to seeing how this all develops and potentially helping out when I can.

thekaveman:
@kyle-nycdot welcome to the party! Thank you for your insights.

Are you familiar with the [`/status_changes`](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/master/provider#status-changes) endpoint that is also part of the Provider API? I think this is the place where the various "events" you mention: start of trip (independent of its end), long running trips, etc. would manifest.

`/trips` really is all about the geometry (and related attributes) of the full, complete trip.

kyle-nycdot:
@thekaveman Ah, got it. 

And thanks for the welcome!

ezheidtmann:
Glad to see this is fixed in 0.3.x. But my understanding of the 0.2.x behavior is different!

It seems to me that a ride from 1:45-2:15pm will *not* show up in Query 1 *nor* Query 2! Here's why:

 - In Query 1, the trip's `end_time` is too late.
 - In Query 2, the trips `start_time` is too early.

Maybe I'm missing something, but it seems likely that consumers of the 0.2.x API are unintentionally excluding trips due to this query parameter design.

babldev:
(FWIW, Lyft is already using the 0.3.x rules of dividing data by `end_time` only so data is bucketed by time cleanly)

ezheidtmann:
Awesome, thanks for verifying my thinking! 

And thanks for the info about your 0.2.x implementation -- to be clear, you're using `start_time` and `end_time` parameters but both apply to the `end_time` attribute?

babldev:
That's correct, effective min and max `end_time`.

`end_time` is an always-incrementing field based on the current timestamp, so it's the best choice given that data is not available for sharing until the ride is ended.

thekaveman:
> Maybe I'm missing something, but it seems likely that consumers of the 0.2.x API are unintentionally excluding trips due to this query parameter design.

Just to round out the conversation across multiple threads - this can of course be mitigated *today* in `0.2.x` by extending the query window greater than an hour. For example, query for the previous 3 hours, 6 hours, etc.

hyperknot:
> Just to round out the conversation across multiple threads - this can of course be mitigated _today_ in `0.2.x` by extending the query window greater than an hour. For example, query for the previous 3 hours, 6 hours, etc.

The problem today, is that you don't know how much you need to query back. For example, if you consider trips up to 1 day in length, you'll need to query as much as 24 hours backwards, and then discard 99% of that data, just to make sure that you get all trips within a given time frame.

This PR is really nicely thought out and I'm looking forwards to have it implemented by all providers.

Issue #129, opened 2018-10-10
Write up Release procedure


thekaveman:
There are manual and automated steps that must happen for each release.

In the interest of keeping everyone in the loop and making sure others can release if needed, a short document describing the release process would be beneficial.

dyakovlev:
A release checklist would be awesome. Related to that, some sort of published guidance on how MDS semver is intended to work would be helpful. (I'm looking to get an idea of answers to questions like, when are releases cut? when are versions deprecated? how is implementation of new versions required? is there a version negotiation process at request time? at some other time?)

hunterowens:
Shipped!

PR #147, opened 2018-10-23
Release guidelines

thekaveman:
For #129 

Early draft. The checklist is definitely not complete yet.

### DIFF
--- a/ReleaseGuidelines.md
+++ b/ReleaseGuidelines.md
@@ -60,30 +60,44 @@ release should not surprise anyone that wishes to be informed. To that end, rele
 
 * Be PR-driven, to give the community ample awareness and time for feedback
 
+* Be announced on the [`mds-announce`][mds-announce] Google Group
+
 More detail on this can be read in the Release Checklist below.
 
 ## Release Checklist
 
 The following steps **must** be followed for **every** release of MDS:
 
-1. Ensure the [Milestone][mds-milestones] for this release is at `100%`
+1. Ensure the [Milestone][mds-milestones] for this release is at `100%`.
 
-1. Run the schema generator to ensure the schema files are up to date
+1. Run the schema generator to ensure the schema files are up to date:
 
    ```console
    cd generate_schema/
    pipenv run python generate_provider_schema.py
    ```
 
-1. Update [`ReleaseNotes.md`](ReleaseNotes.md) following the existing format
+1. Update [`ReleaseNotes.md`](ReleaseNotes.md) following the existing format:
+
+    ```md
+    ## 1.2.3
+
+    > Released yyyy-MM-dd
+
+    High level summary of the release.
+
+    * Specific change referencing a PR [#555](https://github.com/CityofLosAngeles/mobility-data-specification/pull/555)
+
+    * Another change summary referencing a PR [#777](https://github.com/CityofLosAngeles/mobility-data-specification/pull/777)
+    ```
 
 1. [Open a PR][mds-pr-new] against `master`, comparing either `dev` (for a MINOR release) or a release branch (e.g. `0.2.x`) for a PATCH release.
 
     In the case of a new MINOR version, allow a minimum of 24 hours for community discussion and review of the PR.
 
-1. Once the PR has been sufficiently reviewed, `rebase merge` into `master`
+1. Once the PR has been sufficiently reviewed, `rebase merge` into `master`.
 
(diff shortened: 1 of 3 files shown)


thekaveman:
I think I'll call out "squash and merge" as the preferred option for merging a release (when merging into `master`), since this maintains a single clean reference to the release PR.

thekaveman:
In general for all PRs, yes.

But specifically when merging a new release (e.g. merging `dev` into `master`) I'm suggesting this is **always** done as a squash.

I think a rebase in this specific context has the potential to make following the changes betweeen releases more challenging. Whereas a squash points directly back to the PR, which then lists all the individual commits, comments, reviews, etc.

hunterowens:
huh. I kinda think the opposite on merging dev in master - that is should be a rebase, and not a squash, so that master codebase is compliant, and not destructive with dev. 

basically, if we squash on master, then there is a divergence the 2nd time around between master and dev as they will have split histories, since squash is destructive.

thekaveman:
I am trying to play it out all the way...

Let's assume for everything below that the current version is `0.3.0`, and `0.2.0` and `0.2.1` also exist:

---------------------------------------------------------------

**PATCH release**:

I'm guessing these will often come in pairs, especially if originating with the lower of the maintenance versions. For example, a patch from `0.2.1` to `0.2.2` might spur a corresponding patch from `0.3.0` to `0.3.1` (or at least will need to be considered for the next release to the higher maintenance versions).

1. rebase-merge `0.2.x` to `master`
2. tag master with the next patch version `0.2.2`
3. rebase `dev` onto `master`? *potentially messing with other work in-flight, open PRs, `0.3.x`, etc*

**MINOR release** e.g. `0.3.1` to `0.4.0`:

These may come with a final patch release in the lower version to clean up any outstanding work before it ceases to be part of the maintenance series. 

1. rebase-merge `dev` to `master`
2. tag master with the next version
3. cut next release branch (e.g. `0.4.x`)
4. remove prior maintenance series release branch (e.g. `0.2.x`)

-----------------------------------------------------------------------------

I'm hoping PATCHes would be minimal, because I don't know that I like the idea of having to rebase `dev` every time a PATCH goes out (3a. above)... it's the best I can think of at the moment though.

hunterowens:
I've gone ahead and tested the above pattern in a test-repo and am happy with it. 

The only additional thing that would be good to add is some sort of email listserv for release announcements. I've asked some folks in our department to set something up that anybody can join but only admins can post to.

ReleaseGuidelines.md
```diff
@@ -0,0 +1,84 @@
...
+* [Versioning](#versioning)
+* [Branch Mechanics](#branch-mechanics)
+* [Schedule](#schedule)
+* [Checklist](#release-checklist)
+
+## Versioning
+
+MDS uses [Semantic Versioning][semver]. Each release is associated with a [`git tag`][mds-tags] of the form `X.Y.Z`.
+
+Given that MDS is stabilizing under MAJOR version `0.x` right now, it should be assumed that MINOR version increments (e.g. `0.2.0` to `0.3.0`) are equivalent to MAJOR version increments and may contain breaking changes.
```
noonhub: Why? This just feels like we're ignoring semvar for now...

thekaveman: Not quite, see point 4: https://semver.org/#spec-item-4. 

I think we'd all obviously like to minimize the impact of version changes, so that's why we're building in time and awareness opportunities before making a new release. But the ability to clarify crucial questions that come up in this early stage (like *how is time represented*) is key to a longer-term success.

noonhub: Cool, I figured there was a term like that but I lazily did not read them all.

So we are totally adhering to to semvar, just calling out that per the semvar guidelines, v0.x is not stable and *may* introduce breaking changes.

thekaveman: Exactly that, @noonhub.

thekaveman:
Rebased on `0.2.x`

hunterowens:
created google group `mds-announce@lacity.org`, anybody can subscribe for updates. We should normalize some practice of sending emails to that group for more info.

noonhub:
Awesome! Can you provide the link to the actual google group to subscribe?

hunterowens:
@thekaveman here is a link to the [Google Group](https://groups.google.com/a/lacity.org/forum/#!forum/mds-announce)

Can you add "send email to mds announce" to the steps and then I think this is good to merge.

PR #156, opened 2018-10-30
RESTful responses using standard HTTP codes

lionel-panhaleux:
A proposal for issue https://github.com/CityOfLosAngeles/mobility-data-specification/issues/105

The idea is to provide more RESTful responses for the agency API.
Following the discussion on the issue, this is a specification proposal for well-structured errors in the 400 response. Error categories are largely inspired by Python error names.

Thanks in advance for your comments.


karcass:
#195 covers this

hunterowens:
fixed in #195. @lionel-panhaleux let us know if we didn't get anything you think needs to be covered in it, gonna close this PR for now.

PR #161, opened 2018-10-31
A more restful Agency API

lionel-panhaleux:
A proposal for a more RESTful Agency API.

This adresses issues https://github.com/CityOfLosAngeles/mobility-data-specification/pull/116 and https://github.com/CityOfLosAngeles/mobility-data-specification/issues/63.

All fields are kept, this is just some moving around on endpoints and HTTP methods.
The `provider_id` and `unique_id` fields are systematically in the path (endpoint URL).
I took the liberty to align `propulsion_type` and `vehicle_type` on the provider API definitions in a separated commit.

About responses, I treated this part in a separated PR : https://github.com/CityOfLosAngeles/mobility-data-specification/pull/156 .

Note that this PR highlights a possible confusion between the `vehicle_id` field for vehicle registration (containing the VIN) and the `vehicle_id` field for starting a new trip, supposed to contain the unique vehicle ID used at registration (not the VIN). I wanted to avoid renaming fields as much as possible in the PR, but renaming at least the VIN to something else (ie. `identification_number`) could be more clear.

Comments and insights are welcome.

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -8,30 +8,34 @@ This specification contains a collection of RESTful APIs used to specify the dig
 
 ## Table of Contents
 
-* [register_vehicle](#register_vehicle)
-* [deregister_vehicle](#deregister_vehicle)
-* [update_vehicle_status](#update_vehicle_status)
-* [start_trip](#start_trip)
-* [end_trip](#start_trip)
-* [update_trip_telemetry](#update_trip_telemetry)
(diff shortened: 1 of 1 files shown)


hunterowens:
Thanks @lionel-panhaleux! This PR is similar to the changes in #194, so we are going to use that moving forward. If there are any changes that should be in Agency that we missed from this PR, let me know.

Issue #168, opened 2018-11-07
[agency] Consider using GEOJSON features


sebdiem:
A few GEOJSON objects are used in the agency API:
- Point in update_vehicle_status and start/end_trip
- MultiPolygon in service_areas

Wouldn't it be more flexible to take advantage of GEOJSON features to enable attaching properties to the GEOJSON object ?
- Point could be replaced by a Feature.
- MultiPolygon could be replaced by a Feature Collection.

This would enable, for instance, attaching a pricing rate as a property to the geometry of a service_area. Or attach the accuracy to the location.

What do you think ?

hunterowens:
@sebdiem this seems like a good idea! I'm wrapping it with #169, which would add JSON Schema support in agency and as something to target for the 0.3.0 release.

karcass:
See recent changes and updates to agency.  GeoJSON has largely been eliminated, because posting the entire route has been eliminated.  GeoJSON is not applicable for vehicle registration, and does not add anything useful for events or telemetry.

The service_areas endpoint will use GeoJSON to match provider.

oderby:
Should we close this issue now @hunterowens? Per @karcass comment above, this doesn't seem relevant anymore.

toddapetersen:
Agree with @oderby.  No longer relevant with updated agency.

PR #172, opened 2018-11-08
Agency schema

### LINKED ISSUE #169: JSON Schema for Agency
Currently, there is no JSON schema for agency. We should build out.

ian-r-rose:
Fixes #169. Partial fix for #168.

I have tried to hew as closely as I could to the description in the README, with a couple of exceptions for things I thought were mistakes or unclear:
* Should `reason_code` in the update vehicle status API instead be the same as `event_type_reason` from the `status_change` API? It seemed to me that it should from the narrative description.
* I have gone ahead and changed several of the points to be `MDS_Feature_Point`s as suggested in #168.
* `vehicle_type` and `propulsion_type` seemed out of sync with the provider API. I updated them.

These schema are as yet not very well tested, and I am not aware of any synthetic data to put them through the works. Do you have anything you could point me towards @hunterowens or @thekaveman?
Finally, I notice that the spec does not include `version` fields in them, as is the case for `provider`. Should we consider adding that?

I have not added schema for any of the agency response types at the moment.

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -83,7 +83,7 @@ Body:
 | `timestamp` | Unix Timestamp | Required | Date/time that event occurred. Based on GPS clock. |
 | `location` | Point | Required | Location at the time of status change in WGS 84 (EPSG:4326) standard GPS projection  |
 | `event_type` | Enum | Required | [Event Type](#event_type) for status change.  |
-| `reason_code` | Enum | Required | [Reason](#reason_code) for status change.  |
+| `event_type_reason` | Enum | Required | [Reason](#event_type) for status change.  |
 | `battery_pct` | Float | Require if Applicable | Percent battery charge of device, expressed between 0 and 1 |
 
 Response:
@@ -106,7 +106,7 @@ Body:
 | `provider_id` | String | Required | Issued by city |
 | `vehicle_id` | String | Required | Provided by the Vehicle Registration API |
 | `timestamp` | Unix Timestamp | Required | Date/time that event occurred. Based on GPS clock. |
-| `location` | Point | Required | Location at the time of status change in WGS 84 (EPSG:4326) standard GPS projection  |
+| `location` | GeoJSON [Point Feature](#geographic-data) | Required | Location at the time of status change in WGS 84 (EPSG:4326) standard GPS projection  |
 | `accuracy` | Integer | Required | The approximate level of accuracy, in meters, represented by start_point and end_point. |
 | `battery_pct_start` | Float | Require if Applicable | Percent battery charge of device, expressed between 0 and 1 |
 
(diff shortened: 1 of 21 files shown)


ian-r-rose:
Ooh, I see this will conflict with #161. Happy to deal with those conflicts however you see fit.

hunterowens:
We'll have to rebase and redo this after #194 

Let's hold off on doing this until #195 #196 and #197 are done.

hunterowens:
@karcass yes, and #203 if we go that route.

hunterowens:
that being said, I don't think it would be a bad idea to get JSON schema definitions of the expected return values for an Agency API for now. We can ignore the `/service_areas` endpoint.

karcass:
Agreed, let me put a schema on the agency to-do list.  It should be super simple.  Register just reads back the vehicle it got, and `/event` and `/telemetry` basically just say "got it, thanks!"  :)

hunterowens:
hmm yeah. 

perhaps it is a good idea to have valid schema definitions for both the `inputs` and `outputs` for an Agency API. Ie, what is valid to submit / and recieve. 

To utilize the common datatype, you can see how we generate the official schemas in `generate_schema/generate_provider_schema.py` @ian-r-rose may be able to provide additional direction.

karcass:
I will likely require some guidance to be efficient / effective.

hunterowens:
@karcass quick question - you guys did swagger / openAPI specs for the reference implementation, or I am not remembering that correctly.

karcass:
@hunterowens I'm evaluating options.  Swagger seems like it's more suited for start-with-swagger (code generation etc.) rather than bolt-on-afterwards, and I didn't start with it.  I am tentatively leaning towards using `apidoc` to extract my JavaDoc-style comments in the ref impl into doc files.  Do you strongly favor swagger, is apidoc (or similar) acceptable?

toddapetersen:
Closing this as the PR is no longer inline with the latest version of Agency spec.  We will revisit the issue (#169 ) to generate a schema once the Agency sandbox evaluation period has wrapped up.

ian-r-rose:
I'm happy to update this once you feel like outstanding changes to the agency spec have been resolved (#197, #203, perhaps others?). Do you have any comments on my questions above, especially regarding `status_change`?

We may also want to migrate to openapi, as suggested above, though my inclination is to do things one-at-a-time.

karcass:
@ian-r-rose If you're offering to do my job for me, I graciously accept!  I didn't think it was fair for us to rejigger everything and have you keep chasing a moving target.  

Yes, `agency` has deliberately reframed `event_type` and `event_type_reason` as `state` and `event_type`.  I apologize for the confusion, and I'm not sure how to fully reconcile `agency` and `provider` (and at present, perfect alignment is not a hard requirement).  In the context of the agency API, we (various E&A folks, coding the first agency implementation) reached the conclusion that a state-machine representation of edges (events) and nodes (states) was closer to the ground-truth of what we were trying to capture.  So, that's where we're at.  

Happy to answer any additional questions.

karcass:
@ian-r-rose do you plan to update this PR?

ian-r-rose:
Closing in favor of #238

PR #179, opened 2018-11-25
Dev timestamp

hunterowens:
This PR is a replica of #146, which is being reverted from the `0.2.x` branch since it is a breaking change. This PR will move all that stuff to `dev`, prior to change. 

PR to revert all the stuff from `0.2.x` to follow.


hunterowens:
also we should squash and merge this.

generate_schema/generate_provider_schema.py
```diff
@@ -53,6 +53,7 @@ def get_feature_schema(id=None, title=None, geometry=None, properties=None, requ
 
     f_properties = feature["properties"]["properties"]
     if required is not None:
+        del f_properties["oneOf"]
```
thekaveman: I think this change makes sense in `0.2.x`

provider/status_changes.json
```diff
@@ -47,12 +47,11 @@
       "properties": {
         "type": {
           "type": "string",
-          "emum": [
```
thekaveman: I fixed these in #180 so as not to re-introduce the typo...

provider/trips.json
```diff
@@ -322,25 +322,20 @@
...
-                ]
+                "$ref": "#/definitions/timestamp"
               },
               "parking_verification_url": {
                 "$id": "#/properties/data/properties/trips/items/properties/parking_verification_url",
-                "type": "string",
+                "type": [
+                  "string",
+                  "null"
+                ],
```
thekaveman: These changes should all be in `0.2.x` too, just clarifying what `null` means and what's allowed.

PR #183, opened 2018-11-27
Remove unused bounding box parameter from Provider API

babldev:
**Issue**
The `bbox` parameter exists so that Agencies querying the Provider API can filter for rides in their operating area.

Geo-based filtering is expensive for Providers to support (geo-based index), and based on current traffic from agencies it is not even being used.

**Proposal**
Since the `bbox` parameter isn't being used, let's remove it to keep the Provider API as focused as possible.

Instead of making the bounding box part of the request, Provider + Agency should agree upon bounding box in advance and associate it with their credentials. That way, providers can optimize for the optimal way to index without needing to maintain a complex geo index.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -143,17 +143,6 @@ The trips API should allow querying trips with a combination of query parameters
 * `vehicle_id`
 * `min_end_time`: filters for trips where `end_time` occurs at or after the given time
 * `max_end_time`: filters for trips where `end_time` occurs before the given time
-* `bbox`
-
-All of these query params will use the *Type* listed above with the exception of `bbox`, which is the bounding-box.
-
-For example:
-
-```
-bbox=-122.4183,37.7758,-122.4120,37.7858
-```
-
-Gets all trips within that bounding-box where any point inside the `route` is inside said box. The order is defined as: southwest longitude, southwest latitude, northeast longitude, northeast latitude (separated by commas).
 
 ### Vehicle Types
 
@@ -243,13 +232,6 @@ The status_changes API should allow querying status changes with a combination o
 
 * `start_time`: filters for status changes where `event_time` occurs at or after the given time
 * `end_time`: filters for status changes where `event_time` occurs before the given time
-* `bbox`: filters for status changes where `event_location` is within defined bounding-box. The order is definied as: southwest longitude, southwest latitude, northeast longitude, northeast latitude (separated by commas). 
-
-Example:
-
-```
-bbox=-122.4183,37.7758,-122.4120,37.7858
-```
 
 ### Event Types
 


babldev:
@hunterowens @asadowns

asadowns:
This makes sense to me. `bbox` doesn't seem like it provides any value given agencies already get data for their entire municipal area. It also avoids tricky situations where the requested `bbox` may be outside or partially outside of the city's jurisdiction. 

Not using the `bbox` param also allows much more accurate results. Currently, we provide data via MDS based on detailed polygons of the agency area which is simply not possible with `bbox`. 

I agree that agency geometry should be associated to credentials.

dyakovlev:
>  so I am wondering if any of them are seeing a need to have bbox, or they will rely on the client credential flow described in #143 to do authentication for the data they access?

We've not really used the bbox parameter so far. As long as tokens are city-scoped by providers, our approach seems like it will work. One interesting question we haven't really explored fully is what happens to devices that enter/exit a city's service area bbox, or move between those of adjacent cities (like, ahem, LA and Santa Monica).

asadowns:
After some experimentation and false starts, we feel that time-based pagination is the most efficient with the caveat that you keep the time windows relatively small (<5 minutes) and paginate. For larger periods of time it might be worth discussing some sort of bulk historical aggregate endpoint.

As for @dyakovlev question my feeling is a reasonable expectation from the city's perspective is to receive all trips that start or end inside their city as well as all status changes that occur in their city.

To add some concreteness to this discussion, this is the Polygon we currently use to represent Los Angeles for return status_changes and trips is the following. As you see it allows a lot more granular definition of what constitutes Los Angeles than a bounding box.

http://geojson.io/#id=gist:asadowns/26f95826695ce2f3b101a36655a2dede&map=11/34.0211/-118.4117

babldev:
From @hunterowens 
> This additionally makes sense to me as well, assuming that all relationships are city agency <-> provider. I know that a few cities are starting to use aggregators, (ie, Remix, RideReport, Populus), so I am wondering if any of them are seeing a need to have bbox, or they will rely on the client credential flow described in #143 to do authentication for the data they access?

Even for `Provider <-> Aggregator` data sharing I think it makes sense to agree upon what data will be included/excluded in advance and associate it with the credentials. The two parties can decide what bounding boxes to include, what ride types to include, inclusion/exclusion of rides that intersect the region but don't start/end there, etc. Then the provider starts collecting time series data in the standard format.

> Also would love thoughts re scaling on how best to cut data as volume (ie, num of devices) goes. It seems like time best pagination is the most efficient but curious if others have opinions.

IMO time bounds as we use now are the way to go.

hunterowens:
Yeah, I, I think that depends on trip _ids_ being consistent, since at least under our regulatory guidance, we should see trips that for example, started in SM, went through LA, and ended in Culver City. I don't think there is any constraint on the companies that trip_id must remain constant across jurisdictions but would be odd to see it that way. 

Anyways, I think this is good to merge for an 0.3.0

hunterowens:
@johnclary No discussion on that as of yet, what would be the use case?

johnclary:
@hunterowens e.g., I'm an agency who does not wish to retrieve the `route` data. If I could request trips without this field it would significantly reduce the amount of the data sent over the wire. 

Not sure if this would address scaling concerns in a meaningful way.

PR #196, opened 2018-12-12
Support batch telemetry

toddapetersen:
The submission of telemetry data is now provided on `POST /vehicles/telemetry` to allow for batch telemetry data upload.  It will still support single telemetry point submission, but also across an entire fleet of active vehicles or data for an entire trip.  The goal is to make it easier and more flexible for Providers to submit data and to design to support larger scale datasets.  Telemetry data no longer requires a trip_id and will be automatically associated with trips based on timestamps.

The telemetry data for this endpoint is now structured as a simple JSON object and not GeoJSON.  GeoJSON will remain for the GET (retrieval) endpoints.


hunterowens:
@toddapetersen can you have this PR target `dev` now?

toddapetersen:
Conflicts resolved.  Review and approve merge.

agency/README.md
```diff
@@ -222,6 +175,23 @@ List of valid vehicle events and the resulting vehicle status if the event is su
...
 | `deregister`           | A vehicle is deregistered                                                                            | `available`, `unavailable`, `removed`, `elsewhere` | `inactive`          | A vehicle is deactivated from the fleet and unavailable.                |
 
+## Telemetry Data
+
+A standard point of vehicle telemetry. References to latitude and longitude imply coordinates encoded in the [WGS 84 (EPSG:4326)](https://en.wikipedia.org/wiki/World_Geodetic_System) standard GPS projection expressed as [Decimal Degrees](https://en.wikipedia.org/wiki/Decimal_degrees).
+
+| Field          | Type           | Required/Optional     | Field Description                                            |
+| -------------- | -------------- | --------------------- | ------------------------------------------------------------ |
+| `vehicle_id`   | UUIDv4         | Required              | ID used in [Register](#vehicle-register)                     |
+| `timestamp`    | Unix Timestamp | Required              | Date/time that event occurred. Based on GPS clock            |
```
sebdiem: if it's from the GPS shouldn't it be  `gps.timestamp` ?

karcass: I believe the rationale is that we may at some point have non-gps telemetry, either from a vehicle, or from external sensors that aren't necessarily GPS-based.  Also it's nice to have `timestamp` at the top level of the object, as with events.

agency/README.md
```diff
@@ -222,6 +175,23 @@ List of valid vehicle events and the resulting vehicle status if the event is su
...
+| `gps`          | Object         | Required              | Telemetry position data                                      |
+| `gps.lat`      | Double         | Required              | Latitude of the location                                     |
+| `gps.lng`      | Double         | Required              | Longitude of the location                                    |
+| `gps.altitude` | Double         | Required              | Altitude above mean sea level in meters                      |
+| `gps.heading`  | Double         | Required              | Degrees - clockwise starting at 0 degrees at true North      |
+| `gps.speed`    | Float          | Required              | Speed in meters / sec                                        |
+| `gps.accuracy` | Integer        | Required              | GPS accuracy value                                           |
+| `charge`       | Float          | Require if Applicable | Percent battery charge of vehicle, expressed between 0 and 1 |
```
sebdiem: we could imagine grouping "energy" properties using an `energy` key.
E.g.
`energy.charge`,
`energy.autonomy` (the number of miles / kilometers)
...

agency/README.md
```diff
@@ -101,68 +102,29 @@ Body Params:
...
-                    "type": "Point",
-                    "coordinates": [
-                        -118.464851975441,
-                        33.990366257735
-                    ]
-                }
-            }
-        ] }
-```
+400 Failure Response:
```
sebdiem: This should probably not be part of a specific section but rather be defined globally for the agency API.

agency/README.md
sebdiem: This section https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/provider#timestamps could also be added to this spec to clarify what timestamp is.

agency/README.md
```diff
@@ -222,6 +175,23 @@ List of valid vehicle events and the resulting vehicle status if the event is su
...
+| `vehicle_id`   | UUIDv4         | Required              | ID used in [Register](#vehicle-register)                     |
+| `timestamp`    | Unix Timestamp | Required              | Date/time that event occurred. Based on GPS clock            |
+| `gps`          | Object         | Required              | Telemetry position data                                      |
+| `gps.lat`      | Double         | Required              | Latitude of the location                                     |
+| `gps.lng`      | Double         | Required              | Longitude of the location                                    |
+| `gps.altitude` | Double         | Required              | Altitude above mean sea level in meters                      |
+| `gps.heading`  | Double         | Required              | Degrees - clockwise starting at 0 degrees at true North      |
+| `gps.speed`    | Float          | Required              | Speed in meters / sec                                        |
+| `gps.accuracy` | Integer        | Required              | GPS accuracy value                                           |
```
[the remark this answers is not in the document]
karcass: Yes, it should be a float.  It is meant to be the NMEA GGA HDOP value.
We should probably add a `satellites`(int) field too.

agency/README.md
```diff
@@ -238,6 +202,24 @@ List of valid vehicle events and the resulting vehicle status if the event is su
...
 | `deregister`           | A vehicle is deregistered                                                                            | `available`, `unavailable`, `removed`, `elsewhere` | `inactive`          | A vehicle is deactivated from the fleet and unavailable.                |
 
+## Telemetry Data
+
+A standard point of vehicle telemetry. References to latitude and longitude imply coordinates encoded in the [WGS 84 (EPSG:4326)](https://en.wikipedia.org/wiki/World_Geodetic_System) standard GPS projection expressed as [Decimal Degrees](https://en.wikipedia.org/wiki/Decimal_degrees).
+
+| Field          | Type           | Required/Optional     | Field Description                                            |
+| -------------- | -------------- | --------------------- | ------------------------------------------------------------ |
+| `vehicle_id`   | UUIDv4         | Required              | ID used in [Register](#vehicle-register)                     |
```
sebdiem: In register there is `device_id` which is a UUIDv4 but `vehicle_id` is a String.
I suppose you meant `device_id` ?

agency/README.md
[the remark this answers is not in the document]
toddapetersen: Changed to float/hdop and added int/satellites.

agency/README.md
toddapetersen: Let's add to future PR with some additional clarification needed based on feedback from what data providers have available.

agency/README.md
toddapetersen: The responses are specific to each endpoint.  The codes are global.

agency/README.md
toddapetersen: Will add to future PR to align with provider spec in general.

agency/README.md
toddapetersen: Thanks for catching.  Fixed.

PR #197, opened 2018-12-12
Add dynamic properties for service areas

toddapetersen:
Adding the concept of a properties object for service areas.  The idea is that this will provide an extensible way to attach dynamic properties to returned service areas that is specific to the Provider.  The first proposal is a vehicle quota to let Providers know how many vehicles they can deploy in service areas available to them.


agency/README.md
```diff
@@ -164,30 +164,46 @@ The route must include at least 2 points, a start point and end point. Additiona
...
+| `end_date`         | Unix Timestamp                      | Optional          | If exists, Date at which this service area was replaced.                                    |
+| `area`             | MultiPolygon                        | Required          | GeoJson [MultiPolygon](https://tools.ietf.org/html/rfc7946#section-3.1.7) in WGS84 degrees. |
+| `prev_area`        | UUIDv4                               | Optional          | If exists, the UUID of the prior service area.                                              |
+| `replacement_area` | UUIDv4                               | Optional          | If exists, the UUID of the service area that replaced this one                              |
+| `type`             | Enum                                | Required          | See [area types](#area-types) table                                                         |
+| `properties`       | [Area Properties](#area-properties) | Optional          | Additional properties of a service area                                                     |
```
hunterowens: This PR removes the changes made in #110. Please rebase and resubmit.

agency/README.md
```diff
@@ -164,30 +164,46 @@ The route must include at least 2 points, a start point and end point. Additiona
...
-| ----- | ---- | ----------------- | ----- |
-| `service_area_id` | UUID | Required |  |
-| `service_start_date` | Unix Timestamp | Required | Date at which this service area became effective |
-| `service_end_date` | Unix Timestamp | Required | Date at which this service area was replaced. If currently effective, place NaN |
-| `service_area` | MultiPolygon | Required | |
-| `prior_service_area` | UUID | Optional | If exists, the UUID of the prior service area. |
-| `replacement_service_area` | UUID | Optional | If exists, the UUID of the service area that replaced this one |
+| Field   | Types   | Required/Optional | Field Description                                               |
+| ------- | ------- | ----------------- | --------------------------------------------------------------- |
+| `quota` | Integer | Optional          | Number of allowable vehicles that can be deployed by Provider. |
```
hunterowens: This doesn't differentiate between vehicles in motion and vehicles . We should probably remove this.

hunterowens: If we want to implement vehicle max and vehicle min, we need to be more thoughtful about it.

HenriJ: As we want to be able to implement richer rules that just per-provider quota (time-based rules, dependent on the type of vehicle, ... and that could trigger a full-featured pricing policy instead of just restricted/unrestricted/preferred), it might be more future-proof to have 2 endpoints: a `/areas` that exposes the areas like today and a `/policies` that exposes the list of policies that are attached to areas and other rules to trigger specific policies. What do you think ?

hunterowens:
@toddapetersen can you have this PR target dev now?

agency/README.md
hunterowens: @HenriJ yeah, so sort of `rules` model makes sense. I am thinking that we should tie to it to the event type statuses, so ie, you can have allowed status_changes, and not_allowed.

toddapetersen:
Conflicts resolved.  Review and approve merge.

karcass:
@toddapetersen this has conflicts again, alas

PR #198, opened 2018-12-13
Change Lyft route to specify region.

rowillia:
As part of some refactoring we're moving some routes around to include the region they're serving (similar to GBFS - https://github.com/NABSA/gbfs/blob/master/systems.csv).  The old route will continue to work for SMO/LAX, but ideally we'd migrate over in the near future.

### DIFF
--- a/providers.csv
+++ b/providers.csv
@@ -3,5 +3,5 @@ JUMP,c20e08cf-8488-46a6-a66c-5d8fb827f7e0,https://jump.com,https://api.uber.com/
 Lime,63f13c48-34ff-49d2-aca7-cf6a5b6171c3,https://li.me,https://data.lime.bike/api/partners/v1/mds
 Bird,2411d395-04f2-47c9-ab66-d09e9e3c3251,https://www.bird.co,https://mds.bird.co
 Razor,6ddcc0ad-1d66-4046-bba4-d1d96bb8ca4d,https://www.razor.com/share,https://razor-200806.appspot.com/api/v2/mds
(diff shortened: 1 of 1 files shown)


hunterowens:
@rowillia 

Since the `systems.csv` represents a global list of MDS systems/provider_ids, I would recommend that we keep the listing in providers.csv the same as currently, even if to make full requests you will need the path info. 

Some implementations may require a postfix/prefix (ie, LAX, los-angeles) to form correct URLs. Currently, that is support inside the `.config` of the [MDS provider library](https://github.com/cityofsantamonica/mds-provider).

thekaveman:
It's a little confusing (any suggestions for how to improve this are welcome) But basically you can configure a `Provider` instance with an object of config data here https://github.com/CityofSantaMonica/mds-provider/blob/dev/mds/providers.py#L37

And the `ApiClient` looks for this when constructing the URL here: https://github.com/CityofSantaMonica/mds-provider/blob/dev/mds/api/client.py#L40

Definitely non-standard behavior. Maybe we should consider if this can be built into the spec.

rowillia:
@thekaveman Yeah, I'm happy to propose something.  If MDS picks up in popularity we're going to want to segment endpoints by region (similar to how most GBFS feeds operate) to control who has access to what data.

hunterowens:
@rowillia Yeah, happy to add this into the core MDS if wanted to providers.

hunterowens:
Additionally, you can control which users have access to which pieces of information via the token mechanism, JWT additionally allows you to encode more information in the token.

hunterowens:
closing this for now.

Issue #88, opened 2018-09-22
Clarification of `associated_trips` in Provider API Status Changes endpoint?


oderby:
I'm curious to hear more about the `associated_trips` field in a status change. My questions are:
* Why/when would a single reservation event have multiple trips related to it?
* What additional information does this field provide that you couldn't determine by joining with trips on device_id and event_time?

thekaveman:
I'm reopening this issue because I'm not convinced it was adequately solved, as @oderby points out. 

Is there a case where multiple values should show up in this array? I don't quite understand that yet. If not, can we add to `0.2.1` and make this an optional `UUID` as suggested by @oderby. It's no more breaking in my mind than #146, which has already been slated for inclusion...

black-tea:
I have seen a provider add data into this field (when it applies). However, I have never seen two trip IDs for that field and I am not sure the circumstances we would see more than one.

black-tea:
@hunterowens One additional clarification question: Right now, under the Required/Optional, the field is listed as "Optional based `event_type_reason`". Would a provider be in compliance if it never provided the `associated_trips` field? 

My understanding is that the field is required for user-generated event types? If so, would it be more clear to be listed as 'Required if applicable based on `event_type_reason`', with the requirement being any user-generated `event_type`? Or perhaps I am mistaken?

hunterowens:
Closed with #217

PR #202, opened 2018-12-28
Associated Trip Fix

black-tea:
This breaking change addresses #88 for the `associated_trip` field in the `status_changes` endpoint.

Changes:
- Clarifies that associated trip is singular, not plural in provider README.md
- Modifies field name (associated_trips -> associated_trip) in JSON Schema and README.md
- Modifies JSON Schema from optional to required if applicable using implication logic
- Modifies JSON Schema type from array of UUIDs to single UUID

### DIFF
--- a/provider/status_changes.json
+++ b/provider/status_changes.json
@@ -300,6 +300,28 @@
                 }
               }
             ],
+            "anyOf": [
+              {
+                "not": {
+                  "properties": {
+                    "event_type_reason": {
+                      "enum": [
+                        "user_pick_up",
+                        "user_drop_off"
+                      ]
+                    }
+                  }
+                } 
+              },
+              {
+                "properties": {
+                  "associated_trip": {
+                    "description": "For user-generated event types, associated trip_id",
+                    "$ref": "#/definitions/uuid"
+                  }
+                }
+              }
+            ],
             "properties": {
               "provider_name": {
                 "$id": "#/properties/data/properties/status_changes/items/properties/provider_name",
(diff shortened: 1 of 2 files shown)


provider/status_changes.json
```diff
@@ -300,6 +300,28 @@
                 }
               }
             ],
+            "anyOf": [
```
oderby: Is this just `anyOf` instead of `oneOf` because you can't specify `oneOf` twice for the same object? Not to get too far into the weeds, but wouldn't it be more semantically correct to structure this whole sub-schema as a single `allOf` with three sub-schemas: 1. the current `oneOf` defined above for valid `event_type` and `event_type_reason` combinations
2. this, as a `oneOf`, defining when `associated trips` is required
3. the rest of the properties, defined below.

More specifically, this would look like (non-relevant intermediate lines omitted for brevity):
```json
"items": {
  "$id": ...,
  ...
  "required": [...],
  "allOf": [
    { "oneOf": [ "schema_for_valid_event_types_and_reasons" ] },
    { "oneOf": [ "this_schema_for_associated_trip" ] },
    { "properties": "...properties listed below...." }
  ]
}
```

(caveat this is the first time I've read the JSON schema spec closely, so take this with a dose of skepticism/it may be way wrong)

provider/status_changes.json
```diff
@@ -300,6 +300,28 @@
                 }
```
oderby: Also just realized - this file is auto-generated. I believe the one you want to change is https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/generate_schema/provider/status_changes.json

provider/status_changes.json
hunterowens: tagging @ian-r-rose,

hunterowens:
Closing as this is supplanted in #217.

PR #203, opened 2019-01-01
Refactor `agency/service_areas` to express geofences.

hunterowens:
This PR refactors agency implementation of the service areas to expresses geofences fully. 

@black-tea wrote up a [document](https://docs.google.com/document/d/170W8rWGMmBDb7U-4OvNAQgk_Jusk6UKrsE-BP-Gk5VQ/edit) a expressing what cities have done with a geofence so far, I recommend reading it as the design spec for this PR. 

The implementation is fairly simple. `service_areas` becomes an ordered list of `services_area` that are implement in a last to first order, with `allowed_events` and `forbidden_events` based on the event type from `provider/status_changes`. 

Additionally, this PR removes that provider maintains a list of service areas, per comments in #110 since agencies, rather than providers, are the source of truth for service_areas.


hunterowens:
@thekaveman Whoops, fixed that.

agency/README.md
```diff
@@ -166,7 +166,11 @@ The route must include at least 2 points, a start point and end point. Additiona
 
 ## service_areas
 
-Gets the list of service areas available to the provider.
+Gets the list of service areas available to the provider. A service area is a rules engine that provides a set of service areas that implements [geofences and incentive areas](https://docs.google.com/document/d/170W8rWGMmBDb7U-4OvNAQgk_Jusk6UKrsE-BP-Gk5VQ/edit) along with a base layer set of rules for operation of devices inside a jurisdiction running the Agency endpoint. 
+
+The endpoint returns an a ordered list of GeoJSON MultiPolygons, per provider, shows which type of events are allowed in what areas. The list of the events in the same as defined in [provider/status_changes](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/provider#status-changes).
```
thekaveman: Probably want to link to `master` here.

agency/README.md
```diff
@@ -176,28 +180,22 @@ Query Parameters:
...
+| `service_end_date` | Unix Timestamp | Optional | Datetime at which the service area is no longer in effect. For example, a temporary restriction for an event can be placed in advance and set to end at a certain time. |
+| `service_area` | MultiPolygon | Required | Per [GeoJSON spec](https://tools.ietf.org/html/rfc7946), must follow the right-hand rule to designate interior.   |
+| `allowed_events` | List | Required | List of permitted events inside the area | 
+| `forbidden_events` | List | Required | List of events that is not permitted inside the area | 
+| `speed_cap` | Integer | Optional | Speed cap to be enable on device, in integers, for devices that support governors. Expressed in miles per hour |
+| `preferred_pick_up` | Boolean | Optional | If true, means this service area is an incentive zone for user pick up, and should be expressed to users |
+| `preferred_drop_off` | Boolean | Optional |  If true, means this service area is an incentive zone for user drop offs, and should be expressed to users in app |
```
thekaveman: Does it make sense to refactor those last three columns `speed_cap`, `preferred_pick_up`, `preferred_drop_off` into an enum or sub-object? It feels like there might be more of these "attributes" in the future - for example a field that mirrors the `allowed_events` as a list of enum types. 

I haven't thought about this too hard yet. Just throwing it out there.

provider/README.md
```diff
@@ -253,42 +253,6 @@ The status_changes API should allow querying status changes with a combination o
 | | | `rebalance_pick_up` | Device removed from street and will be placed at another location to rebalance service |
 | | | `maintenance_pick_up` | Device removed from street so it can be worked on |
 
-## Service Areas
```
thekaveman: 👍

agency/README.md
hunterowens: hmmm yeah. I think speed_cap is worth separating because `type(int)`, but how does something like: 

`zones: [list_of_zone_types]`

with `zone_types` == preferred_pick_up, preferred_drop_off, valet_station, any others?

thekaveman: 👍 I like `zone_type`. Also agreed that `speed_cap` is different enough.

One clarification on `speed_cap` - this is intended to be the new maximum speed the device is supposed to be able to go, within this zone - right? We should probably be crystal clear in the definition column there.

agency/README.md
```diff
@@ -176,28 +180,22 @@ Query Parameters:
...
-
-### Area types
-
-| `type` | Description |
-|------------| ----------- |
-| unrestricted | Areas where devices may be picked up/dropped off. A provider's unrestricted area shall be contained completely inside the agency's unrestricted area for the provider in question, but it need not cover the entire agency unrestricted area. See the provider version of the service areas endpoint |
-| restricted | Areas where device pick-up/drop-off is not allowed |
-| preferred_pick_up | Areas where users are encouraged to pick up devices |
-| preferred_drop_off | Areas where users are encouraged to drop off devices |
+| `service_start_date` | Unix Timestamp | Required | Datetime at which this service area enters into effect. For the base service_area, the time should be the start of permitted operations in the City. |
```
karcass: I thought the spec was standardizing on milliseconds everywhere, not seconds.  I realize millis is overkill for this, but the spec should use the same notion of `timestamp` everywhere.

agency/README.md
```diff
@@ -176,28 +180,22 @@ Query Parameters:
...
+| `service_end_date` | Unix Timestamp | Optional | Datetime at which the service area is no longer in effect. For example, a temporary restriction for an event can be placed in advance and set to end at a certain time. |
+| `service_area` | MultiPolygon | Required | Per [GeoJSON spec](https://tools.ietf.org/html/rfc7946), must follow the right-hand rule to designate interior.   |
+| `allowed_events` | List | Required | List of permitted events inside the area | 
+| `forbidden_events` | List | Required | List of events that is not permitted inside the area | 
+| `speed_cap` | Integer | Optional | Speed cap to be enable on device, in integers, for devices that support governors. Expressed in miles per hour |
```
karcass: Prefer meters/sec, with KPH as a second choice.

agency/README.md
```diff
@@ -166,7 +166,11 @@ The route must include at least 2 points, a start point and end point. Additiona
 
 ## service_areas
 
-Gets the list of service areas available to the provider.
+Gets the list of service areas available to the provider. A service area is a rules engine that provides a set of service areas that implements [geofences and incentive areas](https://docs.google.com/document/d/170W8rWGMmBDb7U-4OvNAQgk_Jusk6UKrsE-BP-Gk5VQ/edit) along with a base layer set of rules for operation of devices inside a jurisdiction running the Agency endpoint. 
+
+The endpoint returns an a ordered list of GeoJSON MultiPolygons, per provider, shows which type of events are allowed in what areas. The list of the events in the same as defined in [provider/status_changes](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/master/provider#status-changes). 
+
+Rules are given from least-specific to most-specific. For any given point, the most recent point-in-polygon operation should be considered valid. For example, if you have two overlapping Polygons, the polygons who appeared closest to the end of the list represents in the in affect rules, subject to time rounds.
```
karcass: Similar wording confusion.  Perhaps:

`Where geofences overlap in space and time, rules are enforced according to the linear order of the geofences as returned by the API.  When two or more in-effect geofences overlap, the fence closest to the end of the list represents in the applicable set of rules.`

agency/README.md
```diff
@@ -166,7 +166,11 @@ The route must include at least 2 points, a start point and end point. Additiona
 
 ## service_areas
 
-Gets the list of service areas available to the provider.
+Gets the list of service areas available to the provider. A service area is a rules engine that provides a set of service areas that implements [geofences and incentive areas](https://docs.google.com/document/d/170W8rWGMmBDb7U-4OvNAQgk_Jusk6UKrsE-BP-Gk5VQ/edit) along with a base layer set of rules for operation of devices inside a jurisdiction running the Agency endpoint. 
+
+The endpoint returns an a ordered list of GeoJSON MultiPolygons, per provider, shows which type of events are allowed in what areas. The list of the events in the same as defined in [provider/status_changes](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/master/provider#status-changes).
```
karcass: The wording is a little garbled here.  How about
`The endpoint returns an ordered list of GeoJSON MultiPolygons, per provider, detailing which type of vehicle events are allowed in those areas.`

agency/README.md
hunterowens: good call. @karcass will change.

oderby:
How does this relate to https://github.com/CityOfLosAngeles/mobility-data-specification/pull/197? The changes to the Agency API seem to directly conflict.

hunterowens:
moving this to 0.4.0, since we are still undecided on how best to implement this.

dyakovlev:
@hunterowens should the service areas endpoint revert portion of this PR be retained for 0.3.0 still?

agency/README.md
```diff
@@ -176,28 +180,22 @@ Query Parameters:
...
-
-| `type` | Description |
-|------------| ----------- |
-| unrestricted | Areas where devices may be picked up/dropped off. A provider's unrestricted area shall be contained completely inside the agency's unrestricted area for the provider in question, but it need not cover the entire agency unrestricted area. See the provider version of the service areas endpoint |
-| restricted | Areas where device pick-up/drop-off is not allowed |
-| preferred_pick_up | Areas where users are encouraged to pick up devices |
-| preferred_drop_off | Areas where users are encouraged to drop off devices |
+| `service_start_date` | Unix Timestamp | Required | Datetime at which this service area enters into effect. For the base service_area, the time should be the start of permitted operations in the City. |
+| `service_end_date` | Unix Timestamp | Optional | Datetime at which the service area is no longer in effect. For example, a temporary restriction for an event can be placed in advance and set to end at a certain time. |
+| `service_area` | MultiPolygon | Required | Per [GeoJSON spec](https://tools.ietf.org/html/rfc7946), must follow the right-hand rule to designate interior.   |
```
lionel-panhaleux: "service_area" is already the endpoint name, maybe consider renaming this field ? (eg. "polygons")

hunterowens:
@karcass you have time to revisit this with teh revised "policies" model you referenced?

karcass:
Yes, sorry for the delay.  We have been prototyping a policy-interpreter to make sure the policy schema is sensible and easily converted into compliant/not-compliant outputs.  Hoping to get you a draft Monday before releasing to the wider community.

karcass:
Recommend we close in favor of Policy proposal

hunterowens:
agreed. closing.

Issue #157, opened 2018-10-30
Agency:  Seperate TNC / Fleet Owner


toddapetersen:
The Agency does not contemplate having a separate fleet manager (currently provider_id) and TNC.   We may need to add a provider_ID and potentially TNC_id  when starting / ending trips or issue provider_id to both providers and TNCs.

HenriJ:
@toddapetersen Is this issue still relevant ?

toddapetersen:
Seems like we have moved past this.   We can close it out.   Thanks for asking.

Issue #95, opened 2018-09-27
Consider incorporating TNCs/Ride-hail services


sdrewc:
Given that TNCs are typically considered to be part of a "mobility-as-a-service" package, and currently the most impactful, they seem like a natural extension to this data standard.  The key changes would be to add 'automobile' as a vehicle_type, and to create an out-of-service API analogous to the trip API.  Or even more simply, a field for 'in-service' or 'out-of-service'.

migurski:
This is a great idea, 👍 to `vehicle_type=automobile`. One wrinkle would be the device ID: it doesn’t identify any person for a scooter, but TNC vehicles are privately owned cars with drivers who have an expectation of privacy. The `device_id` should be optional in this case.

hunterowens:
I'm going to try and assemble a draft PR for this today. 

One question I am struggling with - how to represent shared trips, ie, when the trip has more than one passenger.

migurski:
I spoke with Stephanie Dock from DDOT about this on Wednesday and she helped me see that shared trips represent a big change to the MDS trip data model. Maybe MDS should set TNCs aside for the moment to ensure that it’s successful for single-occupant micromobility devices?

hunterowens:
Yeah. There is some urgency from leadership in LA to support TNC for certain use cases. My thought would be to have each mode have its own directory in provider, as such, in `provider`. 
```
├── auth.md
├── micromobility
│   ├── README.md
│   ├── status_changes.json
│   └── trips.json
└── tnc
    └── README.md
```

We could then use the common datatypes from JSONSchema in `generate_schema` to share datatypes, enums, etc.

chursaner:
@hunterowens Drafted for_hire_vehicle trips on my fork here https://github.com/chursaner/mobility-data-specification/commit/7eeb9373c3c4ba09ebd1683ffd2d98c74605765f . I can add it as a pull request once you reorganize the directory!

chursaner:
Would the Accuracy field still be applicable?

sdrewc:
@migurski "TNC vehicles are privately owned cars with drivers who have an expectation of privacy. The device_id should be optional in this case."
The vehicles are privately owned, but they're providing a commercial service.  Taxi medallions and vehicles may also be privately owned, but don't have a similar presumption, and in fact often report both the vehicle and driver.  Definitely IDs should be anonymized, but there is analytical value in being able to link trips since research shows there is almost as much deadhead VMT as in-service VMT.

[...]

PR #204, opened 2019-01-02
Schema cleanup

hunterowens:
This PR renames `generate_schema` to `schema`, and removes the `Pipfile`, because we have a `requirements.txt`


thekaveman:
I like the cleanup idea; part of the reason for the verbose directory naming was so that this directory was not confused as containing *the actual schemas*, which are in the `provider` directory. 

However with work on #203 and discussions in #95 and #157, I think it makes sense to have a dedicated `schema` directory, perhaps with code and schema files alike?

Maybe create an additional subdirectory `schema/generate`, with existing code renamed to `provider.py`, and the existing `.json` moved in (as they are really schema partials/templates). Then we can generate the actual schemas into the root of the `schema` directory or sub-directories as outlined in #95.

Alternatively, leave the python alone and create a `schema/templates` or similar subdirectory for the existing `.json`? So they are not confused as the actual schemas.

hunterowens:
closing in favor of #328

PR #205, opened 2019-01-07
Add GBFS Feed Url to providers.csv

asadowns:
Add GBFS API url to providers.csv

From the README
> All MDS compatible `provider` APIs must expose a public [GBFS](https://github.com/NABSA/gbfs) feed as well.

Currently there is no way to discover MDS provider URLs to ensure compatibility and make this information
        available to the public.

As part of being MDS compliant providers MUST provide this information and make it available to the public.

For today, the best way to do this is to add it to the `providers.csv` and have providers put the URL there
        as part of their fulfillment of the MDS specification

This feels like it should be added to the 0.2.1 release tag rather than waiting for the 0.3.0 release as
        it is a current requirement


karcass:
I think at bare minimum you'll want to add a comma to the rest of the providers who have not yet provided their GBFS API URL.

asadowns:
@hunterowens you raise a good point but I actually think GBFS has indicated what the solution to this is. Providers should provide a single _discovery_ endpoint from which all of their feeds can be accessed. 

`gbfs.json` should not be city specific but should provide all `feeds` that are available from a given provider. In the example you gave above something like `https://jumpbikes.com/opendata/gbfs.json` should provide 
```
{
  "last_updated": 1434054678,
  "ttl": 0,
  "data": {
    "en": {
      "feeds": [
        {
          "name": "New York City freebikes",
          "url": "https://nyc.jumpbikes.com/opendata/gbfs.json"
        },
        {
          "name": "Seattle freebikes",
          "url": "https://sea.jumpbikes.com/opendata/gbfs.json"
        }
      ]
    }
  }
}
```

See: [GBFS](https://github.com/NABSA/gbfs/blob/master/gbfs.md#gbfsjson)

hunterowens:
@asadowns yeah, let's amend the realtime data section in provider to add that you should have the single entrypoint inside `providers.csv`. Once we do that we can merge this in for 0.3.0.

asadowns:
@hunterowens I've amended the Realtime data section. Let me know if there's an additional tweaks you'd like.

hunterowens:
Thanks @asadowns. Merging!

PR #217, opened 2019-02-01
Updated associated trip patch

oderby:
I had some time, so I fixed up and rebased #202. Same fix, but with json schema fixes (it wasn't actually requiring `associated_trip`) and cleanup.

### DIFF
--- a/provider/status_changes.json
+++ b/provider/status_changes.json
@@ -52,6 +52,7 @@
           ]
         },
         "properties": {
+          "type": "object",
           "required": [
             "timestamp"
           ],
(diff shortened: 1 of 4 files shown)


provider/status_changes.json
```diff
@@ -363,19 +292,103 @@
...
+                "description": "valid event_type and event_type_reason combinations",
+                "oneOf": [
+                  {
+                    "properties": {
+                      "event_type": {
+                        "enum": [
+                          "available"
+                        ]
+                      },
+                      "event_type_reason": {
```
hunterowens: @oderby I may be reading this wrong, but shouldn't the only `event_type_reasons` be `user_pick_up` and `user_drop_off`

oderby: Yeah - that's specified at [line 375](https://github.com/CityOfLosAngeles/mobility-data-specification/pull/217/files#diff-f4b85af6fa965f8d99ba6dceda9e1d80R375) for the sub-schema restricting when `associated_trip` is required. Here, this is the sub-schema for valid combinations of `event_type` and `event_type_reason`

PR #218, opened 2019-02-01
Update README to include Austin TX MDS-related projects

mateoclarke:


mateoclarke:
@johnclary, just a heads up on the PR in case you have anything you'd like to add/change.

PR #219, opened 2019-02-04
Add "car" vehicle type

dbaty:
I am working with BlueLA, a car sharing service in Los Angeles. We have implemented the MDS API. 

The MDS API currently has 2 types of vehicle: `bicycle` and `scooter`. I have seen discussions about adding new types (at least #95 #167 #170) but they seem to be stale and/or related to broader subjects (and thus less actionable).

I think that "just" adding a new `car` vehicle type could be an easy and good enough first step.

PR #220, opened 2019-02-04
Add provider: BlueLA

dbaty:
BlueLA is an electric car sharing service in Los Angeles. We have implemented the MDS API and believe we are ready for inclusion in the list of providers.

I believe that we are the first service with *cars*. Apart from the vehicle type that we obviously had to customize (we use "car" for now [1]), we've been able to successfully implement the MDS (as well as the parts of GBFS that are mandated by the MDS).

[1] Speaking of the vehicle type, I have seen discussions about that, albeit much broader (and, as such, probably less actionable). I took the liberty of proposing #219 that adds a new `car` vehicle type.


karcass:
@hunterowens any reason to hold this PR?

hunterowens:
@dbaty looks like the GBFS feed requires an HTTP Authorization Header? Otherwise looks good.

PR #223, opened 2019-02-05
Update provider to allow precision based truncate. fixes #209

### LINKED ISSUE #209: Provider Route Precision
The Provider `README.md` states

> routes must include all possible GPS samples collected by a provider.

and the example route points include coordinates such as `33.9909333514159` with 13 degrees of precision. 

In practice, rounding to 5-6 decimal places is already at the limit of GPS accuracy and useful precision with the remainder of the decimals being noise. 

Is it acceptable for providers to round number of decimals to 5 or 6 places. This would reduce payload and storage size for route data substantially.

hunterowens:
This allows providers to truncate the GPS to the maximum level of precision given the accuracy of whichever measurement system the use.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -169,7 +169,7 @@ A device may have one or more values from the `propulsion_type`, depending on th
 
 To represent a route, MDS `provider` APIs must create a GeoJSON [`FeatureCollection`](https://tools.ietf.org/html/rfc7946#section-3.3), which includes every [observed point][geo] in the route.
 
-Routes must include at least 2 points: the start point and end point. Additionally, routes must include all possible GPS samples collected by a provider.
+Routes must include at least 2 points: the start point and end point. Additionally, routes must include all possible GPS samples collected by a provider. However, providers can truncate the the latitude and longitude to the level of precision representing the maximum accuracy of the specific measurement. (For example, a-GPS is accurate to 5 decimal places, differential GPS is generally accurate to 6 decimal places. Providers should truncate those readings to the appropriate number for their systems.) 
 
 ```js
 "route": {
--- a/provider/README.md
+++ b/provider/README.md
@@ -169,7 +169,7 @@ A device may have one or more values from the `propulsion_type`, depending on th
 
 To represent a route, MDS `provider` APIs must create a GeoJSON [`FeatureCollection`](https://tools.ietf.org/html/rfc7946#section-3.3), which includes every [observed point][geo] in the route.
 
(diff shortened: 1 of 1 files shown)


provider/README.md
```diff
@@ -169,7 +169,7 @@ A device may have one or more values from the `propulsion_type`, depending on th
 
 To represent a route, MDS `provider` APIs must create a GeoJSON [`FeatureCollection`](https://tools.ietf.org/html/rfc7946#section-3.3), which includes every [observed point][geo] in the route.
 
-Routes must include at least 2 points: the start point and end point. Additionally, routes must include all possible GPS samples collected by a provider.
+Routes must include at least 2 points: the start point and end point. Additionally, routes must include all possible GPS samples collected by a provider. However, providers can truncate the the latitude and longitude to the level of precision representing the maximum accuracy of the specific measurement. (For example, a-GPS is accurate to 5 decimal places, differential GPS is generally accurate to 6 decimal places. Providers should truncate those readings to the appropriate number for their systems.)
```
oderby: @hunterowens curious on use of `truncate` vs `round` here?

asadowns: This makes a lot of sense. And I guess I would favor rounding over truncating but the difference should obviously be small in either case.

thekaveman: > However, providers can truncate **the** **the** latitude ...

hunterowens: difference is small but yeah, `round` is the right word. 

@thekaveman fixed typo.

provider/README.md
```diff
@@ -169,7 +169,7 @@ A device may have one or more values from the `propulsion_type`, depending on th
 
 To represent a route, MDS `provider` APIs must create a GeoJSON [`FeatureCollection`](https://tools.ietf.org/html/rfc7946#section-3.3), which includes every [observed point][geo] in the route.
 
-Routes must include at least 2 points: the start point and end point. Additionally, routes must include all possible GPS samples collected by a provider.
+Routes must include at least 2 points: the start point and end point. Additionally, routes must include all possible GPS samples collected by a provider. However, providers can round the latitude and longitude to the level of precision representing the maximum accuracy of the specific measurement. (For example, a-GPS is accurate to 5 decimal places, differential GPS is generally accurate to 6 decimal places. Providers should truncate those readings to the appropriate number for their systems.)
```
thekaveman: > Providers should **truncate** those readings to the appropriate number for their systems.

Was the decision to stick to `round` for the verbiage?

thekaveman: Fixed this.

PR #225, opened 2019-02-05
 Require each event to include a UUID

### LINKED ISSUE #191: No identifier specified for `status_changes` entities
Status change entities don't currently have an identifier, so it's not clear how consumers should de-duplicate these entities.

oderby:
Per discussion in #191, while in theory every event should be
unique for every `(provider_id, device_id, event_time)` tuple,
in practice this is not the case, for a variety of reasons. To
help facilitate data ingestion, referential integrity, and general
debugging/discussion, we're adding a required UUID column to every
status_change event.

Fixes #191

### DIFF
--- a/generate_schema/provider/status_changes.json
+++ b/generate_schema/provider/status_changes.json
@@ -36,6 +36,7 @@
               "vehicle_id",
               "vehicle_type",
               "propulsion_type",
+              "event_id",
               "event_type",
               "event_type_reason",
               "event_time",
@@ -76,6 +77,11 @@
                 "$id": "#/properties/data/properties/status_changes/items/properties/propulsion_type",
                 "description": "The type of propulsion; allows multiple values",
                 "$ref": "#/definitions/propulsion_type"
+              },
+              "event_id": {
+                "$id": "#/properties/data/properties/trips/items/properties/event_id",
+                "description": "A unique ID for each event",
+                "$ref": "#/definitions/uuid"
               },
               "event_time": {
                 "$id": "#/properties/data/properties/status_changes/items/properties/event_time",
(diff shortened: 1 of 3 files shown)


karcass:
I'm not enthused about this.  What do you think @hunterowens?  If we add this in provider, we should probably add it to agency.

oderby:
I feel pretty strongly that MDS needs to have stable and persistent unique identifiers for status change events. It's important if we want to move towards more of a streaming API, for example. 

However, I think it's unreasonable to include this in 0.3.0, given that it would be a fairly major change and some providers don't seem to have systems in place yet to support this. So, I'd vote that we table this for now, and consider it for a future release?

hunterowens:
@oderby seems reasonable to me, just flagged this for 0.4.0. I'm going to do the same with #191.

lionel-panhaleux:
The way it is worded right now does not ensure any kind of stability _per se_.

As @oderby idea is to use this to move to a streaming API, I fell we need to add:
- the event UUID must be stable over time
- the same UUID must be used on the agency side of things.
- and maybe, it should change if the event data changes ? Or require the event data to stay unaltered over time (which is a strong requirement)

For now, if the provider has no UUID on the event to begin with (and he may have a unique ID, but an UUID ? There is no guarantee on such a large internal dataset), he could decide either:
- To generate unique UUID-4 or UUID-5 on the fly on the Provider API (they will not be stable)
- To use the UUID-1 if the (device, timestamp) is unique
- To migrate the whole dataset and add a stable UUID-4 on each event.
Only the third solution would be stable.

Note the current `uuid` definition (in `common.json`) asks for an UUID-4 but the given example is an UUID-1. Asking for a unique ID formatted as a string is sensible, asking for a UUID is a strong format requirement, asking for a specific version of the UUID standard seems a little bit much.

karcass:
I believe the consensus at the Provider call was that we were going to consider a sequence-id rather than a UUID, and that @lionel-panhaleux was going to write that up.

lionel-panhaleux:
Totally, I wrote this comment before the meeting. Sorry for the delay about the sequence-ID thins, I was in holidays for a week. I'll get to it today.

karcass:
Recommend we close this.

hunterowens:
going to close, publication_time should help on this. @oderby let me know if you feel strongly otherwise.

Issue #214, opened 2019-01-24
Status change events for entering and leaving visible area


rf-:
Let’s call the area where a given Provider API consumer is allowed to see all status changes the “visible area” for purposes of this issue. This is likely to be the boundaries of the jurisdiction of the agency making the request, but it currently varies between providers.

Right now, it’s unclear what events should appear in the status changes API when a user reserves a vehicle and takes it out of the visible area or vice versa.

It seems like there are currently two possibilities:

- The status changes API includes events reflecting both endpoints of any trip that touches the visible area. If the user takes a vehicle into a neighboring city, the end of that trip will create an `available` status change, but whatever happens to that vehicle afterwards won’t be reflected in the API.

- The API only includes events that happen inside the visible area, so the end of a trip that leaves the visible area won’t appear as a status change.

In the first case, clients that want to put together a complete picture would need to observe that the trip ended outside the visible area and use that as a cue to not expect future events related to that vehicle (as opposed to the naive alternative of treating the vehicle as permanently `available` at whatever the end point of the trip was). This would work, but it depends on the provider and the client having a precise shared definition of the visible area, which doesn’t currently exist in MDS. Another downside of this approach is that, in the case where the user starts outside the visible area and then enters it, the provider will have to retroactively add a `reserved` event for the beginning of the trip once they see that the user has crossed into the visible area, making the contents of the feed inconsistent over time.

In the second case, the last event the client sees for a vehicle leaving the visible area is `reserved`. This prevents the client from using the status changes API to calculate whether the provider is compliant with vehicle limits, since they don’t know if or when the vehicle actually left the visible area, or when a vehicle entered the visible area before being dropped off. This also makes it difficult to validate the integrity of the event stream, since basically any event can follow a `reserved` event.

One solution to this problem would be to add explicit status change events when a vehicle leaves or enters the visible area, defined as the area within which the current API client is allowed to see events. The least invasive implementation would be to add a `left_visible_area` event type reason for the `removed` event type and an `entered_visible_area` reason for `reserved`. If that feels like an inappropriate use of `removed`, it could make sense to add a new event type instead.

rf-:
Interesting, so maybe the path forward is to port these events to the Provider side?

karcass:
I think so :)

hyperknot:
[...]

PR #226, opened 2019-02-06
[Provider] Clarify how endpoints relate to municipality boundaries.

oderby:
In conversations and from ingesting data, it's become clear there is
ambiguity in how parties are reporting data in their MDS Provider API
implementations with respect to the boundary of the municipality they
are operating in. This commit adds some clarifications and definitions
to capture the original intent of the spec and what most cities (Santa
Monica and LA) have been expecting in this regard.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -127,7 +127,19 @@ represented as a GeoJSON [`Feature`](https://tools.ietf.org/html/rfc7946#section
 }
 ```
 
+#### Intersection Operation
+For the purposes of this specification, the intersection of two geographic datatypes is defined according to the [`ST_Intersects` PostGIS operation](https://postgis.net/docs/ST_Intersects.html)
+
+    If a geometry or geography shares any portion of space then they intersect. For geography -- tolerance is 0.00001 meters (so any points that are close are considered to intersect)
+
+    Overlaps, Touches, Within all imply spatial intersection. If any of the aforementioned returns true, then the geometries also spatially intersect. Disjoint implies false for spatial intersection.
+
 [Top][toc]
+
+### Municipality Boundary
+Municipalities requiring MDS Provider API compliance should provide an unambiguous digital source for the municipality boundary. This boundary must be used when determining which data each `provider` API endpoint will include.
+
+The boundary file should be provided in Shapefile or GeoJSON format. The boundary file should be hosted online at a published address that all providers and `provider` API consumers can access and download.
 
 ### Timestamps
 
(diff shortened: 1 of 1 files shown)


ezheidtmann:
Thanks for highlighting this problem and filing this issue! What other approaches have been considered? What approaches are providers taking? Why are these changes the best option at this point?

billdirks:
Has posting a status change event to each municipality when a vehicle crosses a municipal boundary (ie exceeds the tolerance out of 1 municipality and into another) been considered? For example, a removed event could be posted to the first municipality and a reserved to the second when a rented vehicle crosses the border. Multiple crossing in one reservation would all have the same trip id. This would allow one to get a view of the vehicles in a municipality without looking at the trips endpoint. In addition, it would make it clear to agencies when a provider considers a vehicle in a new municipality, which can then be evaluated.

oderby:
@ezheidtmann @billdirks see #214 for a lengthy conversation on the various ideas considered here. This PR is an attempt at capturing the current status quo, so we can all agree on it before 0.3.0 is cut. More involved solutions may still be reasonable, and should be considered in future issues/PRs.

billdirks:
@oderby Thanks for pointing me to #214, I'll read it over.

provider/README.md
```diff
@@ -127,8 +127,19 @@ represented as a GeoJSON [`Feature`](https://tools.ietf.org/html/rfc7946#section
...
+
+> If a geometry or geography shares any portion of space then they intersect. For geography -- tolerance is 0.00001 meters (so any points that are close are considered to intersect).<br>
+> Overlaps, Touches, Within all imply spatial intersection. If any of the aforementioned returns true, then the geometries also spatially intersect. Disjoint implies false for spatial intersection.
+
 [Top][toc]
 
+### Municipality Boundary
+Municipalities requiring MDS Provider API compliance should provide an unambiguous digital source for the municipality boundary. This boundary must be used when determining which data each `provider` API endpoint will include.
+
+The boundary should be defined as a polygon or collection of polygons. The file defining the boundary should be provided in Shapefile or GeoJSON format and hosted online at a published address that all providers and `provider` API consumers can access and download.
```
asadowns: Could we specify GeoJSON explicitly? Shapefile is generally not the best format to work with in my experience. Obviously, we could do either but I like specifying `geojson` as that seems to be what MDS and `mds-provider-services` use.

thekaveman: I agree that Shapefile isn't as generally useful *directly*, but there are many tools to convert Shapefile -> geoJSON, and Shapefile is more of a de-facto standard format for many municipalities.

asadowns: makes sense!

thekaveman:
@oderby can you please rebase on `dev` so this reflects the related clarifications from #223

oderby:
@thekaveman rebase pushed. There was no conflicts, and I didn't see how any of the language being added here would need to be updated as a result of #223. Please let me know if there's a specific section that needs to be updated/fixed?

thekaveman:
@oderby you are right, I was looking at the (non-green/red) text *below* the changes you made, and saw it wasn't up to date. Thanks for rebasing anyway.

PR #231, opened 2019-02-12
Add Lime & Spin GBFS

mikey186:
Here's Lime's link ```https://data.lime.bike/api/partners/v1/gbfs/los_angeles/gbfs.json``` and Spin ```https://web.spin.pm/api/gbfs/v1/los_angeles/gbfs```


hunterowens:
@mikey186 thanks for the heads up! As you'll notice, these are the LA specific `gbfs.jsons`. We'll be working with Lime and Spin to get there autodiscovery urls in place.

Issue #170, opened 2018-11-07
Determine how to support multiple modes


hunterowens:
ref #95 #167 #157. 

We need to figure out structurally how to support different modes aside from bikes/scooters.

HenriJ:
@hunterowens do you already have a list of the different modes that should be supported ?

* Bikes/scooters = station-based (we might require an API to register stations like we have an API for areas ?) or dockless with no information on trip destination
* Car-sharing = station-based or free-floating. Trip destination can sometimes be known but not always
* TNC (& taxis ?) = confidentiality issues as reported by #95 . We should always know the trip destination.

Attributes of the vehicles are different for each mode but the macro-structure of the API should stay the same ?

d-wasserman:
I think one overarching aspect that might change is an aspect related to occupancy if possible for other modes. Thinking of services like Via or TNC's pooled services, I could see that being a useful attribute. Currently, the macro-api (correct me if I am wrong) assumes a single trip corresponds to a single person. Down the line, I could see something like this interacting with a tolling APIs, etc.

hunterowens:
moving to `0.4.0`, as I don't have a great answer for this yet but would like to get the 0.3.0 changes out the door.

hunterowens:
closed in #366!

PR #237, opened 2019-02-13
Add new `battery_charged` event type reason

### LINKED ISSUE #210: New "battery_ok" event type reason to mirror "low_battery"
Context: I am working on an implementation of MDS for BlueLA, a station-based electric car sharing service in Los Angeles. The suggestion below would apply there and also for any service where devices can autonomously recharge their battery on docks (charge points).

When a customer returns a device to a dock, the battery of the device may be too low for the device to be available for another customer. The service then yields an `unavailable` event type with the `low_battery` reason. That part is fine. Then the device charges itself at the dock (without needing any human interaction). When it's done, the device becomes available again. I think that the `maintenance_drop_off` event type reason is not appropriate here, because there is no (human) maintenance per se. I hence suggest a new `battery_ok` event type reason for the `available` event type.

If this sounds reasonable, I'll happily write a pull request.

This is related to #77, albeit different.

dbaty:
This pull request addresses #210.

---

When the battery of a device is too low, the service yields the
`unavailable` event type with the `low_battery` reason. Then the
device may charge itself (for example because it has been plugged to a
charging dock by the last customer, or because the device has a solar
panel). Once the battery is charged enough, the service can yield an
`available` event type with the `battery_charged` reason.

[closes #210]


hunterowens:
@karcass this PR needs to wait on determine how to support multiple modes, see #170. 

Can you prepare a PR to revert this in Dev?  This also needs some tweaks on the JSON schema

hunterowens:
@dbaty, just to clarify why I asked for a reversion of this PR. 

I think we need to have a clear way we add an support additional modes to MDS, as there are a number of difference in the core technologies, data we need and legal frameworks to support for different modes. 

In #95, I propose a directory structure, with the `generate_schema` directory handling the heavy lifting of unifying subdatatypes across different modes (ie, `timestamps`). 

What PR that we have to close #170 should have a how to guide about writing MDS support for different modes.

dbaty:
It's not clear to me how this event is linked to different transport modes. Could they not apply to electric bikes and scooters that are currently supported?

PR #238, opened 2019-02-13
Agency schema redux

### LINKED ISSUE #169: JSON Schema for Agency

ian-r-rose:
Fixes #169.

Supersedes #172.

I have tried to follow the current `agency/README.md` in `dev` as closely as possible. If there are existing open PRs which I should also try to take into account, feel free to let me know.

Also, this is not particularly well-tested in the absence of synthetic agency data, so there are likely some issues that will need to be ironed out, here or in follow-ups.


karcass:
Apologies to keep moving the goalposts @ian-r-rose, but there's another round of changes that make this PR obsolete.  I think we might push this to 0.3.2?  We'd really like to cut 0.3.1 soon.

ian-r-rose:
Updated in light of #258, #249, #245.

The API is getting rich enough that it's probably time to investigate using OpenAPI rather than pure JSONSchema to better document the responses that are possible from the Agency API. However, I think that's a big enough undertaking that it may be better for a follow-up.

karcass:
Looks good to me.  Would like eyeballs on from @hunterowens and @toddapetersen.

hunterowens:
@karcass this looks good to me. I can clean up, rebase and merge in?

karcass:
@hunterowens yep!

PR #245, opened 2019-02-22
Relax telemetry constraints

lionel-panhaleux:
Depending on the actual device on board of the vehicle,
some information can be unavailable.


karcass:
The concern is that if those are optional, then nobody will send them ever, even if they have them.  We will need to amend the wording in the compliance document to reflect that if they are _available_, they are required.

lionel-panhaleux:
Agree, I reworded for clarification.

karcass:
Cool, looks good to me.  Let me run it past some other LADOT folks to confirm consensus.

hunterowens:
LGTM as well.

hunterowens:
In `provider`, we use "required if applicable" (eg _`associated_trip`_), but I think "required if available" is clearer meaning here. Bit of a bikeshed, though.

PR #247, opened 2019-02-22
agency_api: no transition check on vehicle events

lionel-panhaleux:
The previous formulation suggested that
the state transitions would be checked by the Agency.
This should not  be the case because:

- Events order is hard to guarantee for the provider
- Agency API should not be blocking for providers
- No specific finite-state machine should be imposed


karcass:
Agree.  We originally considered using Agency for both data-ingest and enforcement. For the reasons you list, these are in conflict.  We will formulate additional APIs for enforcement.

rf-:
I would suggest keeping the FSM definition and just removing the expectation that the API will reject invalid events. Even if the provider sends some events out of order, having a well-defined FSM will be useful for validating the eventual completeness/accuracy of the stream.

karcass:
See competing #250.  Leaves state-machine intact, drops rigorous enforcement of state machine, removes 412 error code and reference to business or policy logic.

PR #249, opened 2019-02-22
drop UUID4 requirement in favor of simply UUID

karcass:
to make consistent with Provider


amalakar:
This is great, thank you!

PR #251, opened 2019-02-23
add PUT /vehicles/{device_id}

karcass:
also clean up some telemetry wording

This was requested by Bird.  They have cases where the vehicle-id gets scratched off by a vandal and a new (different) one is applied by their maintenance staff.


toddapetersen:
I believe we need is to modify vehicle_ID.   Device_ID should be immutable based on the description


vehicle_id | String | Vehicle Identification Number (vehicle_id) visible on vehicle

karcass:
Sandbox is updated with new behavior.  Ready for merge.

Issue #242, opened 2019-02-20
Expected output of agency /vehicles endpoint when no device_id is specified is unclear


johnpena:
The [documentation](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/agency#vehicles) for the agency API's `/vehicles` endpoint specifies that the `device_id` parameter is optional and that "If provided, retrieve the specified vehicle". What is the expected output if `device_id` is not provided? The documentation should specify the expected outcome when `device_id` is not provided given that it's optional. 

I would expect that either a) the endpoint should return a list of all registered vehicles when no `device_id` is provided b) the endpoint should return an error code, and the documentation should be changed to specify that `device_id` is required.

karcass:
Ah yeah, sorry about that.  We never reached consensus as to whether or not `GET /vehicles` should behave as your (a) or (b).  Not sure that it matters as long as we pick one.  I think we should get the opinion of a few folks representing providers to express a preference one way or the other.

karcass:
What do you think @johnpena?

johnpena:
I'd prefer option (a) as it'll give providers transparency into the fleet data the agency has on hand, and to audit/correct issues as needed.

karcass:
Sounds legit.  I figured that's what most folks would prefer.  I'll write up a PR.

johnpena:
Thanks for looking into this!

PR #252, opened 2019-02-25
specify behavior for GET /vehicles with no device_id

karcass:
Per discussion in #242

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -3,7 +3,7 @@
 This specification contains a collection of RESTful APIs used to specify the digital relationship between *mobility as a service* Providers and the Agencies that regulate them.
 
 * Authors: LADOT
-* Date: 12 Feb 2019	
+* Date: 25 Feb 2019	
 * Version: BETA
 
 ## Table of Contents
@@ -30,7 +30,7 @@ As with the Provider API, `timestamp` refers to integer milliseconds since Unix 
 
 ## Vehicles
 
-The `/vehicles` endpoint returns the specified vehicle.  Providers can only retrieve data for vehicles in their registered fleet.
+The `/vehicles` endpoint returns the specified vehicle (if a device_id is provided) or a list of known vehicles.  Providers can only retrieve data for vehicles in their registered fleet.
 
 Endpoint: `/vehicles/{device_id}`
 Method: `GET`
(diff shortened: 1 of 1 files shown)


agency/README.md
```diff
@@ -41,8 +41,17 @@ Path Params:
 | ------------ | ---- | ----------------- | ------------------------------------------- |
 | `device_id` | UUID  | Optional          | If provided, retrieve the specified vehicle |
 
+Query Params:
+
+| Param        | Type | Required/Optional | Description                                 |
+| ------------ | ---- | ----------------- | ------------------------------------------- |
+| `skip` | Integer  | Optional          | If provided, offset into list |
+| `take` | Integer  | Optional          | If provided, number of vehicles to load |
```
oderby: With this change, we'll need to add support for pagination to the spec. However, this proposal feels like an implementation detail - in the provider API, [compliance with JSON API spec for pagination is required](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/0.3.0/provider#pagination), and that's been sufficient to get everyone on the same page. Might we consider taking a similar approach here?

karcass: Bleah, I am not a fan that pagination spec.  It's too prescriptive and heavy-weight.

rf-: A limit/offset pagination API is probably not ideal since changes to the underlying data while the client is traversing the list can easily cause omitted or repeated records. I sympathize with not loving JSON API, but we'd need something more robust than this anyway, and I don't know how much sense it makes for MDS to include more than one pagination system.

karcass: All pagination systems are vulnerable to underlying data changes, not just limit/offset.  JSON API doesn't address that:

> Note: JSON:API is agnostic about the pagination strategy used by a server.

It's pretty simple to pick a sort criteria amenable to consistent results (e.g. creation-date of the record).

karcass: Point taken on multiple pagination strategies.  I just despise JSON API's.

karcass: Vehicle records are big enough that a single query for, say, 20K vehicles is out of the question, so IMHO pagination is a firm requirement.  I guess I'll suck it up and use the established one.

rf-: Using cursor-based pagination makes it possible to avoid these problems, although still not trivial. But it sounds like you're convinced, so I'll skip that argument 😅

karcass: I'm still going to use skip/take.  I'm just going to make up a page size in the reference implementation and construct my link URLs accordingly.  We can put off further bikeshedding until another day.

karcass: (The bikeshedding crack is squarely directed at myself.)

rf-: Sure, yeah, it's fine for the reference implementation to use skip/take as long as the spec allows implementors to use something fancier :+1:

karcass: Smart

karcass: Sandbox updated with new behavior

karcass:
@hunterowens I believe this is ready, let me know what you think.  Or I can bug @whereissean.

PR #254, opened 2019-02-26
Revert "Add new `battery_charged` event type reason"

karcass:
Reverts CityOfLosAngeles/mobility-data-specification#237


Retzoh:
Hi @karcass , sorry for returning to this PR later on but why was the `battery_charged` transition removed? It actually makes a lot of sense in the carsharing context.

karcass:
I don't recall.  I think @hunterowens had some concern about it.  Probably should have written that reasoning down!  I'd be fine with re-adding `battery_charged` as a transition from `unavailable` to `available`.

thekaveman:
I think at least [some of the reasoning](https://github.com/openmobilityfoundation/mobility-data-specification/pull/237#issuecomment-467602300) was related to how it wasn't totally clear how new modes (specifically carshare) are represented in MDS. I think we have a better idea now after #336.

PR #256, opened 2019-02-28
Updating schema version

thekaveman:
There's a step missing in the ReleaseGuidelines, and we forgot to update the schema version regex before releasing `0.3.0`. 

This updates the version and adds the step to ensure it isn't missed again.

### DIFF
--- a/ReleaseGuidelines.md
+++ b/ReleaseGuidelines.md
@@ -70,6 +70,8 @@ The following steps **must** be followed for **every** release of MDS:
 
 1. Ensure the [Milestone][mds-milestones] for this release is at `100%`.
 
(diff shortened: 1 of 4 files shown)


hunterowens:
I think we probably will need to do an `0.3.1` for this, right?

hunterowens:
@thekaveman I think we need to create an `0.3.x` branch, based off the `0.3.0` tag. 

I know there are also a bunch of agency PRs that should be included in `0.3.1` - @karcass ?

karcass:
@hunterowens I'll review the open PRs now and tag anything that I think needs to make it to 0.3.1.  Anything that doesn't I'll leave unlabeled.

PR #258, opened 2019-03-01
Clarify and fix event state model

whereissean:
Fixing a number of event/state flow transitions to incorporate initial feedback received from providers. Adding `event_reason` field to capture more granular event designations and add the `annotation` for missing vehicles.  Adding a visual diagram to better describe the expected transitions.


agency/README.md
```diff
@@ -158,8 +158,9 @@ Body Params:
...
 | ----------- | ---------------------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
-| `event_type` | Enum                         | Required          | [Vehicle Event](#vehicle-events)                                                                                                           |
-| `timestamp`  | Timestamp                    | Required |Date of last event update                                                     |
+| `event_type` | Enum                         | Required          | see [Vehicle Events](#vehicle-events)                                                                                                           |
+| `event_reason` | Enum                        | Required if Available | see [Vehicle Events](#vehicle-events)                                                                                                           |
```
hunterowens: Let's rename this to `event_type_reason` to better match the provider spec.

agency/README.md
```diff
@@ -241,24 +242,24 @@ Query Params:
...
+| `cancel_reservation` |                                                         | Customer cancels reservation                                                                   | `reserved`                                         | `available`         |                                                                         |
+| `trip_start`         |                                                         | Customer starts a trip                                                                         | `available`, `reserved`                            | `trip`              |                                                                         |
+| `trip_enter`         |                                                         | Customer enters a service area managed by agency during an active trip.                        | `removed`, `elsewhere`                             | `trip`              |                                                                         |
```
hunterowens: This should be clarified from `service_area` to `geographic boundary` or similar, ie, we don't want events when you leave all the specific `service_areas`, but rather, just when you leave / end municipal boundary.

hunterowens: I believe we have a solid boundary definition in provider now, take a look for guidance.

agency/README.md
```diff
@@ -241,24 +242,24 @@ Query Params:
...
+| `trip_leave`         |                                                         | Customer enters a service area managed by agency during an active trip.                        | `trip`                                             | `elsewhere`         |                                                                         |
+| `trip_end`           |                                                         | Customer ends trip and reservation                                                             | `trip`                                             | `available`         |                                                                         |
+| `deregister`         | `missing`                                             | A vehicle is deregistered                                                                      | `available`, `unavailable`, `removed`, `elsewhere` | `inactive`          | A vehicle is deactivated from the fleet.                                |
+
+
+![Event State Diagram](images/MDS_agency_event_state.png?raw=true "MDS Event State Diagram")
```
hunterowens: Maybe add some lang here about what this diagram is for.

agency/README.md
whereissean: @hunterowens agree the intent is not to receive events on entry and lead from each service area, but actually the boundary that the agency is responsible for managing.  I do think that `service_areas` are meant to define that boundary and that should be the single source of truth.  I will modify to clarify that crossing into or out of the region that is the union of all `service_areas` is the desired trigger.

rf-: In the case where an agency doesn't want to allow providers to operate in all areas of their jurisdiction, I wouldn't necessarily expect `service_areas` to cover the whole jurisdiction but I'd definitely expect providers to include events that happen in those prohibited areas. It's probably worth either distinguishing service areas from jurisdictional boundaries or clarifying that agencies are expected to include either `unrestricted` or `restricted` areas covering their full jurisdiction in the `service_areas` endpoint.

rf-: The former makes more sense to me—it seems unnecessarily onerous to make agencies carve up their whole city boundary into restricted and unrestricted zones to serve in the API, and also I would imagine providers would rather have a stable concept of a given city's boundaries instead of having to trust that the agency will always give them a set of service areas that add up to a consistent boundary.

whereissean: @hunterowens I resolved the wording to be consistent with Provider. @rf you are bringing up some good points and needed clarifications as to what constitutes `service_areas`.  I would like to continue this discussion in a separate issue that I will open, since the purpose of the PR is to clarify and make some fixes to the event model.

rf-: Sounds good, thanks! If it wasn't clear, the reason I mentioned those issues here was just to express support for using language that is about the municipal area and not `service_areas`; the update you just made looks good to me. Happy to hash out remaining issues with `service_areas` elsewhere.

PR #261, opened 2019-03-07
Compliance Patch - Telemetry Update

toddapetersen:
Approved by LADOT on 4Feb2019.  Updated in-trip telemetry compliance for scooters.


hunterowens:
@toddapetersen this should target the `0.3.1` branch, not the `dev` branch. I don't think it should break anything,  so I'm gonna switch from here.

PR #273, opened 2019-03-26
provider, agency: event timestamps

lionel-panhaleux:
For a given device, different events must have different timestamp.
Otherwise, the order of events and the resulting status can not be inferred by the agency.

See https://github.com/CityOfLosAngeles/mobility-data-specification/issues/272


karcass:
One comment, otherwise looks good to me.

hunterowens:
just wrapping up various things, but I'm going to close that as I think guidance is to use the order implied in the event_type / event_type_reason combo.

PR #277, opened 2019-03-26
Provider: order status_changes chronologically

lionel-panhaleux:
For ease of use, status changes should present a chronological order.
They are most likely seen  as a timeseries. See https://github.com/CityOfLosAngeles/mobility-data-specification/issues/276


thekaveman:
I think this was handled by #301?

rf-:
301 is about the timestamps themselves, not the order of the records in API responses. Wasn’t there another PR merged that specified the server is allowed to return records in any order though?

hunterowens:
closing, because any order per @rf-

PR #283, opened 2019-04-02
Added to vehicle event deregister the event_type_reason decommissioned

dirkdk:
When a vehicle is taken off the street because it is damaged beyond repair, call event `deregister` with `event_type_reason` of value `decommissioned `


karcass:
We can target this for 0.3.1 or 0.3.2 I think.  Deferring to @hunterowens.

PR #290, opened 2019-04-04
Invalid telemetry data 400 res

avatarneil:
In the case that none of the provided telemetry data is valid, respond with a 400 error. In the case that at least one of the provided data points is successfully processed, respond with a 201 specifying how many data points were successful out of the total, and a list of failed points (if there are any).

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -197,7 +197,10 @@ Body Params:
 
 201 Success Response:
 
-_No content returned on success._
+| Field     | Type                           | Field Description                                                                                       |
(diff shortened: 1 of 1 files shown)


agency/README.md
```diff
@@ -197,14 +197,18 @@ Body Params:
 
 201 Success Response:
 
-_No content returned on success._
+| Field     | Type                           | Field Description                                                                                       |
+| --------- | ------------------------------ | ------------------------------------------------------------------------------------------------------- |
```
janedotx: 👍

PR #291, opened 2019-04-04
Add event_type_reason for agency admin access. 

hunterowens:
Per @margodawes, this adds the event type reasons for agency admin access. 

I used the term `agency` rather than `admin` for clarity.


ccolgrove:
👍

hunterowens:
whoops, fixing - that should be a removed event, not unavailable.

PR #296, opened 2019-04-17
Update Providers.csv with corrected SPIN API URL

hunterowens:
Need to update SPIN API URL based on [docs](https://web.spin.pm/datafeeds)

[  ] Yes, breaking
[ * ] No, not breaking
[  ] I'm not sure

**`Provider` or `agency`**
[  ] `provider`
[  ] `agency`
[ * ] both


dirkdk:
as Spin employee, I say the new info is correct! not sure where that old info came from

Issue #289, opened 2019-04-04
Add an associated trip id to status changes when a trip ends


billdirks:
Currently the provider specification only requires an `associated_trip` for status change events if the `event_type_reason` is `user_pick_up` or `user_drop_off`. For trip starts an event with `event_type = reservation` and `event_type_reason = user_pick_up` appears mandatory. For trip ends, however, it seems possible that the vehicle doesn't become `available` (a scooter with low battery when a trip ends for example). In this example a trip would end with the state transition (`available`, `reserved`) to (`unavailable`, `low_battery`) and no `associated_trip` would be present in the status change event. It seems the intent is to have an `associated_trip` id present at end of trips even when there is no allowable `user_drop_off` `event_type_reason`.

rf-:
It seems like a good idea to me to require an `associated_trip` for any status change that marks the beginning or end of a trip, regardless of the specific `event_type` and `event_type_reason` it represents. The spec is currently somewhat ambiguous about whether `associated_trip` is allowed for events that don't have a reason of `user_pick_up` or `user_drop_off`. I'd suggest the following language for 0.4.0:

> Trip UUID (foreign key to Trips API), required if `event_type_reason` is `user_pick_up` or `user_drop_off`, or for any other status change event that marks the end of a trip.

For 0.3.1, we could find a middle ground that explicitly allows providers to include the field for other event types but doesn't require it.

If and when the provider state machine is defined more clearly in a future version of MDS, I think it would make sense to allow edges going from the `reserved` state to `available`, `unavailable`, and potentially other statuses, with the stipulation that all edges going into or out of `reserved` should have an `associated_trip`.

thekaveman:
The language of the spec was updated in #297 to be more explicit that `associated_trip` is allowed with any status_change event type. I think that closes this issue for now.

PR #297, opened 2019-04-17
Explicitly allow `associated_trip` for any event type

rf-:
This implements the change suggested in #289 by clarifying that any
trip-ending event can and should have an `associated_trip`, not just
`user_drop_off` events.

Since the Provider event system currently doesn't define which states
are allowed to transition to which other states, it's already
spec-compliant to follow a `reserved`/`user_pick_up` event with an
`unavailable`/`low_battery` event, and this is the most
semantically-accurate way to describe the scenario where a device shuts
down during a trip because of low battery. In this situation, the
implementer should include an `associated_trip` field in both events
so that API consumers can correctly identify these events as the
endpoints of the relevant trip.

**Is this a breaking change**
* [ ] Yes, breaking
* [X] No, not breaking—clients that aren't currently processing this field for these event types can continue to ignore it
* [ ] I'm not sure

**`Provider` or `agency`**
* [X] `provider`
* [ ] `agency`
* [ ] both

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -276,7 +276,7 @@ Schema: [`status_changes` schema][sc-schema]
 | `event_time` | [timestamp][ts] | Required | Date/time that event occurred, based on device clock |
 | `event_location` | GeoJSON [Point Feature][geo] | Required | |
 | `battery_pct` | Float | Required if Applicable | Percent battery charge of device, expressed between 0 and 1 |
-| `associated_trip` | UUID | Required if Applicable | Trip UUID (foreign key to Trips API) required if `event_type_reason` is `user_pick_up` or `user_drop_off` |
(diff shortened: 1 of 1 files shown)


thekaveman:
@rf- @hunterowens so just to circle back, are we OK with this being a language-only change - and leaving the JSON schema alone? Or do we want to [add additional `event_type_reason`](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/provider/status_changes.json#L370) that would qualify as requiring the `associated_trip`?

rf-:
I think `user_pick_up` and `user_drop_off` are the only ones that can require it at the schema level. For any other event type, whether it's required depends on context from other stuff in the event stream, so it's probably not possible to nail down further except as part of a more comprehensive state machine definition.

PR #301, opened 2019-04-29
Change `event_time` definition to require correct ordering

rf-:
This proposal reflects the consensus of the status change ordering/uniqueness breakout group that it's reasonable to ask providers to construct an accurate and causally-consistent timeline of status
change events. This change will improve API consumers' ability to understand the ordering and timing of events in the status changes API.

See [the notes from the first breakout group call](https://groups.google.com/forum/#!topic/mds-announce/RcKFkpJ7qrA) for a little bit more context.

### Is this a breaking change

* [x] Yes, breaking
* [ ] No, not breaking
* [ ] I'm not sure

### `Provider` or `agency`

* [x] `provider`
* [ ] `agency`
* [ ] both

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -289,10 +289,14 @@ Schema: [`status_changes` schema][sc-schema]
 | `propulsion_type` | Enum[] | Required | Array of [propulsion types](#propulsion-types); allows multiple values |
 | `event_type` | Enum | Required | See [event types](#event-types) table |
 | `event_type_reason` | Enum | Required | Reason for status change, allowable values determined by [`event type`](#event-types) |
-| `event_time` | [timestamp][ts] | Required | Date/time that event occurred, based on device clock |
+| `event_time` | [timestamp][ts] | Required | Date/time that event occurred at. See [Event Times](#event-times) |
 | `event_location` | GeoJSON [Point Feature][geo] | Required | |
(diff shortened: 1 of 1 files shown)


rf-:
An open question here is whether we should also update the `start_time` and `end_time` fields on trips to use the same language. I left it out for now since I didn't want to go past the scope of the discussion on the call, but it would probably make sense to be consistent.

karcass:
This will definitely affect Agency, as the data collected via Agency is made available via Provider.

karcass:
Seems weird that I can change the check boxes randomly

toddapetersen:
> This will definitely affect Agency, as the data collected via Agency is made available via Provider.

Agree Max...  Need to put the breaks on this until we can sync with Agency changes.   It is probably time we re-unite the two API's into a single process anyway...they were never meant to be separate.

karcass:
> Agree Max... Need to put the breaks on this until we can sync with Agency changes. 

It's not really a code change, though, it's a semantic change if I understand correctly.  I'm okay with it as-is.

rf-:
We had consensus about this change on the Provider call today—@toddapetersen and @karcass, do either of you still have any concerns?

karcass:
@rf- works for me

PR #308, opened 2019-05-06
update sherpa gbfs api url

yoimnik:
correcting sherpa gbfs api url in preparation for launch


hunterowens:
@yoimnik I'm getting 404 on that URL... I assume this is because sherpa hasn't launched yet to my knowledge?

yoimnik:
@hunterowens Yes, Sherpa has not launched yet.

karcass:
Looks good to me

hunterowens:
@yoimnik sorry for not merging this earlier. FYI, link is still 404ing

yoimnik:
@hunterowens hey, this does not seem to be 404ing:

https://mds.bird.co/gbfs/platform-partner/sherpa/gbfs.json

the spec says to use that file as a way to explore all other URLs https://github.com/NABSA/gbfs/blob/master/gbfs.md

when looking at this CSV, is it assumed that what is written as the gbfs_url is a base url, and that clients have to append it with `/<file_name>.json` ? i have assumed that, which is why i wrote the url this way

i believe this assumption applies to the `mds_url`, so i assumed the same thing for the gbfs_url

PR #311, opened 2019-05-10
[0.2.x] explicitly state no ordering on data from Provider APIs

yoimnik:
### `Provider` or `agency`

* [x] `provider`
* [ ] `agency`
* [ ] both

### Additional context

Discussed on the MDS Provider web conference 5/9 11am PT


karcass:
I would say there's no point in back-patching 0.2.x, but this is not a strongly-held opinion.

hunterowens:
I don't feel like there is a huge reason to back patch and do another 0.2.x release.

PR #314, opened 2019-05-14
[provider readme] fix version legacy headers

yoimnik:
`Accept: */*` is used by very common prevalent clients like python requests and postman

important to note here when implementing versioning via accept headers to properly support client code


hunterowens:
pinging @yoimnik / @thekaveman on this- any resolution to the above questions?

Issue #287, opened 2019-04-03
Modify README.md to better reflect current state of project


jfh01:
README.md is currently the home page for MDS (and the top Google result when searching for "mobility data specification"). But the page doesn't do a great job at capturing the current state and goals of the project, or showcasing involvement from cities other than LA.

Three proposed changes to README.md:

1) Add a paragraph to tell a little more about the origin story and values of MDS "Created by LADOT to be an open data standard, MDS is currently supported by multiple cities, mobility providers, and software companies. Participation in the project is open to all."

2) Replace the announcements (currently LA-focused) with a directory of links to city-specific micro-mobility pages. Any City that has adopted MDS would be allowed to provide a link to a page on their micromobility program and/or their use of MDS/data.

3) Modify Contact section to reference the MDS provider release process and point to relevant documents for details.

hunterowens:
@jfh01 let me know if the changes in #332 covered this. Going to close for now, but feel free to reopen this one.

PR #320, opened 2019-05-23
Updated README text

jpayne00:
Changes were applied to the README at the request of LADOT Executive Staff.
The intent is to make the language more comprehensible to the general public.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Mobility providers are required to share data with LADOT as part of the City of 
 
 **MDS Components** 
 
+MDS is currently comprised of two distinct components:
+
 * The [`provider`][provider] Application Program Interface (API) was created in May 2018 to be implemented by mobility providers. When a municipality queries information from a mobility provider, the Provider API automatically generates a historical view of operations in a standard format.
 
 * The [`agency`][agency] API was created in April 2019 to be implemented by regulatory agencies. Providers query the Agency API when an event occurs, like a trip starting or ending.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,24 +1,43 @@
...
 
-**MDS** is currently comprised of two distinct components:
+**MDS Components** 
 
-* The [`provider`][provider] API is to be implemented by *mobility as a service* providers, for data exchange and operational information that a municipality will query. `provider` presents the historical view of operations.
+* The [`provider`][provider] Application Program Interface (API) was created in May 2018 to be implemented by mobility providers. When a municipality queries information from a mobility provider, the Provider API automatically generates a historical view of operations in a standard format.
 
-* The [`agency`][agency] API is to be implemented by *municipalities* and other regulatory agencies, for providers to query and integrate with during operations. `agency` provides tools to inform and permit future operations.
+* The [`agency`][agency] API was created in April 2019 to be implemented by regulatory agencies. Providers query the Agency API when an event occurs, like a trip starting or ending.
```
karcass: Hm, I wouldn't say it was created in April of 2019.  It was conceived in the same time frame as Provider.  It did go through some revisions, and it wasn't implemented by LADOT until February of 2019.  April 2019 is when the reference implementation went into production.

hunterowens:
I believe this is related to #287

jannasmith:
Some of this information may be better suited to the ladot.io page about MDS, such as "How MDS is used in Los Angeles" and the LADOT-specific data sharing requirements. Are people pointing to ladot.io as a "general public"-facing landing page, or still this github link? The github readme should be focused on an audience of operators, cities' technical staff, and other potential contributors.

We should leave in the information about what kind of API spec MDS is, and reference to GBFS - these are more beneficial for contributors who want to suggest edits and updates to the spec and cities/operators who want to look at the code itself. This audience also probably doesn't need a lot of detail about LADOT's dockless permitting, etc

README.md
hunterowens: @karcass @jpayne00 how does 

`The [`agency`][agency] API is implemented by the regulatory agency, ie,  LADOT or the County of Los Angeles. Providers query the API during operations to get information from the regulatory agency.`

hunterowens:
closing in favor of #332

PR #322, opened 2019-05-25
Add Policy API draft

karcass:
### Explain pull request

Add new Policy API and associated schema for digitally communicating micromobility policy from Agencies to Providers

### Is this a breaking change

* [ ] Yes, breaking
* [x] No, not breaking
* [ ] I'm not sure

### `Provider` or `agency`

* [ ] `provider`
* [ ] `agency`
* [ ] both
* [x] neither

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -183,9 +183,6 @@ Body Params:
 
 The vehicle `/telemetry` endpoint allows a Provider to send vehicle telemetry data in a batch for any number of vehicles in the fleet.
 
-The Update Telemetry endpoint (/telemetry) shall be called for the specific trip within 24 hrs after the vehicle trip is over.   
-For any given trip, data reported via the (/telemetry) endpoint shall contain temporal and location data for every 300 ft (91 meters) while vehicle is in motion and 30 seconds while at rest. For Mobility Service Providers who do not calculate distance in real-time, a periodic rate of 14 seconds can be used while vehicle is in motion.
-
 Endpoint: `/vehicles/telemetry`
 Method: `POST`
 


karcass:
Feedback solicited!

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+<a name="background"></a>
+## Background
+
+The goal of this specification is to enable Agencies to create, revise, and publish machine-readable micromobility policy, as set of rules for individual and collective micromobility device behavior exhibited by both providers and riders. Examples of policies include:
+
+* City-wide and localized caps (e.g. “Minimum 500 and maximum 3000 scooters within city boundaries”)
+* Exclusion zones (e.g. “No scooters are permitted in this district on weekends”)
+* Cap allowances (e.g. “Up to 500 additional scooters are permitted near train stations”)
+* Speed-limit restrictions (e.g. “15 mph outside of downtown, 10 mph downtown”)
+* Idle-time and disabled-time limitations (e.g. “5 days idle while rentable, 12 hours idle while unrentable, per device”)
```
hunterowens: I think it would be important to add incentive zones to this distribution, as preferred drop offs and preferred pick up zones are crucial

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+A machine-readable format will allow Providers to download policies and compute compliance for policies where it can be determined entirely by data obtained internally.  Providers can then continually measure their own compliance against policies without further API calls.
+
+References to geography (areas, street segments, etc.) will be done via UUID.  Geographic data will be available as GeoJSON via the `/geographies` endpoint.  In a future revision of Agency, we will reconcile this with the existing `/service_areas` endpoint.  `/service_areas` currently only handles GeoJSON MultiPolygon and Polygon objects, and Policies might refer to street segments or points.
```
hunterowens: This should be reconciled with the `service_areas` before we merge, IMO, as it might become unclear which rule to follow.

hunterowens: I feel that we should represent steets as Polygons rather than lines, since they _actually_ are polygons and it could be interpreted as not of the street when one actually is. This could be a use case for OSM or SharedStreets street IDs on a basemap... worth solving and discussing more.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+A machine-readable format will allow Providers to download policies and compute compliance for policies where it can be determined entirely by data obtained internally.  Providers can then continually measure their own compliance against policies without further API calls.
+
+References to geography (areas, street segments, etc.) will be done via UUID.  Geographic data will be available as GeoJSON via the `/geographies` endpoint.  In a future revision of Agency, we will reconcile this with the existing `/service_areas` endpoint.  `/service_areas` currently only handles GeoJSON MultiPolygon and Polygon objects, and Policies might refer to street segments or points.  
+
+This initial draft proposes a subset of possible policies for consideration, and should not be taken to be the a comprehensive enumeration of all possible policies.
+
+<a name="distribution"></a>
+## Distribution
+
+Policies may be published by Agencies or their authorized delegates via JSON objects via REST API.
```
hunterowens: We should make this simpler rather than requiring full REST. Ideally, IMO, a city should be able to have a static endpoint or arcGIS online endpoint that contains a policy page. This will enable all cities to publish policy documents easily.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+Each policy will have a unique ID (UUID).
+
+Published policies should be treated as immutable data.  Obsoleting or otherwise changing policies is accomplished by publishing additional policies with a field named `prev_policies`, a list of UUID references to the previous policy or policies.
```
hunterowens: 👍 

Should `prev_policies` be changed to "prior revision" and be a single item. Essential a linked list starting from policy _current_ to the first.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+<a name="distribution"></a>
+## Distribution
+
+Policies may be published by Agencies or their authorized delegates via JSON objects via REST API.
+
+Each policy will have a unique ID (UUID).
+
+Published policies should be treated as immutable data.  Obsoleting or otherwise changing policies is accomplished by publishing additional policies with a field named `prev_policies`, a list of UUID references to the previous policy or policies.  
+
+Policies should be stored and accessible indefinitely so that the set of active policies at a given time in the past can be retrieved from the `/policies` endpoint.
```
hunterowens: We should add additional flex here so that a city could just host a `policies.json` or similar to lower the barrier to entry.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+Policies will typically be linked to one or more associated geographies.  Geography descriptions (e.g. geofences or lists of street segments) must also be maintained by the Agency indefinitely.  Policies without specific geographies (global policies) are assumed to apply to the entire service area managed by the Agency.
```
hunterowens: Typically or always? Policies must always have an assocated geography, right?

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+Policies should be re-fetched whenever (a) a Policy expires (via its `end_date`), or (b) at an interval specified by the Agency, e.g. "daily at midnight".
+
+
+<a name="schema"></a>
+## Schema
```
hunterowens: We shouldn't merge this until we have a JSON schema, so that is easy to build "policy" validator tools. We might even want to include a script to validate policies in the spec more than just JSON schema. More work I know but I think it is really important that we make sure there is no way to publish a valid policy that can be misinterpreted.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+<a name="policy-fields"></a>
+### Policy Fields
+
+| Name            | Type      | R/O | Description |
```
hunterowens: R/O??

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+| `description`   | String    | R   | Description of policy |
+| `start_date`    | timestamp | R   | Beginning date/time of policy enforcement |
+| `end_date`      | timestamp | O   | End date/time of policy enforcement |
+| `prev_policies` | UUID[]    | O   | Unique IDs of prior policies replaced by this one |
+| `rules`         | Rule[]    | R   | List of applicable rule elements (see [“rule fields”](#rule-fields)) |
+
+Note that the provider-scope is at present not included.  A Provider should not be able to tell whether the policy is specific to them, or to a subset of Provider, or all Providers.
+
+<a name="rule-types"></a>
+### Rule Types
```
hunterowens: per above, I think we need to figure out how to setup incentive zones.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+Note that the provider-scope is at present not included.  A Provider should not be able to tell whether the policy is specific to them, or to a subset of Provider, or all Providers.
+
+<a name="rule-types"></a>
+### Rule Types 
+
+| Name    | Description |
+| ---     | --- |
+| `count` | Fleet counts based on regions.  Rule max/min refers to number of devices. |
+| `time`  | Individual limitations on time spent in one or more vehicle-states.  Rule max/min refers to number of minutes. |
+| `speed` | Global or local speed limits.  Rule max/min refers to miles per hour. |
```
hunterowens: mph or should we allow different units here. I'm not an i18n expert but we should probably do some research on how best to handle this.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+| `maximum`    | integer | O | Maximum value, if applicable (default unlimited) |
+| `start_time` | time    | O | Beginning time-of-day (hh:mm:ss) when the rule is in effect (default 00:00:00) |
+| `end_time`   | time    | O | Ending time-of-day (hh:mm:ss) when the rule is in effect (default 23:59:59) |
+| `days`       | day[]   | O | Days `["sun", "mon", "tue", "wed", "thu", "fri", "sat"]` when the rule is in effect (default all) |
+| `messages`   | { string:string } | O | Message to rider user, if desired, in various languages, keyed by language tag (see [Messages](#messages))|
+| `value_url ` | URL     | O | URL to an API endpoint that can provide dynamic information for the measured value (see [Value URL](#value-url)) |
+
+### Order of Operations
+
+Rules are ordered most-specific to most-general.  E.g. an “earlier” cap rule would take precedence over a “later” cap rule.  The internal mechanics of ordering are up to the Policy editing and hosting software.
```
hunterowens: Seems like this should be enforced and standardized by the spec itself? unless there is some reason I'm not seeing here.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+<a name="messages"></a>
+### Messages
+
+Some Policies as established by the Agency may benefit from rider communication.  This optional field contains a map of languages to be (optionally?) shown to the user.  (Unanswered question: shown with what frequency?  At sign-up?  Once per day?)
```
hunterowens: need to resolve frequency question and mandatory question before merge

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+The payload returned from a GET call to the `value_url` will have the following immutable fields:
+
+| Name         | Type      | R/O | Description |
+| ----         | ----      | --- | ----------- |
+| `value`      | integer   | R   | Value of whatever the rule measures |
+| `timestamp`  | timestamp | R   | Timestamp the value was recorded |
+| `policy_id`  | UUID      | R   | Relevant `policy_id` for reference |
+
+MDS will likely include some sort of streaming mechanism in the upcoming releases that may supplant or replace this mechanism.
```
hunterowens: This seems really hack-y, and unless there is a imminent demand from an agency to implement this, I feel like we should try for a follow up PR with a better solve than this.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+<a name="endpoints"></a>
+## Endpoints
```
hunterowens: This list seems overly complex. In order to make it easier to cities (who might only have a GIS analyst or similar) to be able to implement this, I'd replace it with  the ability to publish a single policy file with associated info.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+Parameters: none
+
+Note: Intentionally omitted `GET /geographies` until a compelling use-case can be identified.
+ 
+<a name="authoring"></a>
+## Authoring
+
+Creating, editing, and publishing policies may be performed via a variety of mechanisms, and are therefore not specified here.  Authoring tools would optionally provide schema extensions for tooling, including author, Provider-specificity, etc.  We may add a specific instance of such extensions in a later revision of this document, but at present that’s TBD.
+
+One mechanism may be via a source-control repository, where pull-requests to Policy objects are proposed, left open to public commentary, etc., and served as static content via the endpoints listed above.
```
hunterowens: I'd remove this couple lines, as this is more relevent for a PR message rather than for actual inclusion in the spec.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+Another possibility would be a policy-editing REST API, where drafts of Policy objects are mutable, pending publication.  This would be the API for manual policy-creation with GUI tooling.  LADOT may propose a specific separate policy-editing API in the future.
+
+Certain policies could be fully dynamic, e.g. caps could be raised and lowered via algorithm on a day-to-day or even hour-to-hour basis.  No-fly-zones could be created in quasi-real-time in response to emergencies or road-closures. 
+
+Dynamic caps can also be implemented by replacing the “maximum” integer with a URL to a source for dynamic data.  This could be for provider-specific caps that go up and down, or for global caps e.g. “total 500 scooters at the coliseum”.  Dynamic data sources would be required to have historical data so that validating prior information, and downloaded dynamic data would have a time-to-live.
+
+<a name="compliance"></a>
+## Compliance
```
hunterowens: Again, we should either solve or remove this from the documentation here.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+A Compliance API will be described in a separate MDS specification.  In brief, it will take as inputs the MDS Agency data stream, the MDS geography data, and these MDS Policy objects and emit Compliance objects.  This work is in draft form but is closely informed by this Policy specification.
```
hunterowens: same comment as above. This seems overly complex

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+        "rule_type": "count",
+        "geographies": ["4c2015c6-6702-48a6-ab58-94d963911182"],
+        "statuses": ["available", "unavailable", "reserved", "trip"],
+        "vehicle_types": ["bicycle", "scooter"],
+        "maximum": 2500
+    }]
+}
+```
+
+Idle time limits example.  Scooters and bikes can be in the public right-of-way for up to three days if rentable, one day if not.
```
hunterowens: 👍 to this example!

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+      "geographies": ["f3637b37-ee80-46e2-b01f-3a7f7f819dd9"],
+      "statuses": ["available", "trip"],
+      "vehicle_types": ["bicycle", "scooter"],
+      "maximum": 300
+    }]
+  }
+}
+```
+
+More examples to come.  :)
```
hunterowens: Remove line from spec here. Commentary.

hunterowens:
also, I'm curious how y'all thought #203 would have worked / should interface with this PR

policy/README.md
karcass: Required/optional

policy/README.md
karcass: Roger

policy/README.md
karcass: Revised, will update PR in a bit

policy/README.md
karcass: Work in progress, stay tuned

policy/README.md
karcass: Agree.  What do you think?  These are open questions.

policy/README.md
karcass: We will definitely need streaming for some upcoming aspects of MDS as a whole.  Can kick that can down the road though.  Updated PR.

policy/README.md
karcass: Line 196 is meant to address your static-served-files ask.  Were you asking to lose line 194 too?

policy/README.md
karcass: I'd like to leave this forward-pointer.  We will have the Compliance API spec soonish, once Policy firms up.

policy/README.md
karcass: Agreed, removing.

policy/README.md
karcass: Good point.  Working on it.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+<a name="rule-fields"></a>
+### Rule Fields 
+
+| Name         | Type     | R/O | Description      |
+| ---          | ---      | --- | ---              |
+| `name`       | String   | R   | Name of cap item |
+| `rule_type`  | enum     | R   | Type of policy (see [“rule types”](#rule-types))|
+| `rule_units` | enum     | O   | Measured units of policy (see [“rule units”](#rule-units))|
+| `geographies`| UUID[]   | R   | List of Geography UUIDs (non-overlapping) specifying the covered geography |
+| `statuses`   | Status[] | R   | Vehicle `status` to which this rule applies.  See [MDS Agency state diagram](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/agency/README.md#vehicle-events). |
```
rf-: I'm still trying to get my head around the implications of this model for expressing rules. It seems well-suited to some use cases, but it seems like it might be problematic to not be able to distinguish between different events that lead to the same status. For example, the agency may want to limit or prohibit provider deployments in a given area but not user dropoffs—is there a way to model that?

policy/README.md
karcass: Service areas will co-exist with Policy until Policy is widely adopted, and will exist in parallel until we decide to fully deprecate them in the future.  The function of service_areas is a subset of what can be expressed in Policy.

Dockless Policy does not currently require references to streets, so we'll remove that for now.  We will be evaluating options for representing street segments in the future.

policy/README.md
karcass: A static endpoint/file would prevent the use of Provider-specific Policy objects, e.g. dynamic caps, which are a key requirement from LADOT.  Agree that a subset of Policy could be done with static files, and that that's a worthy goal.  I'd like to defer discussion of static-endpoint until next release.

policy/README.md
karcass: We started out that way, and one of our internal reviewers noticed that you might have two or more Policy replaced by a single subsequent Policy.

policy/README.md
karcass: See comment above.  Yes, for a subset, but we'll need to think through the implications.

policy/README.md
karcass: Units are specified in the `Rule` object.  `mph` or `kph` for `speed` rules.

policy/README.md
karcass: We are using the fact that JSON has ordered lists to express the order in which the Rules should be evaluated.  Agree that the wording is a bit confusing, will fix that.

policy/README.md
karcass: Amendment, we're going to add all of the current Policy objects in use by LADOT for compliance measurements.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+<a name="rule-fields"></a>
+### Rule Fields 
+
+| Name         | Type     | R/O | Description      |
+| ---          | ---      | --- | ---              |
+| `name`       | String   | R   | Name of cap item |
+| `rule_type`  | enum     | R   | Type of policy (see [“rule types”](#rule-types))|
+| `rule_units` | enum     | O   | Measured units of policy (see [“rule units”](#rule-units))|
+| `geographies`| UUID[]   | R   | List of Geography UUIDs (non-overlapping) specifying the covered geography |
```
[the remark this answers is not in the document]
karcass: I totally get your point.  However, we have a design goal of publishing immutable Policy _and_ immutable Geography objects, so that any point in time in the past can be accurately reconstructed for adjudication of Policy, and our reference implementation enforces that.  Pointers to external data sources open the door to security and cross-site issues.  I could imagine relaxing this in a subsequent revision, after input from a variety of cities interested in using this Policy / Geography paradigm.

policy/README.md
```diff
@@ -0,0 +1,353 @@
...
+| `statuses`   | Status[] | R   | Vehicle `status` to which this rule applies.  See [MDS Agency state diagram](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/agency/README.md#vehicle-events). |
+| `vehicle_types` | VehicleType[] | O | Applicable vehicle categories, default “all”.  See MDS shared data types document.  (link forthcoming) |
+| `minimum`    | integer | O | Minimum value, if applicable (default 0) |
+| `maximum`    | integer | O | Maximum value, if applicable (default unlimited) |
+| `start_time` | time    | O | Beginning time-of-day (hh:mm:ss) when the rule is in effect (default 00:00:00) |
+| `end_time`   | time    | O | Ending time-of-day (hh:mm:ss) when the rule is in effect (default 23:59:59) |
```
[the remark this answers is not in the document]
karcass: will fix

policy/README.md
karcass: We've addressed this particular point in a soon-to-be-amended version of this PR.  Good catch!

policy/README.md
owenoak: FWIW, I found "R/O" confusing as well.  Perhaps "Required" with "Y" and "N" values?

policy/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+Parameters: none
+
+`GET /geographies/{id}`
+
+Returns: One immutable GeoJSON FeatureCollection with an embedded UUID identifier property of `id`
+
+Parameters: none
+
+Note: Intentionally omitted `GET /geographies` until a compelling use-case can be identified.
```
owenoak: From a front-end perspective it would be useful to be able to retrieve a number of geographies in a single request.  This would facilitate, e.g. showing multiple the geographies in a map view quickly.

policy/README.md
karcass: Example included

policy/README.md
karcass: Should be addressed

policy/README.md
karcass: I made a couple changes to support this.  More verbiage coming by end of week.

policy/README.md
karcass: Addressed

policy/README.md
karcass: Schema added

policy/README.md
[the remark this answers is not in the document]
karcass: fixed

policy/README.md
karcass: fixed

karcass:
Updated extensively.  Looking for re-review.  @hunterowens @rf-

evansiroky:
I could see this being useful in a trip planning context. I've been working on improving OpenTripPlanner for the Portland, OR area where we have these manual files that are consumed that represent areas where travel with eScooters is banned and where parking eScooters is banned. It'd be great to have an actual API endpoint to fetch this data from cities either via this proposal or some other kind of GBFS-for-cities kinda thing.

For my two use cases, I guess an agency could model the banned parking area by creating a rule that applies to `trip_end` vehicle statuses and then perhaps model travel not allowed in an area by specifying a maximum value of 0 for a speed rule. 

Anywho, I'm kind of just commenting here to follow along and see where the conversation goes for this topic.

policy/Examples.md
```diff
@@ -0,0 +1,788 @@
...
+
+#### Walkthrough
+
+##### Abstract
+
+The Venice Special Operations Zone Restrictions published by LADOT !FIXME NEED LINK! specifies a series of valid drop off points within a greater overarching no-dropoff zone.
+
+##### Policy Design Approach
+
+The Venice Special Operations Zone Policy consists of two rules, first, a rule which contains geographies representing the 'valid' `provider_drop_off` zones, and a second rule which has the overarching Venice geography. During the evaluation of the policy, a set of vehicles will be provided, we'll refer to this set as `V`. The first rule will match all vehicles which have been `provider_drop_off`'d within one of the given geographies, let's call this group of vehicles `v_1`. When evaluating the second rule, the vehicles which will be considered in the evaluation will be a subset `V \ v_1`. At this point, we are considering a subset which cannot have a vehicle in a valid drop off zone for this policy; and because of this, all vehicles that match with the second rule will be considered in violation of the policy.
```
evansiroky: This explanation of the above rules doesn't make sense to me and it's especially confusing since the JSON of each rule seems essentially identical. Why wouldn't the geography in the 2nd rule just be listed among those in the first rule? And this explanation makes me wonder if somehow the order of the rules are important and should be combined together to produce some kind of hybrid rule.

karcass: The rule order is important (per the spec).  The JSON is similar because it applies to the same types of vehicles and operations (`provider_drop_off`), but different because some of the specified areas are permitted and others aren't.

karcass: The goal of the Policy schema and its corresponding Compliance engine is to express as many different Policy concepts as possible, with a bare minimum of abstractions and exceptions.

evansiroky: Ah, I didn't see that section about how the order of the rules is important. My bad. 

However, it's still not clear how to determine which areas are permitted and which ones aren't because it appears that the only data available in the geographies/{id} endpoint is the id of the geography.

antrim:
Note this GBFS pull request regarding adding geofencing 
https://github.com/NABSA/gbfs/pull/92#issuecomment-531331920.

GBFS is seeking to align semantics with MDS. We'll add notes on this MDS proposal to our discussion and comparison of open proposals for geofencing.

policy/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+| Name         | Type      | R/O | Description                                    |
+| ------------ | --------- | --- | ---------------------------------------------- |
+| `start_time` | timestamp | O   | Earliest effective date, default=effective now |
+| `end_time`   | timestamp | O   | Latest effective date, default=all future      |
+
+Note: provider_id is an implicit parameter and will be encoded in the authentication key
+
+Returns: List of policy objects effective during the timespan described by greater than or equal to `start_time` and less than or equal to `end_time`
+
```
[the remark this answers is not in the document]
karcass: Added flat-file semantics

karcass:
Added flat-file format @hunterowens @jfh01 @rf- @thekaveman @antrim @evansiroky

policy/README.md
```diff
@@ -0,0 +1,296 @@
...
+### Order of Operations
+
+Rules are ordered most-specific to most-general. E.g. an “earlier” cap rule would take precedence over a “later” cap rule. The internal mechanics of ordering are up to the Policy editing and hosting software.
+
+#### Rule Ordering
+
+Rules, being in a list, are implicitly ordered according to the JSON Specification. Rules are a very specific form of pattern matching; you specify the conditions for which a given rule is 'met', and a vehicle (or series of vehicles) may match with that specific rule. If a vehicle is matched with a rule, then it _will not_ be considered in the subsequent evaluation of rules within a given policy. This allows for expressing complex policies, such as a layer of 'valid' geographies in an earlier rule, with overarching 'invalid' geographies in later rules: see [LADOT Venice Beach Special Operations Example](./Examples.md#venice-beach-spec-ops)
+
+##### Evaluation Pseudocode
```
hunterowens: remove this.

policy/README.md
```diff
@@ -0,0 +1,296 @@
...
+	        },
+```
+
+<a name="value-url"></a>
+
+### Value URL
+
+An Agency may wish to provide dynamic or global rules, e.g. "Within 300 yards of the stadium, 1000 total extra scooters may be deployed, across all Provider(s)." In this case, compliance is not computable from the information available to a single Provider. The Agency would provide a specified endpoint to get the current count of vehicles in the service-area, so that individual Providers could decide whether adding some number to those present can be done.
+
+There are potential complexities that are as-yet unaddressed, such as the latency between a Provider deciding to enter such an area, and the number of vehicles in the area when the scooters arrive. As this specification is Alpha, feedback and refinements are welcome.
```
rf-: Now that this PR is close to being merged, what's the plan for aspects like this that are called out as not being fully baked yet?

policy/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+One mechanism may be via a source-control repository, where pull-requests to Policy objects are proposed, left open to public commentary, etc., and served as static content via flat files or the endpoints listed above.
+
+Another possibility would be a policy-editing REST API, where drafts of Policy objects are mutable, pending publication. This would be the API for manual policy-creation with GUI tooling. LADOT will propose a specific policy-editing API in the future.
+
+Certain policies could be fully dynamic, e.g. caps could be raised and lowered via algorithm on a day-to-day or even hour-to-hour basis. No-fly-zones could be created in quasi-real-time in response to emergencies or road-closures.
+
+Dynamic caps can also be implemented by replacing the “maximum” integer with a URL to a source for dynamic data. This could be for provider-specific caps that go up and down, or for global caps e.g. “total 500 scooters at the coliseum”. Dynamic data sources would be required to have historical data so that validating prior information, and downloaded dynamic data would have a time-to-live.
```
rf-: I'd suggest removing this whole section before merge. This is a spec for the API—it doesn't really seem necessary to list various ways it could be implemented.

Also, it seems like there's no provision in the spec right now for historical queries against the dynamic cap data. Is that something you're thinking about adding still?

policy/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+<a name="compliance"></a>
+
+## Compliance
+
+A Compliance API will be described in a separate MDS specification. In brief, it will take as inputs the a snapshot of the MDS status at a particular time, the MDS geography data, and these MDS Policy objects and emit Compliance JSON measurements. MDS status can be generated from either Provider or Agency data.  This work is in draft form but is closely informed by this Policy specification.
```
rf-: This also seems like something that would make sense to take out before merge. I wonder if it would make sense to think about some kind of light RFC process for future proposals, since information like this is valuable for people who are assessing the proposal but doesn't seem like it really belongs in the spec itself.

policy/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Order of Operations
+
+Rules are ordered most-specific to most-general. E.g. an “earlier” cap rule would take precedence over a “later” cap rule. The internal mechanics of ordering are up to the Policy editing and hosting software.
+
+#### Rule Ordering
+
+Rules, being in a list, are implicitly ordered according to the JSON Specification. Rules are a very specific form of pattern matching; you specify the conditions for which a given rule is 'met', and a vehicle (or series of vehicles) may match with that specific rule. If a vehicle is matched with a rule, then it _will not_ be considered in the subsequent evaluation of rules within a given policy. This allows for expressing complex policies, such as a layer of 'valid' geographies in an earlier rule, with overarching 'invalid' geographies in later rules: see [LADOT Venice Beach Special Operations Example](./Examples.md#venice-beach-spec-ops)
```
rf-: This describes how rules are prioritized within a given policy, but what happens when rules belonging to different policies collide with each other? It doesn't seem like there's a way for the server to control the resolution of such conflicts since policies are returned in time order instead of priority order.

policy/README.md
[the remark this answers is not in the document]
karcass: @rf- See update to include flat-file format

policy/README.md
```diff
@@ -0,0 +1,257 @@
...
+| --------------- | --------------------------- | --- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `name`          | String                      | R   | Name of cap item                                                                                                                                                                                                                                                                                                                                        |
```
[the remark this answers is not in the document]
karcass: Fixed

policy/README.md
```diff
@@ -0,0 +1,257 @@
...
+| `statuses`      | { Status: Vehicle Event[] } | R   | Vehicle `statuses` to which this rule applies. Optionally, you may provide specific `event_type`'s for the rule to apply to as a subset of a given status, providing an empty list or null defaults to "all". See [MDS Agency state diagram](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/agency/README.md#vehicle-events). |
+| `vehicle_types` | VehicleType[]               | O   | Applicable vehicle categories, default “all”. See MDS shared data types document. (link forthcoming)                                                                                                                                                                                                                                                    |
```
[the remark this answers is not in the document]
karcass: fixed

policy/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+The provider-facing Policy API consists of the following two endpoints.
+
+`GET /policies`
+
+Parameters:
+
+| Name         | Type      | R/O | Description                                    |
+| ------------ | --------- | --- | ---------------------------------------------- |
+| `start_time` | timestamp | O   | Earliest effective date, default=effective now |
```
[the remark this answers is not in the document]
karcass: fixed

Issue #281, opened 2019-03-27
Use OpenAPI to define MDS API request shapes and query parameters


dyakovlev:
The existing documentation of query parameters and headers is in narrative format in the Provider and Agency README files. This is useful to give an overview of what's allowed, but because it is not programmatically enforced and left up to interpretation, it's common to see divergences from the spec. [1]

I'd like to propose adding API endpoint parameter definitions to Provider (and maybe Agency?) endpoints using a format designed to support automatic validation. 

Implementers of MDS endpoints and software that interacts with them would be able to use these definitions to validate the shape of query parameters and headers for incoming requests. Additionally, a rigorous definition of the request shape would let MDS evolve to support more advanced request formats in a safe and transparent way.

A framework frequently used for this kind of schema is [OpenAPI v3](https://swagger.io/docs/specification/describing-parameters/), an evolution of Swagger API definition formats. It has a lot of OSS and third party tools for API documentation and validation that may come in useful, and is reasonably well documented. Based on prior experience, I'd recommend using this format. [2]

---

[1] A similar thing to this already exists for validating the shape of responses via JSON Schema, and has been extremely useful for deciding if an API response conforms to the specification. However, JSON Schema is a toolkit for asserting the shape of JSON objects and  does not really support defining the shape of request parameters and headers. 

[2] I also considered suggesting [JSON Hyper Schema](https://json-schema.org/latest/json-schema-hypermedia.html) ([narrative intro](https://blog.apisyouwonthate.com/getting-started-with-json-hyper-schema-184775b91f)), an extension to JSON Schema that builds a vocabulary for defining URI shapes. However, it is very young and does not have a really mature ecosystem of tools around it yet.

dyakovlev:
notes from 5/23 call

- general consensus for need
- check that OpenAPI can support validating bulk data

some good thoughts about tool applications in the readme here - https://github.com/wework/json-schema-to-openapi-schema

craastad:
I'm all for this, especially the choice of using the OpenAPI or JsonSchema tooling. I found the PHP tooling with JsonSchema to have issues with performance and completeness (i.e. adding massive overhead to requests to build responses and not not rendering nulls in paging params). I find the tooling to be better with OpenAPI. Please reach out if you'd like help with implementing or validating this change.

hunterowens:
Any progress on this @dyakovlev @craastad ?

dyakovlev:
not from my side, sorry

[...]

PR #325, opened 2019-05-31
Create Data Definitions and Common Standards

whereissean:
Document a common set of data definitions and standards between MDS components.

### Explain pull request

### Is this a breaking change

* [ ] Yes, breaking
* [x] No, not breaking
* [ ] I'm not sure

### `Provider` or `agency`

* [ ] `provider`
* [x] `agency`
* [ ] both


hunterowens:
given that all the data definitions are being made in the `schema` directory, and are concated together by `generate_schema.py`, I think we shouldn't have a 2nd or 3rd place for information to live as a source of truth. 

Ideally, I think we need to implement #281 to get to full data definitions.

hunterowens:
closing for now but happy to reconsider this approach

PR #327, opened 2019-06-06
Update the 406 response when an invalid version is requested.

billdirks:
The specification has a vague statement about returning a list of available versions when the API returns a 406. Since we now have a standard way to negotiate versions, this PR updates the text to refer to that. This was discussed in the MDS provider specification meeting on 6/6/2019.

https://github.com/CityOfLosAngeles/mobility-data-specification/issues/313


hunterowens:
Thanks @billdirks! Merging

PR #329, opened 2019-06-11
Add `publication_time` fields to trips and status changes

### LINKED ISSUE #307: Provider API: Add a "recorded" field to status changes

Polling status changes using the timestamp field is unreliable: an event can be recorded late, and ignored by the polling system if it uses the timestamp of the last polled event to fetch new ones.

Adding a `recorded` field in the status changes, and a query parameter in the `status_changes/` endpoint to query the events using this information.

### Is this a breaking change

* [x] Yes, breaking
* [ ] No, not breaking
* [ ] I'm not sure

### `Provider` or `agency`

* [x] `provider`
* [ ] `agency`
* [ ] both

A global sequence ID seems less natural and would not convey the additional "time of recording" information.

### Additional context

The idea has been tested with our partners and is satisfactory. We would love to see it in MDS 0.4

rf-:
These fields will be optional in 0.3.2 and mandatory in a future version. They reflect the time that the provider made the relevant data available through the API.

This commit partially fixes #307, but it's not a complete fix since it doesn't include the ability to query by publication time.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `provider`

### Additional context

The next minor version should make these fields mandatory and also add query params allowing clients to use them for filtering.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -191,6 +191,7 @@ Schema: [`trips` schema][trips-schema]
 | `accuracy` | Integer | Required | The approximate level of accuracy, in meters, of `Points` within `route` |
 | `start_time` | [timestamp][ts] | Required | |
 | `end_time` | [timestamp][ts] | Required | |
+| `publication_time` | [timestamp][ts] | Optional | Date/time that trip became available through the trips endpoint |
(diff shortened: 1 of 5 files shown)


hunterowens:
Thanks @rf-

PR #332, opened 2019-06-27
Readme updates 2019

hunterowens:
### Explain pull request

This PR incorporates PRs #320/#323 from @jpayne00, along with changes to address @thekaveman and @karcass's questions.  The PR is slightly out of date already given the launch of the OMF, but better to get this cleanup done now and a subsequent one regarding OMF. 

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * both

### Additional context

Hopefully, this address the issues that @jfh brough up in #287. If not, we can do another PR as well.


hunterowens:
@thekaveman @karcass can you two review and see if I missed any of your comments while cleaning up @jpayne00's PR?

thekaveman:
FWIW, this looks good 😃

PR #336, opened 2019-07-11
Support for multiple modes (255 redux)

hunterowens:
### Explain pull request

This PR clarifies that the current provider endpoints and JSON schema are dockless bike/scooter share only. It is a redux of #255 based on everybody's suggestions. 

### Is this a breaking change

 * Yes, breaking - we move the JSON Schema. 

### `Provider` or `agency`

 * `provider`


thekaveman:
I know back on the 5/23 call (at least as documented in the [comments on #255](https://github.com/CityOfLosAngeles/mobility-data-specification/pull/255#issuecomment-495344524)) the idea of outputting all schemas into the root of `/provider` was decided upon. I can't remember all the details of that conversation (I recall that I may have even been apathetic to the choice). But seeing the changes in-situ and thinking about it more, the sub-directory approach seems like it might make more sense.

Especially in light of [@jfh01 comments from a prior (4/4) call](https://github.com/CityOfLosAngeles/mobility-data-specification/pull/255#issuecomment-480926192) and some of what we started talking about related to #281 around a hierarchical API - without knowing more on the OpenAPI side, my feeling is sub-directories would allow for better organization and clearer lines of separation.

rf-:
@thekaveman When you say "sub-directory approach", do you mean an approach like this PR but outputting the schemas into a mode-specific subdirectory? Or do you mean more like #255 where both the schemas and READMEs are duplicated between modes?

thekaveman:
@rf- 

> an approach like this PR but outputting the schemas into a mode-specific subdirectory

Exactly! So instead of e.g. `/provider/dockless_trips.json` we have `/provider/dockless/trips.json`

rf-:
That sounds good to me :+1:

hunterowens:
Fixes #170 !

PR #356, opened 2019-09-17
Adding min and max values for GPS coordinates in common JSON schema

sweis:
### Explain pull request

Several telemetry fields lack input validation supported by json-schema. Currently, the API will accept invalid GPS coordinates. 

This change restricts telemetry GPS latitude and longitude values to be valid coordinates of [-90,90] for latitude and [-180,180] for longitude. JSON validation can reject malformed GPS coordinates rather than propagate bad data.

### Is this a breaking change

This is a **non-breaking change for code**.

It is a potentially **breaking change for legacy bad data** with invalid GPS coordinates. That's ultimately the intent. City agencies will need to assess whether there is a practical impact based on the quality of existing data.

### `Provider` or `agency`

This pull request impacts the **Agency API's** vehicle event and telemetry POST endpoints.

### DIFF
--- a/agency/post_vehicle_event.json
+++ b/agency/post_vehicle_event.json
@@ -176,11 +176,15 @@
           "properties": {
             "lat": {
               "type": "number",
-              "description": "Latitude of the location"
+              "description": "Latitude of the location",
+              "minimum": -90,
+              "maximum": 90
             },
             "lng": {
               "type": "number",
-              "description": "Longitude of the location"
+              "description": "Longitude of the location",
+              "minimum": -180,
+              "maximum": 180
             },
             "altitude": {
               "type": "number",
--- a/agency/post_vehicle_telemetry.json
+++ b/agency/post_vehicle_telemetry.json
@@ -31,11 +31,15 @@
           "properties": {
             "lat": {
               "type": "number",
-              "description": "Latitude of the location"
+              "description": "Latitude of the location",
+              "minimum": -90,
+              "maximum": 90
             },
             "lng": {
               "type": "number",
-              "description": "Longitude of the location"
+              "description": "Longitude of the location",
+              "minimum": -180,
+              "maximum": 180
             },
             "altitude": {
               "type": "number",
--- a/provider/dockless/status_changes.json
+++ b/provider/dockless/status_changes.json
@@ -22,9 +22,19 @@
         "coordinates": {
           "type": "array",
           "minItems": 2,
-          "items": {
-            "type": "number"
-          }
+          "items": [
+            {
+              "type": "number",
+              "minimum": -180.0,
+              "maximum": 180.0
+            },
+            {
+              "type": "number",
+              "minimum": -90.0,
+              "maximum": 90.0
+            }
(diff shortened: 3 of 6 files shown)


thekaveman:
This seems like a useful change for data validation. I wonder how we might support this on the `provider` side as well, given that we [piggyback on the canonical GeoJSON schemas](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/schema/generate_schema.py#L19) for location data in that schema. Is this something you would be willing to look into / add to this PR @sweis?

sweis:
@thekaveman, I'd say let's punt trying to fix GeoJSON to another PR. It's more involved than adding some min/max values.

GeoJSON specifies that its coordinates may be "longitude and latitude, or easting and northing" (UTM): https://tools.ietf.org/html/rfc7946#section-3.1.1 . It would have been nice if they typed GeoJSON so you couldn't mix up two different coordinate systems in the same field, but they did not.

The scales and signs of both coordinate systems are not compatible. Json-Schema's numeric validation only supports basic range boundaries, so isn't going to be able to check two different, incompatible ranges: http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.2.

I don't think there is a prospect of getting geojson.org to add validators as long as they want to support multiple coordinate systems.

I thought about if you could modify a schema's existing fields using Json-schema's structuring: https://json-schema.org/understanding-json-schema/structuring.html#extending . I don't think it has the capacity to replace fields and can only graft in new fields to an existing schema. I might be wrong.

That leaves you with the option to fork or reproduce GeoJSON, which isn't great.

thekaveman:
You're right about GeoJSON's spec @sweis; fortunately MDS Provider has taken a more [prescriptive approach for Geographic data](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/provider/README.md#geographic-data).

We already tweak the GeoJSON schemas to fit our needs, I was only suggesting additional tweaks to enforce the lat/lon bounds along the lines of (in [`get_point_schema`](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/schema/generate_schema.py#L19)):

```python
def get_point_schema():
    """
    Get the canonical schema for a GeoJSON point.
    """
    p = requests.get("http://geojson.org/schema/Point.json")
    point = p.json()
    # Modify some metadata
    point.pop("$schema")
    point["$id"] = get_definition("point")
    # enforce lat/lon bounds
    point["properties"]["coordinates"]["maxItems"] = 2
    point["properties"]["coordinates"]["items"] = [
        {
          "type": "number",
          "minimum": -180.0,
          "maximum": 180.0
        },
        {
           "type": "number",
           "minimum": -90.0,
           "maximum": 90.0
        }
    ]
    return point
```

This is using the [Tuple validation](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation) feature in JSON Schema arrays.

sweis:
Good idea, @thekaveman. I added another commit to set the properties in generate_schema.py. Works fine.

PR #357, opened 2019-09-17
Allow static file data storage for Provider /status_changes and /trips

### LINKED ISSUE #268: Discussion: make MDS queries more predictable
I've been maintaining the MDS provider API at Lime since earlier this year. We've seen the latency across our API endpoints creep up on us as more agencies have adopted MDS and as more trips have been taken and added to our trips and status changes datasets. 

We'd like to decrease latency as much as possible, but we've had some issues doing this because the datasets returned by the API are difficult to precompute and cache. Particularly, being able to query across arbitrary start and end times and having down-to-the-second query values means that we can't reliably cache an entire trip or status change dataset ahead of time. Users can make a query for a minute or month's worth of data, and we have to generate the results on the fly. 

I'd like to brainstorm ways we can make MDS queries more predictable. In particular, if we could present a way for a user to ask for a specific day or hour of data, it would allow us to resolve the query ahead of time and return the result to them much faster.

### LINKED ISSUE #350: Remove vehicle_id and device_id params from Trips endpoint

On the [2019.08.08 call](https://github.com/CityOfLosAngeles/mobility-data-specification/wiki/Web-conference-notes,-2019.08.08), it turns out nobody is using the vehicle_id and device_id query params and it is essentially limited to the time params. Removing this will allow caching to be more easily implemented and enable a solution to #268 

Remove unused query params 
### Is this a breaking change

 * Yes, breaking

### `Provider` or `agency`

 * `provider`

### LINKED ISSUE #385: Clearer language for the Provider's time range query params
The language of the provider spec isn't clear on whether the time range query parameters for the `/trips` and `status_changes` endpoints are required or optional. If optional, the spec doesn't outline the default behavior for the data returned, even if paging is implemented. Should it be returning data since the beginning of time (with pagination)? Should it default to a given time period (example: last 24 hours)?

As the data exposed by providers grows, pagination becomes costly for very large time ranges. Some clarity in the spec would make it very helpful to optimize those edge cases.

The provider spec should clearly outline whether the query parameters are optional or required, and whether it is ok for providers to return an error if they are not provided. If optional, the spec should clearly outline the default behavior.

### Is this a breaking change
Potentially yes, for clients that expect to send requests without time range query parameters.

### `Provider` or `agency`
Provider

We are currently considering an implementation that makes those fields required, but would like to confirm that this strategy does not violate the provider spec.

hunterowens:
### Explain pull request

In #268, @johnpena discusses the need to make MDS queries more predictable. @babldev attempted to address the issue in #354, but during the discussion on 9-4, it sounded like folks wanted a different type of solution. 

This builds on #354 and specifically @hyperknot's suggestion that we split the possible response with different query parameters. 

It introduces `static_end_time` to both `/trips` and `/status_changes`. This would be an ISO month, day, or hour that could be cached on S3 or a similar static file store that could be returned as a single page or set of pages. 

TBH, adding this complexity makes me want to also finish the openAPI work so that this can be made more clearly and testable by providers and clients alike. I'll take a stab at that shortly, but would love folks thoughts on this. 
### Is this a breaking change

 * Yes, breaking

### `Provider` or `agency`

 * `provider`

### Additional context

Fixes #268.
Fixes #350.
Fixes #385.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -230,6 +230,8 @@ Schema: [`trips` schema][trips-schema]
 
 The trips API should allow querying trips with the following query parameters. 
 
+Without an `end_time` query parameter, `/trips` shall return an error. 
+
 * `end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing an Hour. Timezone UTC. 
 
 | Request range | format | expected output | 
@@ -342,6 +344,8 @@ The status_changes API should allow querying status changes with a combination o
 | Request range | format | expected output | 
 | ------------- | ------ | --------------- | 
 | Hour          | YYYY-MM-DD:HH | That hour worth of status changes, or if the hour is not complete or the file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denotated as the next day's zero hour. | 
+
+Without an `end_time` query parameter, `/status_changes` shall return an error. 
 
 ### Event Types
 
--- a/provider/README.md
+++ b/provider/README.md
@@ -342,15 +342,15 @@ Because of the unreliability of device clocks, the Provider is unlikely to know 
 
 ### Status Changes Query Parameters
 
-The status_changes API should allow querying status changes with a combination of query parameters.
+The `/status_changes` API should allow querying status changes with the following query parameters:
 
-* `end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing a hour. The response will have all status changes whose event time occurred within the hour. For example, if you requested "2019-10-01T07", you should get back all status that happened between `7:00 <= and < 8:00`.Timezone UTC. 
+| Parameter | Format | Expected Output |
+| --------------- | ------ | --------------- |
+| `event_time` | `YYYY-MM-DDTHH`, an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) extended datetime representing an UTC hour between 00 and 23. | All status changes with an event time occurring within the hour. For example, requesting `event_time=2019-10-01T07` returns all status changes where `07:00 <= trip.end_time < 08:00` on October 01, 2019 UTC. |
 
-| Request range | format | expected output | 
-| ------------- | ------ | --------------- | 
-| Hour          | YYYY-MM-DDTHH | That hour worth of status changes, or if the hour is not complete or the file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denoted as the next day's zero hour. | 
+If the data does not exist or the hour has not completed, `/status_changes` shall return a `404 Not Found` error.
 
-Without an `end_time` query parameter, `/status_changes` shall return an error. 
+Without an `event_time` query parameter, `/status_changes` shall return a `400 Bad Request` error.
 
 ### Event Types
 
--- a/provider/README.md
+++ b/provider/README.md
@@ -252,15 +252,15 @@ Schema: [`trips` schema][trips-schema]
 
 ### Trips Query Parameters
 
-The trips API should allow querying trips with the following query parameters. 
+The `/trips` API should allow querying trips with the following query parameters:
 
-Without an `end_time` query parameter, `/trips` shall return an error. 
+| Parameter | Format | Expected Output |
+| --------------- | ------ | --------------- |
+| `end_time` | `YYYY-MM-DDTHH`, an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) extended datetime representing an UTC hour between 00 and 23. | All trips with an end time occurring within the hour. For example, requesting `end_time=2019-10-01T07` returns all trips where `07:00 <= trip.end_time < 08:00` on October 01, 2019 UTC. |
 
-* `end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing an Hour. The response will have all trips whose end time occurred in the hour. For example, if you requested "2019-10-01T07", you should get back all trips that ended between `7:00 <= and < 8:00`. Timezone UTC. 
+If the data does not exist or the hour has not completed, `/trips` shall return a `404 Not Found` error.
 
-| Request range | format | expected output | 
-| ------------- | ------ | --------------- | 
-| Hour          | YYYY-MM-DDTHH | That hour worth of trips, or if the hour is not complete or the file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denoted as the next day's zero hour. | 
+Without an `end_time` query parameter, `/trips` shall return a `400 Bad Request` error.
 
 For the near-ish real time use cases, please use the [events](#events) endpoint. 
 
--- a/provider/README.md
+++ b/provider/README.md
@@ -337,19 +337,11 @@ Because of the unreliability of device clocks, the Provider is unlikely to know 
 
 The status_changes API should allow querying status changes with a combination of query parameters.
 
-* `start_time`: filters for status changes where `event_time` occurs at or after the given time
-* `end_time`: filters for status changes where `event_time` occurs before the given time
-* `static_end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing either a Month, Day, or Hour. Timezone UTC. 
+* `end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing a hour. Timezone UTC. 
 
 | Request range | format | expected output | 
 | ------------- | ------ | --------------- | 
-| Month         | YYYY-DD | That month's worth of status changes, or if the month is not complete or the file does not exist , an error | 
-| Day           | YYYY-MM-DD | That day's worth of status changes, or if the day is not complete or the file does not exist, an error | 
-| Hour          | YYYY-MM-DD:HH | That hour worth of status changes, or if the hour is not complete or hte file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denotated as the next day's zero hour. | 
-
-The `min_end_time` and `max_end_time` endpoints shall not be used in combination with the `static_end_time` endpoints, rather, the the former is good for real-ish time uses cases, while the latter is best for reconstructing the historical record and backfilling. It is considered permissive to return an incorrect query parameter responses for the `min_end_time`  / `max_end_time` if the range contains timestamps older than one month. 
-
-When multiple query parameters are specified, they should all apply to the returned status changes. For example, a request with `?start_time=1549800000000&end_time=1549886400000` should only return status changes whose `event_time` falls in the range `[1549800000000, 1549886400000)`.
+| Hour          | YYYY-MM-DD:HH | That hour worth of status changes, or if the hour is not complete or the file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denotated as the next day's zero hour. | 
 
 ### Event Types
 
@@ -382,6 +374,8 @@ All MDS compatible `provider` APIs must expose a public [GBFS](https://github.co
   - `station_information.json` and `station_status.json` don't apply for MDS
 
 ### Events
+
+The events endpoint is a near-ish real time feed of events, designed to give access to as recent as possible series of events from the vehicle feed. It functions similarly to status chan
 
 [Top][toc]
 
--- a/provider/README.md
+++ b/provider/README.md
@@ -375,7 +375,36 @@ All MDS compatible `provider` APIs must expose a public [GBFS](https://github.co
 
 ### Events
 
-The events endpoint is a near-ish real time feed of events, designed to give access to as recent as possible series of events from the vehicle feed. It functions similarly to status chan
+The events endpoint is a near-ish real time feed of events, designed to give access to as recent as possible series of events from the vehicle feed. It functions similarly to status changes, but shall not included data older than 2 weeks (that should live in `/status_changes.`)
+
+Unless stated otherwise by the municipality, this endpoint must return only those events with a `event_location` that [intersects](#intersection-operation) with the [municipality boundary](#municipality-boundary).
+
+> Note: As a result of this definition, consumers should query the [trips endpoint](#trips) to infer when vehicles enter or leave the municipality boundary.
+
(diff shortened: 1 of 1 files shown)


hyperknot:
Thanks for this PR, it looks really simple and clean to me! One tiny addition I'd include for trips is something like "month's worth of trips (filtered by end_time)", just to make it clear that it's the exact same filtering as in min-max end time, _not_ start time.

ascherkus:
Thanks for writing this up!

My three highest bits of feedback surround:
  1. Whether it's possible to reduce the different time periods (month/day/hour) into a single time period for requesting static data to lower specification complexity and implementation burden
  2. Whether the `static_end_time` response requires / assumes fully complete sets of immutable data (vs. how should we handle mutable data)
  3. Whether it's possible to fold `{min,max}_end_time` into `static_end_time` by combining ideas from (1) and (2)


For (1):
  - The argument (IIRC) for allowing different time periods is to simplify backfill consumption by requesting larger periods of time and thus reducing number of requests
  - That being said... as long as response pagination is allowed there's no guarantee the provider implementation will honor a single request/response for a particular time period
  - For example, if a provider implements the "month" response by paginating by day (which they're permitted to do), we haven't reduced the number of requests compared to the "day" method and thus added more complexity to the specification and each implementation with zero added benefit

In short, I believe pagination is in conflict with the desire to request larger time periods / lower the number of requests because the end result will always be subject to how different providers paginate their data, *not* the duration of the time period being requested.

That leads me to believe that either:
  - We pick a single sensible time period (since responses will get arbitrarily paginated anyway)
  - We disallow pagination on these responses (to respect the use case for minimizing # of requests)

Would love to hear what folks think and whether I missed different use cases for having different time periods!


For (2):
  - The argument for requiring complete and immutable sets of data would clarify expectations for returning errors (e.g., an error is guaranteed to be returned for the month "2019-09" until 2019-10-01 00:00:00 UTC at the earliest)
  - Conceptually, once the data "goes live" a consumer could *feel* more confident that the data set is complete
  - That being said... I'm highly skeptical that providers can guarantee historical data would never change once being published (e.g., a bug gets fixed, erroneous data gets corrected)
  - That means consumers need to check for changes in static data when backfilling to ensure they have the latest copy
  - Which leads me to believe we'd want to have requests and responses leverage `If-Match` / `ETag` / `If-Modified-Since` / `Last-Modified` to efficiently determine if changes have been made when backfilling

So I guess it boils down to how confident we all feel whether historical data can/should ever change, and if our collective confidence is not 100% how do we want providers and consumers to handle that scenario? How important is it to ensure we get this right?



For (3):
  - If we're in agreement on (1) and (2) that we only need to specify a single time period and that static data can change over time... then I see an opportunity to support "real-ish time use cases" by replacing `{min,max}_end_time` by having clients simply poll the most current `static_end_time`, using `ETag` / `Last-Modified` to determine when they've been updated
  - This would imply allowing incomplete data sets for the latest time period for `static_end_time` (e.g., requesting "2019-09" before 2019-10-01 00:00:00 UTC is permissible but the data / `ETag` / `Last-Modified` may keep changing until 2019-10-01 00:00:00 UTC at the earliest)
  - For example, a provider could re-generate the latest `static_end_time` period of data every N minutes, with real-time consumers checking every M minutes and fetching the latest data when needed
  - This would ease specification complexity for consumers and providers by folding the three different request/response methods (`static_end_time`, `{min,max}_end_time` < 1 month, `{min,max}_end_time` >= 1 month) into a single method

I understand this is a larger (and likely more controversial!) change, but if we're going to make a breaking change I'd like to entertain the idea of just how large of a breaking change we can make so that provider continues to scale for both consumers and implementors alike :)

kheraankit:
What should be behavior if none of the date params are specified?  I would expect it to return an error and not have any implicit logic.   

Can the spec describe the default behavior for this condition?

johnpena:
Overall this looks really good and I think is a huge step in the right direction, thanks for putting this together @hunterowens!

To expand on what @kheraankit said, I think we should take the opportunity for the next release to specify whether the time-bounding query parameters for the endpoints (`min_end_time`/`max_end_time` and `start_time`/`end_time`) are optional/required and what happens if they're not provided. Lime's feeds work such that when none of the parameters are provided, we return all of the data we have available. I think changing this behavior to return an error would be better, but we should indicate in the documentation that this is the case.

provider/README.md
```diff
@@ -201,13 +201,17 @@ Schema: [`trips` schema][trips-schema]
...
 * `max_end_time`: filters for trips where `end_time` occurs before the given time
+* `static_end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing either a Month, Day, or Hour. Timezone UTC. 
 
-When multiple query parameters are specified, they should all apply to the returned trips. For example, a request with `?min_end_time=1549800000000&max_end_time=1549886400000` should only return trips whose end time falls in the range `[1549800000000, 1549886400000)`.
+| Request range | format | expected output | 
+| ------------- | ------ | --------------- | 
+| Month         | YYYY-DD | That month's worth of trips, or if the month is not complete or the file does not exist , an error | 
+| Day           | YYYY-MM-DD | That day's worth of trips, or if the day is not complete or the file does not exist, an error | 
+| Hour          | YYYY-MM-DD:HH | That hour worth of trips, or if the hour is not complete or hte file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denotated as the next day's zero hour. |
```
jrheard: Small nitpicks: "hte file" -> "the file", also I think "denotated" should be "denoted"

provider/README.md
```diff
@@ -201,13 +201,17 @@ Schema: [`trips` schema][trips-schema]
...
+| Request range | format | expected output | 
+| ------------- | ------ | --------------- | 
+| Month         | YYYY-DD | That month's worth of trips, or if the month is not complete or the file does not exist , an error | 
+| Day           | YYYY-MM-DD | That day's worth of trips, or if the day is not complete or the file does not exist, an error | 
+| Hour          | YYYY-MM-DD:HH | That hour worth of trips, or if the hour is not complete or hte file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denotated as the next day's zero hour. | 
 
+The `min_end_time` and `max_end_time` endpoints shall not be used in combination with the `static_end_time` endpoints, rather, the the former is good for real-ish time uses cases, while the latter is best for reconstructing the historical record and backfilling. It is considered permissive to return an incorrect query parameter responses for the `min_end_time`  / `max_end_time` if the range contains timestamps older than one month.
```
jrheard: Small nitpick: I think that "uses cases" should be "use cases" and that "permissive" should be "permissible".

I'm actually not sure what "It is considered permissive to return an incorrect query parameter responses for the `min_end_time`  / `max_end_time` if the range contains timestamps older than one month." means - what's an "incorrect query parameter response"? If you're referring to a response with a particular HTTP status code (maybe 400?), I think it would be helpful to directly state that status code here. Thanks!

hyperknot:
Based on my experience with APIs and how providers implement them, I think it'd be a good idea:
- To name these two endpoints differently, for example **static_trips** or **processed_trips**. Architecturally these will be very different implementations, the normal endpoints will do a query in a database while these ones will proxy a static file from an S3 bucket for example.
- Explicitly state that these endpoints contain "final" or processed data. Providers usually take 1-2 days till they finalize their data even today, end this is a very buggy area today: there is simply no way to know if a provider's data is "final" in an endpoint or not.
- Also explicitly state an error message, like {message: "not_yet_processed"} or similar to tell clients when a data has not yet been processed.

About time buckets, I also believe it'd be better to ask only one time range, not 3. With 3, I believe it'll be very common that they'll be out-of-sync, something which will just introduce bugs again. Daily splits do not really work, as they are split by utc days, whereas every processing happens in a region's local time, so it's not really useful. Monthly buckets will probably be too big to return them in a single request. 

=> I recommend only implementing **hourly** static endpoints.

provider/README.md
hunterowens: removed in new draft, thanks for the feedback @jrheard

hunterowens:
Hi all: 

I've pushed a new draft, that has some pretty major changes. 

1) the current trips and status_changes become static file only backed, on hourly buckets only, per @hyperknot and others suggests. 

2) there is a new API endpoint, per @thekaveman others suggestion, called `events,` that represents an as real time as possible endpoint, and only contains the last 2 weeks worth of data. Currently, it inherits the `status_changes` schema, but could change in future versions.   

This allows us to handle a lot of the GBFS like-use cases that @HeidiMG has been (correctly, IMO) talking about moving away from GBFS to MDS. 

Thoughts? @johnpena @babldev @rf- @hyperknot @thekaveman @barbeau / anybody else I forgot to tag (apologies!)

provider/README.md
```diff
@@ -199,19 +201,16 @@ Schema: [`trips` schema][trips-schema]
...
-* `static_end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing either a Month, Day, or Hour. Timezone UTC. 
+* `end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing an Hour. Timezone UTC. 
 
 | Request range | format | expected output | 
 | ------------- | ------ | --------------- | 
-| Month         | YYYY-DD | That month's worth of trips, or if the month is not complete or the file does not exist , an error | 
-| Day           | YYYY-MM-DD | That day's worth of trips, or if the day is not complete or the file does not exist, an error | 
-| Hour          | YYYY-MM-DD:HH | That hour worth of trips, or if the hour is not complete or hte file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denotated as the next day's zero hour. | 
+| Hour          | YYYY-MM-DD:HH | That hour worth of trips, or if the hour is not complete or the file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denoted as the next day's zero hour. |
```
ascherkus: typo: "denotated" -> "denoted"

provider/README.md
```diff
@@ -344,13 +335,48 @@ When multiple query parameters are specified, they should all apply to the retur
...
 
+### Events
+
+The events endpoint is a near-ish real time feed of events, designed to give access to as recent as possible series of events from the vehicle feed. It functions similarly to status changes, but shall not included data older than 2 weeks (that should live in `/status_changes.`)
+
+Unless stated otherwise by the municipality, this endpoint must return only those events with a `event_location` that [intersects](#intersection-operation) with the [municipality boundary](#municipality-boundary).
+
+> Note: As a result of this definition, consumers should query the [trips endpoint](#trips) to infer when vehicles enter or leave the municipality boundary.
+
+The datatype and scheme are the same as `status_changes
```
ascherkus: Looks like you're missing a closing ` here

(also not sure whether it's "scheme" or "schema" here -- I defer to you!)

hunterowens:
@thekaveman @ascherkus 

I went with requiring `end_time`, otherwise error, since if we are making breaking changes, we might as well make it the ones that are the clearest to implement. 

@ascherkus, I added language to describe what should be in the response payload that should clear up those comments / fixed the various typos you caught. LMK if that clears things up. 

@thekaveman: I think I'd rather keep the time definition explicit, as introducing a snap to hour via any timestamps means there is the possibility of something going wrong on the server side as we make an implicit conversion, I think this keeps it cleaner for implementation and debugging between teams.

provider/README.md
```diff
@@ -199,14 +201,17 @@ Schema: [`trips` schema][trips-schema]
...
-* `min_end_time`: filters for trips where `end_time` occurs at or after the given time
-* `max_end_time`: filters for trips where `end_time` occurs before the given time
+Without an `end_time` query parameter, `/trips` shall return an error. 
 
-When multiple query parameters are specified, they should all apply to the returned trips. For example, a request with `?min_end_time=1549800000000&max_end_time=1549886400000` should only return trips whose end time falls in the range `[1549800000000, 1549886400000)`.
+* `end_time`: An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date for a static bucket of files representing an Hour. The response will have all trips whose end time occurred in the hour. For example, if you requested "2019-10-01:07", you should get back all trips that ended between `7:00 <= and < 8:00`. Timezone UTC. 
+
+| Request range | format | expected output | 
+| ------------- | ------ | --------------- | 
+| Hour          | YYYY-MM-DD:HH | That hour worth of trips, or if the hour is not complete or the file does not exist, an error. The minimum value of the hour should be 00 and the max 23, as the midnight hour should be denoted as the next day's zero hour. |
```
hyperknot: That's a `T`, not a `:` in ISO 8601 definition. https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations

hyperknot:
Epoch seconds / milliseconds is a very highly bug prone part of the current specs. Each provider have their own implementation about which one to use, which is especially confusing as they need to support 0.2 and 0.3 clients concurrently. Some providers do an auto-detection and accept both formats (it can be done by saying comparing to year 2100's timestamp in seconds).

Also, unix seconds / milliseconds are not fit for describing a time "block", like an UTC hour, as it can be both placed in the beginning of the hour, somewhere in the middle or at the very end, all raising possible bug cases.

13 byte ISO 8601 strings ("2019-10-21T19") solve these issues and define an hourly block explicitly.

provider/README.md
hunterowens: @hyperknot- Thanks for catching that! fixed in another commit.

hyperknot:
So to conclude, this PR is aiming to replace the current DB based trips and status changes endpoints with the static ones, instead of adding 2 more new endpoints? If so, I agree with the design, it's a clean choice!

+1 endpoint for real-time-ish events (/events), whose future we can discuss in other PRs?

hunterowens:
@hyperknot Yes, the idea is that `/trips`, `/status_changes` become backed by static files, and `/events` is what `status_changes` is today.

hyperknot:
> @hyperknot Yes, the idea is that `/trips`, `/status_changes` become backed by static files, and `/events` is what `status_changes` is today.

I understand, that's a great change.

ascherkus:
Clarity on hour interval text = LGTM! Thanks! Still find use of `end_time` for `status_changes` to be slightly confusing but can live with it given the clarity improvements.

Generally speaking I agree with @hyperknot ... removing pagination would make implementation more consistent across providers and the response sizes seem reasonable enough to warrant removal. On the flip side, I recall @bhandzo (maybe?) saying they'd still prefer to have pagination within the hour intervals so would be good to get some input as to what the use cases are there.

thekaveman:
Rebased this PR on the latest `dev` changes, and made a few language cleanups around query parameters.

Issue #260, opened 2019-03-04
Agency: allow accuracy as well


ezheidtmann:
Some software configurations don't provide an HDOP figure, but do provide an accuracy measurement in meters. For example, both iOS and Android platforms do this. At least one operator has Android running on the vehicle, and at least one operator has used the user's smartphone to collect some GPS data.

Rather than asking operators to back out HDOP in these situations, the spec should support receiving a "radius of accuracy in meters."

karcass:
Seems legit.  I'll add that.  PR forthcoming.

hunterowens:
fixed in #360

PR #360, opened 2019-09-19
Add accuracy field to GPS telemetry

karcass:
### Explain pull request

Per #260 

> Some software configurations don't provide an HDOP figure, but do provide an accuracy measurement in meters. For example, both iOS and Android platforms do this. At least one operator has Android running on the vehicle, and at least one operator has used the user's smartphone to collect some GPS data.
> 
> Rather than asking operators to back out HDOP in these situations, the spec should support receiving a "radius of accuracy in meters."
> 

### Is this a breaking change

 [ ] Yes, breaking
 [x] No, not breaking
 [ ] I'm not sure

### `Provider` or `agency`

 [ ] `provider`
 [x] `agency`
 [ ] both


hunterowens:
this needs associated JSON schema update?

hunterowens:
^ bump

Issue #348, opened 2019-07-30
Field size limitation


patata3000:
Is there a plan to add a field size limitation. For example, the device_id or vehicle_id could be limited in their number of characters.

karcass:
`device_id` is a UUID, so it has fixed length.

`vehicle_id` should probably have a maximum length.  The reference implementation uses a `varchar(32)`, and so far that hasn't been a problem in practice.

karcass:
It probably wouldn't harm anything to bump that to, say, 256 characters, and call it done.

hunterowens:
yeah, I mean, theoretically vehicle_id needs to be printed on the vehicle. I suppose we should have this all done inside the JSONschema.

karcass:
We bumped the implementation of Agency varchar items to 255.  I'll add a PR for that.

hunterowens:
fixed! #361

PR #361, opened 2019-09-19
Add string limit length of 255

karcass:
### Explain pull request

Per #348:

> Is there a plan to add a field size limitation? For example, the device_id or vehicle_id could be limited in their number of characters.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `agency`

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -28,6 +28,10 @@ When making requests, the Agency API expects `provider_id` to be part of the cla
 
 As with the Provider API, `timestamp` refers to integer milliseconds since Unix epoch. 
 
+## Strings
+
+All String fields, such as `vehicle_id`, are limited to a maximum of 255 characters.
+
 ## Vehicles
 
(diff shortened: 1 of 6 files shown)


sweis:
You could enforce this in the schema by using the "maxLength" validator. You'd update get_vehicle.json like this:
```
...
  "vehicle_id": {
      "$id": "#/properties/vehicle_id",
      "type": "string",
      "description": "The Vehicle Identification Number visible on the vehicle itself",
      "maxLength": 255,
      "default": "",
      "examples": [
        "ABC123"
      ],
      "pattern": "^(.*)$"
    },
...
```

See here for examples: 
https://json-schema.org/understanding-json-schema/reference/string.html#length

karcass:
Good call @sweis.  Adding.

hunterowens:
this will be overwritten by the schema generator in all /schema directory. @karcass  - when you generate the schemas, you need to either the code / template schemas in that directory.

hunterowens:
Hey @karcass have you had a chance to update this using the appropriate `schema/generate_schema.py` process?

PR #363, opened 2019-09-19
Update Agency state machine artwork

karcass:
### Explain pull request

The prior agency status diagram left out some available information, such as which states were considered Public Right of Way.  It also had a bunch of colors that did not convey anything useful.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `agency` but only in terms of visual presentation of the spec

### Additional context

None


hunterowens:
any reason why this file is 4x the size of the old one?

karcass:
> any reason why this file is 4x the size of the old one?

Some combination of: different artist, different tool, higher resolution.

hunterowens:
if possible, I'd like to have this autogenerated using a tool that we can run as a program, so that inherits the inside the JSON schema and we don't have to remember to update this. but merging for nwo.

PR #367, opened 2019-09-26
change CLEVR provider

valclevrmobility:
### Explain pull request

change CLEVR provider name to GRUV and api links

### Is this a breaking change

 * I'm not sure

### `Provider` or `agency`

 * `provider`


hunterowens:
thanks @valclevrmobility. 

Typically we remove the region specific information if possible from the URLs? Can you update the PR (unless the URLs are the only region in existence, which means I can merge this?)

valclevrmobility:
Hunter,

thanks for feedback! We'll update.

// Val
CLEVR Mobility


Hunter Owens <notifications@github.com<mailto:notifications@github.com>> wrote:

thanks @valclevrmobility<https://github.com/valclevrmobility>.

Typically we remove the region specific information if possible from the URLs? Can you update the PR (unless the URLs are the only region in existence, which means I can merge this?)

Issue #317, opened 2019-05-21
Error responses for invalid param for Provider API, same as Agency


dirkdk:
We at Spin are supporting both MDS 0.2.0 and MDS 0.3.0 as provider API. We would like to help our users (= cities) to use the right requests with the right parameters. E.G. for `/trips`, `start_time` is a timestamp and this needs to be in milliseconds for 0.3.0. This was in seconds for 0.2.0. For this we would like to return errors with a clear description. Provider API currently has no specification how to return errors. `bbox` is not supported anymore for 0.3.0

### Adopt same error responses as for Agency API
The Agency part of the API already has error responses define: https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/agency/README.md#responses

We propose to adopt the same structure for Provider.

### Is this a breaking change

* [ ] Yes, breaking
* [x] No, not breaking
* [ ] I'm not sure

### `Provider` or `agency`

* [x] `provider`
* [ ] `agency`
* [ ] both

### Describe alternatives you've considered

Alternative to return our own homegrown responses, that is not good

hunterowens:
I think standardizing (or at least standardizing with some variation) the possible error codes should make implementing provider clients significantly easier, thanks for the suggestion @dirkdk

karcass:
I support this 😄

karcass:
In fact I will make a PR proposal for the Provider spec and see if it gets buy-in for 0.4.0.

hunterowens:
fixed in #368

PR #368, opened 2019-09-27
Unify error response format per #317

karcass:
### Explain pull request

From #317 

> Adopt same error responses as for Agency API

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `provider`

### Additional context

Nothing immediately comes to mind

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -75,6 +75,31 @@ Response bodies must be a `UTF-8` encoded JSON object and must minimally include
 }
 ```
 
+#### Provider HTTP Codes
+
+* **200:** OK: operation successful.
+* **400:** Bad request.
+* **401:** Unauthorized: Invalid, expired, or insufficient scope of token.
+* **404:** Not Found: Object(s) do not exist.
+* **500:** Internal server error.
+
+#### Error Format
+
+```json
+{
+    "version": "x.y.z",
+    "error": "...",
+    "error_description": "...",
+    "error_details": [ "...", "..." ]
+}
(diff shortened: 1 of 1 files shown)


provider/README.md
```diff
@@ -75,6 +75,31 @@ Response bodies must be a `UTF-8` encoded JSON object and must minimally include
...
+* **400:** Bad request.
+* **401:** Unauthorized: Invalid, expired, or insufficient scope of token.
+* **404:** Not Found: Object(s) do not exist.
+* **500:** Internal server error.
+
+#### Error Format
+
+```json
+{
+    "version": "x.y.z",
```
thekaveman: I have a feeling that including `version` in the error response might be more confusing than it is worth. E.g. is that the version that was requested? Or the version that the Provider would have responded with had the request been successful? Or...?

provider/README.md
```diff
@@ -75,6 +75,31 @@ Response bodies must be a `UTF-8` encoded JSON object and must minimally include
 }
 ```
 
+#### Provider HTTP Codes
```
thekaveman: Probably don't need the word *Provider* here. Maybe *HTTP Response Codes*?

karcass: Sure.

provider/README.md
karcass: If I take it out, I'll have to modify agency too (which is fine).

karcass:
@thekaveman updated

Issue #355, opened 2019-09-06
Clarify municipality boundary geofiltering for queries


thekidder:
### Is your feature request related to a problem? Please describe.

#226 clarified precisely what data providers are expected to return through the API, by requiring an unambiguous geography boundary:

> Municipalities requiring MDS Provider API compliance should provide an unambiguous digital source for the municipality boundary. This boundary must be used when determining which data each provider API endpoint will include.

> The boundary should be defined as a polygon or collection of polygons. The file defining the boundary should be provided in Shapefile or GeoJSON format and hosted online at a published address that all providers and provider API consumers can access and download.

The wording here could be interpreted to mean that providers need to apply the latest version of this municipality boundary to all queries. This interpretation prevents providers from performing optimizations like pre-processing and storing flat files for faster access (see #354).

### Describe the solution you'd like

I propose clarifying exactly when this municipality boundary is required to be applied to the MDS data. Specifically, add language stating that this filtering can be done at data ingest time and is not required at query time. Agencies may still publish changes to this municipality boundary, but there is no expectation that these changes be applied to past data. This allows providers to use more efficient storage implementations without requiring potentially expensive reprocessing of old data.

### Is this a breaking change

 * No, not breaking

As this is mostly a clarification of what data is required to be returned in the API, I consider it non-breaking.

### `Provider` or `agency`

For which API is this feature being requested:

 * `provider`

PR #369, opened 2019-09-30
Clarify municipality boundary geofiltering for historical data (#355)

thekidder:
### Explain pull request

#226 clarified precisely what data providers are expected to return through the API, by requiring an unambiguous geography boundary:

> Municipalities requiring MDS Provider API compliance should provide an unambiguous digital source for the municipality boundary. This boundary must be used when determining which data each provider API endpoint will include.

> The boundary should be defined as a polygon or collection of polygons. The file defining the boundary should be provided in Shapefile or GeoJSON format and hosted online at a published address that all providers and provider API consumers can access and download.

The wording here could be interpreted to mean that providers need to apply the latest version of this municipality boundary to all queries. This interpretation prevents providers from performing optimizations like pre-processing and storing flat files for faster access (see #354).

This adds a couple of sentences to clarify the intent of the municipality boundary filtering. Specifically, it explicitly does not require historical reprocessing of data when the boundary changes.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `provider`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -157,6 +157,8 @@ Municipalities requiring MDS Provider API compliance should provide an unambiguo
 
 The boundary should be defined as a polygon or collection of polygons. The file defining the boundary should be provided in Shapefile or GeoJSON format and hosted online at a published address that all providers and `provider` API consumers can access and download.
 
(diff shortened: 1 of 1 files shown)


hunterowens:
thanks @thekidder! fixes #355

PR #371, opened 2019-10-02
Remove SLA verbiage from /telemetry endpoint

karcass:
The paragraph specifying latency and time/space resolution should never have been included in the spec.  Such requirements are policy decisions that should be adjudicated by regulating agencies.

### Explain pull request

Remove paragraph

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `agency`

### Additional context

None


hunterowens:
LGTM.

PR #372, opened 2019-10-02
Update ReleaseGuidelines.md with meeting info

sarob:
### Explain pull request

As per issue https://github.com/CityOfLosAngeles/mobility-data-specification/issues/364

I will need to add the meeting calendar URL to the PR before this can be merged.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * None

### Additional context

None


hunterowens:
@sarob I think we did invite via the mailing list rather than separate calendar last cycle. I prep another update.

Issue #271, opened 2019-03-26
Align provider and agency APIs


lionel-panhaleux:
Hello, 

The Provider and Agency APIs list different `event_type` and `event_type_reason` values.
This is cumbersome (and error-prone) on the provider side and the agency side alike. We could maintain a single list of `event_type` and share it between the two APIs ? I could submit a PR if you are interested.

AFAICT the agency side has a more elaborate list of events, so it makes sense to use this list as a first version. However this would be a breaking change for the Provider API. Another approach would be to begin with a non breaking list including all events from the Provider and Agency APIs, then indicate some values to be deprecated in the next major version.

What do you think ?

karcass:
@lionel-panhaleux Agreed that they should be as similar as possible.  Agency has more requirements than Provider, which is the source of the differences.   While we ("we" meaning all of us with a stake in MDS) learn how the expanded list of status-changes plays out in the real-world, E&A decided it would be (in some ways) more expedient to be a superset of Provider rather than forcing Provider to be exactly the same.

karcass:
I would say that we should keep this issue warm until, say, mid-May, and revisit.  That will give us a solid month of production experience with the permitted providers.

rf-:
It might make sense to treat this as two separate requests:
1. Harmonize the names and structure of the events where they currently overlap
2. Expand the list of status changes in Provider to cover some or all of the extra events from Agency

It makes sense to take a wait-and-see approach with the second one, but it doesn't seem like there's much reason for something like a user dropoff event or a maintenance pickup to be represented differently in Agency vs. Provider.

karcass:
@rf- smart

yoimnik:
hey so what's the traction on this? we should get this added in provider asap

specifically, the problem with the nomenclature of event type "reserved" for provider being different from the event type "reserved"  in agency

also, agency's "service_end" puts the bird in an unavailable state, which is in the PROW, while provider's "service_end" is an event_type_reason for making it in a removed state, which is off the PROW

we need to align these. and we need to get provider `unavailable` `off_hours` events and  `service_start` `on_hours` events (or whatever more specific naming)

what do you guys think?

karcass:
There's definitely traction.  It's on me and @hunterowens (mostly me) to make the canonical list of discrepancies and start knocking them down.  Once we have the canonical list we can go one-by-one and adjudicate which approach to take (e.g. change provider, change agency, change both, agree to leave different until a later rev).

[...]

Issue #67, opened 2018-09-13
Lost & Found


bhandzo:
The current status changes don't account for vehicles which go missing (for any number of reasons). These vehicles are not available to be ridden, and most likely are also not in the public right of way. There are two ways this could go:

1. Add a new `event_type` of `lost` and exclude those vehicles from cap counts. 
2. Add a new `event_type_reason` of `lost` to the `removed` `event_type`

Either way, there should then be an `event_type_reason` for `available` of `found`

hunterowens:
Aside from adding lost, we should add stolen. 

@bhandzo can you prepare a PR for adding those event types and event type reasons. @LADOTBikeshare and I will work on how it applies to the cap.

mattwigway:
Damaged might be a good one as well (flat tires, etc.)

thekaveman:
So is the idea here that the provider would mark, within their system, a device as `lost` etc - and this would generate an MDS Provider `status_change` with the associated data?

I have some follow-up thoughts and questions:

  - how does a provider know that a vehicle is lost? Do they know enough about the vehicle to know the `event_time` and `event_location` where it was lost (as required by a `status_change` event)?

  - what is the definition of *lost*, and does every provider now and into the future agree on this definition? If the provider's system loses track of the device (and marks it as lost with a `status_change` as outlined above), but I as a customer find it on the street (or in a bush, or in the creek, or whatever...), can I begin a trip with that device? Or has the provider's system locked it because it was marked as lost, and now no one can use it until it is "found" and unlocked? Does every provider handle this situation the same?

  - Assuming we wanted to add some kind of event for lost devices and we answered the above questions, I disagree about `event_type:lost`. I think this should use the existing `event_type:unavailable` or `event_type:removed` (depending on the answer to the scenario I posed above) with an `event_type_reason:lost`

  - I'm not sure that all of this matters for a regulatory agency. If a provider considers their device to be "lost" - by definition that means the provider doesn't know where it is. They don't know if it's blocking the sidewalk, an ADA ramp, in the ocean, etc. From a *total devices allocated* point of view, the device is still out there, because it hasn't been explicitly brought back in.

thekaveman:
Closing this as the conversation seems to have died off. Feel free to re-open with any additional comments / follow-ups on the thoughts above.

billdirks:
[...]

PR #376, opened 2019-10-11
Adding a /vehicles endpoint to show the current status of the fleet

### LINKED ISSUE #310: Add an available vehicles endpoint to provider

Agencies often use our MDS status change feed to figure out which vehicles are available in their region. Many are trying to calculate parking caps, or trying to get simple counts of vehicle availability. Using status changes is problematic. Most agencies try to do this by replaying all of the status changes in their feed, in an attempt to replay all the state changes in the vehicle state machine. Most just want to answer the simple question: what's the current status of the provider's fleet?

A vehicles endpoint in MDS provider, similar to the one in MDS agency, wherein a provider publishes a list of the vehicles that are currently registered in the region. This endpoint could take as inspiration the same functionality that's [provided in GBFS](https://github.com/NABSA/gbfs/blob/master/gbfs.md#free_bike_statusjson). It could contain the following information:

* Device ID/Vehicle ID
* Most recent Lat/Long
* Current status/availability
* Most recent event ID

### Is this a breaking change

* [ ] Yes, breaking
* [X] No, not breaking
* [ ] I'm not sure

### `Provider` or `agency`

* [x] `provider`
* [ ] `agency`
* [ ] both

The most obvious alternative is to add additional status change types, but this would still require agencies to do their own analysis to replay each vehicle's state machine. I believe this work should be done regardless. But since many agencies are trying to understand vehicle availability, I believe it would be easiest for all parties if the provider made this available through a simple API where little-to-no analytics or data processing would be necessary.

bhandzo:
### Explain pull request

There have been several long running issues & conversations about how to provide real time vs. historical data on vehicles in the PROW. @johnpena specifically requested something like this in #310. 

This goes well with the move to static pagination for historical status changes in #357 , and allows consumers to get the full state of the fleet *right now* rather than having to reconstruct it from a stream of status changes. It's likely this would obviate the need to include GBFS as part of MDS.

This makes MDS more accessible by allowing for a simpler fleet counting method (count the number of vehicles returned in this response). These counts should line up with counts generated from status changes, which would give agencies the ability to fact check real time responses to feel confident they fully reflect what's on the PROW. 

### Is this a breaking change

 * No, not breaking (This adds a new endpoint that consumers could choose to.. consume)
 

### `Provider` or `agency`

 * `provider`

### Additional context

@johnpena @hunterowens @thekaveman @ascherkus @babldev @hyperknot @rf- @fscottfoti Would love thoughts.

Closes #310.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -434,7 +434,7 @@ ttl                 | Yes       | Integer representing the number of millisecond
 | `last_event_type` | Enum | Required | Event type of most recent status change. See [event types](#event-types) table |
 | `last_event_type_reason` | Enum | Required | Event type reason of most recent status change, allowable values determined by [`event type`](#event-types) |
 | `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of vehicle's last event |
-| `current_location` | GeoJSON [Point Feature][geo] | Required | Current location of vehicle if different from last event, and the vehicle is not currently on a trip |
+| `current_location` | GeoJSON [Point Feature][geo] | Required if Applicable | Current location of vehicle if different from last event, and the vehicle is not currently on a trip |
 | `battery_pct` | Float | Required if Applicable | Percent battery charge of device, expressed between 0 and 1 |
 
 [Top][toc]
--- a/provider/README.md
+++ b/provider/README.md
@@ -420,6 +420,7 @@ ttl                 | Yes       | Integer representing the number of millisecond
 **Endpoint:** `/vehicles`  
 **Method:** `GET`  
 **Required/Optional:** Optional starting with `0.4.1`, moving to Required in a future version (`0.5.0`+)  
+**Schema:** [`vehicles` schema][vehicles-schema]  
 **`data` Payload:** `{ "vehicles": [] }`, an array of objects with the following structure
 
 | Field | Type | Required/Optional | Comments |
@@ -441,8 +442,9 @@ ttl                 | Yes       | Integer representing the number of millisecond
 
 [general-information/versioning]: /general-information.md#versioning
 [geo]: #geographic-data
-[sc-schema]: status_changes.json
+[sc-schema]: dockless/status_changes.json
 [status]: #status-changes
 [toc]: #table-of-contents
-[trips-schema]: trips.json
+[trips-schema]: dockless/trips.json
 [ts]: #timestamps
+[vehicles-schema]: dockless/vehicles.json
--- a/provider/README.md
+++ b/provider/README.md
@@ -392,13 +392,22 @@ Should either side of the requested time range be missing, `/events` shall retur
 
 Should either side of the requested time range be greater than 2 weeks before the time of the request, `/events` shall return a `400 Bad Request` error.
 
-All providers should provide an endpoint that displays the current status of the fleet. Data in this endpoint should reconcile with data from the status changes enpdoint.
-
 ### Vehicles
+
+All providers should provide an endpoint that displays the current status of the fleet. Data in this endpoint should reconcile with data from the status changes enpdoint.
 
 Endpoint: `/vehicles`  
 Method: `GET`   
-`data` Payload: `{ "vehicles": [] }`, an array of objects with the following structure
+
+The response should contain header information that indicates when it was last updated and when it will be updated again
+
+Field Name          | Required  | Defines
+--------------------| ----------| ----------
+last_updated        | Yes       | Timestamp indicating the last time the data in this feed was updated
+ttl                 | Yes       | Integer representing the number of seconds before the data in this feed will be updated again (0 if the data should always be refreshed)
+
+
+`data` Payload: `{ "last_updated": , "ttl": ,"vehicles": [] }`, an array of objects with the following structure
 
 | Field | Type | Required/Optional | Comments |
 | ----- | ---- | ----------------- | ----- |
--- a/provider/README.md
+++ b/provider/README.md
@@ -396,17 +396,29 @@ Should either side of the requested time range be greater than 2 weeks before th
 
 All providers should provide an endpoint that displays the current status of the fleet. Data in this endpoint should reconcile with data from the status changes enpdoint.
 
-Endpoint: `/vehicles`  
-Method: `GET`   
+In addition to the standard [Provider payload wrapper](#response-format), responses from this endpoint should contain the last update timestamp and amount of time until the next update:
 
-The response should contain header information that indicates when it was last updated and when it will be updated again
+```json
+{
+    "version": "x.y.z",
+    "data": {
+        "vehicles": []
+    },
+    "last_updated": "12345",
+    "ttl": "12345"
+}
+```
+
+Where `last_updated` and `ttl` are defined as follows:
 
 Field Name          | Required  | Defines
 --------------------| ----------| ----------
 last_updated        | Yes       | Timestamp indicating the last time the data in this feed was updated
-ttl                 | Yes       | Integer representing the number of seconds before the data in this feed will be updated again (0 if the data should always be refreshed)
+ttl                 | Yes       | Integer representing the number of milliseconds before the data in this feed will be updated again (0 if the data should always be refreshed)
 
-`data` Payload: `{ "last_updated": , "ttl": ,"vehicles": [] }`, an array of objects with the following structure
+Endpoint: `/vehicles`
+Method: `GET`
+`data` Payload: `{ "vehicles": [] }`, an array of objects with the following structure
 
 | Field | Type | Required/Optional | Comments |
 | ----- | ---- | ----------------- | ----- |
--- a/provider/README.md
+++ b/provider/README.md
@@ -392,6 +392,25 @@ Should either side of the requested time range be missing, `/events` shall retur
 
 Should either side of the requested time range be greater than 2 weeks before the time of the request, `/events` shall return a `400 Bad Request` error.
 
+All providers should provide an endpoint that displays the current status of the fleet. Data in this endpoint should reconcile with data from the status changes enpdoint.
+
+Endpoint: `/vehicles`  
+Method: `GET`   
+`data` Payload: `{ "vehicles": [] }`, an array of objects with the following structure
+
+| Field | Type | Required/Optional | Comments |
+| ----- | ---- | ----------------- | ----- |
+| `device_id` | UUID | Required | A unique device ID in UUID format |
+| `vehicle_id` | String | Required | The Vehicle Identification Number visible on the vehicle itself |
+| `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
+| `propulsion_type` | Enum[] | Required | Array of [propulsion types](#propulsion-types); allows multiple values |
+| `last_event_time` | [timestamp][ts] | Required | Date/time that last status change occurred at. See [Event Times](#event-times) |
+| `last_event_type` | Enum | Required | See [event types](#event-types) table |
+| `last_event_type_reason` | Enum | Required | Reason for status change, allowable values determined by [`event type`](#event-types) |
+| `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of veicle's last event |
+| `current_location` | GeoJSON [Point Feature][geo] | Optional | Current loaction of vehicle if different from last event |
+| `battery_pct` | Float | Required if Applicable | Percent battery charge of device, expressed between 0 and 1 |
+
 [Top][toc]
 
 [general-information/versioning]: /general-information.md#versioning
--- a/provider/README.md
+++ b/provider/README.md
@@ -416,9 +416,10 @@ Field Name          | Required  | Defines
 last_updated        | Yes       | Timestamp indicating the last time the data in this feed was updated
 ttl                 | Yes       | Integer representing the number of milliseconds before the data in this feed will be updated again (0 if the data should always be refreshed).
 
-Endpoint: `/vehicles`
-Method: `GET`
-`data` Payload: `{ "vehicles": [] }`, an array of objects with the following structure
+**Endpoint:** `/vehicles`  
+**Method:** `GET`  
+**Required/Optional:** Optional starting with `0.4.1`, moving to Required in a future version (`0.5.0`+)  
+**`data` Payload:** `{ "vehicles": [] }`, an array of objects with the following structure
 
 | Field | Type | Required/Optional | Comments |
 | ----- | ---- | ----------------- | ----- |
--- a/provider/README.md
+++ b/provider/README.md
@@ -431,7 +431,7 @@ Method: `GET`
 | `last_event_time` | [timestamp][ts] | Required | Date/time when last status change occurred. See [Event Times](#event-times) |
 | `last_event_type` | Enum | Required | Event type of most recent status change. See [event types](#event-types) table |
 | `last_event_type_reason` | Enum | Required | Event type reason of most recent status change, allowable values determined by [`event type`](#event-types) |
-| `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of veicle's last event |
+| `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of vehicle's last event |
 | `current_location` | GeoJSON [Point Feature][geo] | Required | Current location of vehicle if different from last event, and the vehicle is not currently on a trip |
 | `battery_pct` | Float | Required if Applicable | Percent battery charge of device, expressed between 0 and 1 |
 
--- a/provider/README.md
+++ b/provider/README.md
@@ -402,13 +402,13 @@ Method: `GET`
 
 | Field | Type | Required/Optional | Comments |
 | ----- | ---- | ----------------- | ----- |
-| `device_id` | UUID | Required | A unique device ID in UUID format |
-| `vehicle_id` | String | Required | The Vehicle Identification Number visible on the vehicle itself |
+| `device_id` | UUID | Required | A unique device ID in UUID format, should match this device in Provider |
+| `vehicle_id` | String | Required | The Vehicle Identification Number visible on the vehicle itself, should match this device in provider |
 | `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
 | `propulsion_type` | Enum[] | Required | Array of [propulsion types](#propulsion-types); allows multiple values |
-| `last_event_time` | [timestamp][ts] | Required | Date/time that last status change occurred at. See [Event Times](#event-times) |
-| `last_event_type` | Enum | Required | See [event types](#event-types) table |
-| `last_event_type_reason` | Enum | Required | Reason for status change, allowable values determined by [`event type`](#event-types) |
+| `last_event_time` | [timestamp][ts] | Required | Date/time when last status change occurred. See [Event Times](#event-times) |
+| `last_event_type` | Enum | Required | Event type of most recent status change. See [event types](#event-types) table |
+| `last_event_type_reason` | Enum | Required | Event type reason of most recent status change, allowable values determined by [`event type`](#event-types) |
 | `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of veicle's last event |
 | `current_location` | GeoJSON [Point Feature][geo] | Optional | Current loaction of vehicle if different from last event |
 | `battery_pct` | Float | Required if Applicable | Percent battery charge of device, expressed between 0 and 1 |
--- a/provider/README.md
+++ b/provider/README.md
@@ -394,6 +394,8 @@ Should either side of the requested time range be greater than 2 weeks before th
 
 All providers should provide an endpoint that displays the current status of the fleet. Data in this endpoint should reconcile with data from the status changes enpdoint.
 
+### Vehicles
+
 Endpoint: `/vehicles`  
 Method: `GET`   
 `data` Payload: `{ "vehicles": [] }`, an array of objects with the following structure
--- a/provider/README.md
+++ b/provider/README.md
@@ -394,7 +394,7 @@ Should either side of the requested time range be greater than 2 weeks before th
 
 ### Vehicles
 
-All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should reconcile with data from the status changes enpdoint. This data should be refreshed at most every 5 minutes.
+The `/vehicles` endpoint returns the current status of vehicles on the PROW. Only vehicles that are currently in available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should reconcile with data from the `/status_changes` enpdoint. The data returned by this endpoint should be as close to realtime as possible, but in no case should it be more than 5 minutes out-of-date.
 
 In addition to the standard [Provider payload wrapper](#response-format), responses from this endpoint should contain the last update timestamp and amount of time until the next update:
 
--- a/provider/README.md
+++ b/provider/README.md
@@ -394,7 +394,7 @@ Should either side of the requested time range be greater than 2 weeks before th
 
 ### Vehicles
 
-All providers should provide an endpoint that displays the current status of the fleet. Data in this endpoint should reconcile with data from the status changes enpdoint.
+All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should reconcile with data from the status changes enpdoint. This data should be refreshed at most every 5 minutes.
 
 In addition to the standard [Provider payload wrapper](#response-format), responses from this endpoint should contain the last update timestamp and amount of time until the next update:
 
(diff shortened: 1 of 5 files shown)


karcass:
Did you intentionally drop `provider_id`?  

I think this endpoint might also want pagination.  A single provider might have upward of 10,000 vehicles in large cities.

bhandzo:
@karcass I think it would make most sense for this to work similarly to GBFS with each provider having a unique URL rather than using provider ID, but i'm v open to feedback.

Pagination makes a ton of sense to me, I think in this case it would be great to define how it should work for the sake of clarity.

johnpena:
Agreed that this endpoint should include pagination. Otherwise LGTM, I think this is really valuable and I'd love to see this included in provider.

thekaveman (changes requested):
> Did you intentionally drop `provider_id?`

>> I think it would make most sense for this to work similarly to GBFS with each provider having a unique URL rather than using provider ID

@bhandzo I think @karcass was referring to the data coming back from the endpoint (vs. a query parameter/URL token); each vehicle should be identified with the `provider_id` that vehicle belongs to, and similarly for the `provider_name`. 

These fields are necessary for consumers that may hit `/vehicles` endpoints from multiple providers, and align with the vehicle information currently coming out of `/status_changes` and `/trips`.

bhandzo:
@karcass @thekaveman @johnpena Added `provider_id` and `provider_name`, as well as pagination (copied down briefly from further up).

thekaveman:
@bhandzo thanks for adding those fields.

If this endpoint was specified as having the same payload wrapper as the others, we could avoid duplicating the pagination docs from earlier (and maintain consistency with other parts of the spec). I might re-word like:

> ### Vehicles
>
> In addition to the standard [Provider payload wrapper](#response-format), responses from this endpoint should contain the last update timestamp and amount of time until the next update:
> ```json
> {
>     "version": "x.y.z",
>     "data": {
>         "vehicles": []
>     },
>     "last_updated": "12345",
>     "ttl": "12345"
> }
> ```
> Where `last_updated` and `ttl` are defined as follows:
>
> Field Name          | Required  | Defines
> --------------------| ----------| ----------
> last_updated        | Yes       | [Timestamp](#ts) indicating the last time the data in this feed was updated
> ttl                 | Yes       | Integer representing the number of milliseconds before the data in this feed will be updated again (0 if the data should always be refreshed)
>
> Endpoint: `/vehicles`  
> Method: `GET`   
> `data` Payload: `{ "vehicles": [] }`, an array of objects with the following structure
> ...

I also suggest changing the definition of `ttl` to be the integer number of *milliseconds*, for consistency with other representations of time in this spec.

Finally, we'll want a JSON schema for this endpoint as well - and assuming the above suggestions are agreeable, the common schema will need some edits to handle the additional payload fields.

provider/README.md
```diff
@@ -336,8 +336,52 @@ When multiple query parameters are specified, they should all apply to the retur
 [Top][toc]
 
 ## Realtime Data
+All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in an available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should generally reconcile with data from the status changes enpdoint.
```
rf-: I feel like there maybe needs to be more definition about how this relates to the status changes endpoint. Specifically, I wonder about how this endpoint deals with lost vehicles:
- If the provider is allowed to omit a vehicle from this endpoint when they no longer believe it's in a knowable location on the PROW, it won't be possible in general to reconcile this data with data from the status changes endpoint, since that endpoint has no way of representing a vehicle going missing.
- If it's mandatory for the data from this endpoint to correlate with the status changes endpoint, the provider will have no way to prune missing vehicles, resulting in a monotonically increasing number of "stuck" vehicles in this endpoint over time.

The former option definitely seems more practical, but it also creates a weird situation where the only way for a consumer to know when a provider considers a vehicle "lost" is to watch the realtime endpoint and see when it disappears. If removing vehicles from the realtime endpoint is permissible, it seems like it might make sense to also add a status change capturing that removal so that the two streams can be reliably correlated with each other.

bhandzo: This is the main issue to be resolved here. There have been long running conversations about how to deal with vehicles that go missing, or leave the PROW without an event. 

I've always advocated for the addition of a `lost` event type reason for `removed` which would make the latter option you suggest feasible. We've gone back and forth on the justification for such a reason, and whether or not this would obfuscate valuable data from consumers.

ascherkus (commented):
Minor typo nits... but overall looks great. Agree with @rf- and @bhandzo re: desire to have more explicit way to reconcile this feed with status changes (e.g., `lost` event?), otherwise reconciling relies on some sort implicit / underspecified logic.

I feel it'd be good to avoid a situation where each consumer/municipality uses different implicit logic for status change reconciliation if we can manage it (i.e., a bit of explicitness now would avoid increasing hidden system complexity down the road).

hyperknot:
Thanks for this great PR. This looks like the perfect direction forward! I have a feeling that this + the new static endpoints could even replicate the current provider MDS + GBFS endpoints one day.

I'd like to raise two important points:
1. 
```
| `current_location` | GeoJSON [Point Feature][geo] | Optional | Current loaction of vehicle if different from last event |
```
This should be required. The reason is that if we only supply anything "if different from last event" then this endpoint cannot be used for a "current state" view, as it needs to have all previous "diff" snapshots to reconstruct the current state.

2. What vehicles to include. This is the most important question.
- Currently it states: `Only vehicles that are currently in an available, unavailable, or reserved states should be returned in this payload.`

There are 3 options here:
1. Include only available vehicles, like GBFS. This is not adequate for many known reasons.
2. Include available "available, unavailable, or reserved" like proposed here. The problem with this is that this won't include out-of-service vehicles, what is probably the most important when a city is trying to figure out what vehicle is left over on a sidewalk in a broken state.

=> I'd propose a simplified approach, which solves all these problems:
3. Include all vehicles seen in the last n days (say 3 or 7), showing their last known state.

I believe this approach would solve all the mentioned problems, including lost vehicles. When a vehicle is registered lost, it should simply keep being visible as lost, so whenever a city finds an abandoned vehicle, they'd immediately know if it's in service, out of service, or possibly lost.

provider/README.md
billdirks: What does it mean for a reserved state to be returned in this payload? Does that mean the endpoint is returning the location where a reserved event occurred or is that the realtime location of the vehicle while it is in a trip? If the latter, I'm wondering if there needs to be some language about time resolution for two reasons: this is a moving vehicle and there may be some privacy considerations. That is, this would be realtime data of a person in a trip as opposed to a static vehicle on the street not associated with an individual.

provider/README.md
```diff
@@ -336,8 +336,51 @@ When multiple query parameters are specified, they should all apply to the retur
...
+| `provider_name` | String | Required | The public-facing name of the Provider |
+| `device_id` | UUID | Required | A unique device ID in UUID format, should match this device in Provider |
+| `vehicle_id` | String | Required | The Vehicle Identification Number visible on the vehicle itself, should match this device in provider |
+| `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
+| `propulsion_type` | Enum[] | Required | Array of [propulsion types](#propulsion-types); allows multiple values |
+| `last_event_time` | [timestamp][ts] | Required | Date/time when last status change occurred. See [Event Times](#event-times) |
+| `last_event_type` | Enum | Required | Event type of most recent status change. See [event types](#event-types) table |
+| `last_event_type_reason` | Enum | Required | Event type reason of most recent status change, allowable values determined by [`event type`](#event-types) |
+| `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of veicle's last event |
+| `current_location` | GeoJSON [Point Feature][geo] | Optional | Current loaction of vehicle if different from last event |
```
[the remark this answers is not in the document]
billdirks: I can see this field being used in a few different ways. It could indicate a vehicle is that is moved in a way that's not covered by a status event (eg a passerby moves a vehicle that is blocking the sidewalk). That seems useful. I could also see it being used to update the position of a vehicle that is in a reserved state. That would be very different than what we have today, since it would report realtime events of a vehicle when it is associated to an individual (vs a vehicle unoccupied on the PROW). Are there privacy implications that need to be addressed? An example, perhaps a edge case, of an abuse, would be if person1, with access to this feed, observes person2 in real life starting a trip, then person1 could then track in realtime person2.

billdirks:
@hyperknot Just wanted to make a comment about your point (2). I believe known broken vehicles on the PROW should have a status event "unavailable" with reason "maintenance".

provider/README.md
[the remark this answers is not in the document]
bhandzo: This is a super good point. I think the updating of current location for vehicles that have moved without emitting events seems quite useful, and i agree that updating location during a ride has some privacy implications. I'm going to tweak the language to take a crack at clarifying.

karcass:
> => I'd propose a simplified approach, which solves all these problems:
> 3. Include all vehicles seen in the last n days (say 3 or 7), showing their last known state.
> 
> I believe this approach would solve all the mentioned problems, including lost vehicles. When a vehicle is registered lost, it should simply keep being visible as lost, so whenever a city finds an abandoned vehicle, they'd immediately know if it's in service, out of service, or possibly lost.

+1 for this, and I'd like to suggest that in addition to `removed`, Provider adds `inactive` to indicate that a device is lost/stolen or permanently removed from the fleet. The `event_type_reason` would be `missing` for short-term, and `decommissioned` for permanent.

provider/README.md
```diff
@@ -336,8 +336,51 @@ When multiple query parameters are specified, they should all apply to the retur
 [Top][toc]
 
 ## Realtime Data
+All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in an available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should generally reconcile with data from the status changes enpdoint.
```
whereissean: Total vehicles in PRoW consists of those in states of available, unavailable, reserved, and trip.  Can vehicles in trip be added to this endpoint to ensure accurate counting for the purposes of things like vehicle caps?

provider/README.md
```diff
@@ -336,8 +336,51 @@ When multiple query parameters are specified, they should all apply to the retur
...
+
+| Field | Type | Required/Optional | Comments |
+| ----- | ---- | ----------------- | ----- |
+| `provider_id` | UUID | Required | A UUID for the Provider, unique within MDS |
+| `provider_name` | String | Required | The public-facing name of the Provider |
+| `device_id` | UUID | Required | A unique device ID in UUID format, should match this device in Provider |
+| `vehicle_id` | String | Required | The Vehicle Identification Number visible on the vehicle itself, should match this device in provider |
+| `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
+| `propulsion_type` | Enum[] | Required | Array of [propulsion types](#propulsion-types); allows multiple values |
+| `last_event_time` | [timestamp][ts] | Required | Date/time when last status change occurred. See [Event Times](#event-times) |
```
whereissean: Is `last_event_time` needed if `last_event_location` also contains a required timestamp?

provider/README.md
```diff
@@ -336,8 +336,51 @@ When multiple query parameters are specified, they should all apply to the retur
 [Top][toc]
 
 ## Realtime Data
+All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in an available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should generally reconcile with data from the status changes enpdoint.
 
-All MDS compatible `provider` APIs must expose a public [GBFS](https://github.com/NABSA/gbfs) feed as well. Given that GBFS hasn't fully [evolved to support dockless mobility](https://github.com/NABSA/gbfs/pull/92) yet, we follow the current guidelines in making bike information avaliable to the public. 
+### Vehicles
```
whereissean: Would it make sense to follow REST-like practice of returning a single vehicles if GET `/vehicles/{device_id}` is called?

whereissean (commented):
Good PR.  I'd like to continue to build towards consistency with `/vehicles` in agency specs (perhaps via changes to that spec).

provider/README.md
bhandzo: For provider reserved covers reserved & trip doesn't it?

hunterowens:
hey y'all: 

I think this is really great. given that I'm already super behind on getting `0.4.0` out the door, I think we are gonna fast follow 0.4.0 (within 6 weeks) with an 0.4.1 that absolutely will have this in in it.

thekaveman:
Another option:

Since #357 introduces the near-ish real-time `/events` endpoint with explicit wording that:

> [`/events`] functions similarly to status changes, but shall not include data older than 2 weeks

we could take the main ideas from this `/vehicles` proposal and fold into `/events` and #357:

* `/vehicles` renames the event fields to `last_event_*`, whereas the schema of `/events` is the same as the current `/status_changes` - keeping a consistent schema for the "real-time" endpoint could make implementation on the provider and consumer side much easier.

* from discussion here, it sounds like the concept of `current_location` would be useful; this field could be added to the `/events` endpoint as a required field with the same description as here ("...if different from last event, and the vehicle is not currently on a trip")

* the explicit `ttl` and `last_updated` fields in the payload wrapper of `/vehicles`, as opposed to query params for the time range in `/events` mirroring the current (0.3.x) `/status_changes`. We could drop these payload fields in favor of the current wording in #357 that I mentioned above ("...shall not include data older than 2 weeks") or be explicit and add the payload fields for `/events`. I haven't thought this all the way through yet... comments welcome.

It feels confusing for all to have both `/events` and `/vehicles` co-exist, can we solve the use-cases here with `/events` (potentially including some minor tweaks)?

hyperknot:
There are core differences between `/events` and `/vehicles` endpoints, even though at the end the reflect the near real-time-ish state.

/events is a diff-ing of state, similar to a git repository. To have the latest state, you'd need to start from a known reference state and "built up" all diff-s to get the current state. Downloading only the latest event means nothing in terms of reconstructing the state.

/vehicles is a "snapshot" of the state, similar to a "download ZIP" in GitHub for example. It contains everything needed to reconstruct the current state, but it's inefficient for downloading the diffs.

To have a live map based on MDS, we'd need to either redownload /vehicles snapshot every minute, or download vehicles once and then get the /events every minute.

So in conclusion I believe for a live map like functionality both of them makes sense. The big question is how well each endpoint could be supported by providers, and how much processing latency would they need to add for them.

jfh01:
Great work on this, everyone.

I wanted to offer a general +1 on this PR. I've heard from a few folks that cities without a lot of technical capacity want an easier way to track current vehicle status and location. This approach is more functional than a GBFS feed and doesn't require the full technical capability need to ingest a full MDS provider feed. 

Also, at their last developer workshop, the Mobility Data folks talked about pivoting GBFS towards serving only consumer-facing applications (rather than regulatory ones), and so this offers clear replacement functionality for cities that had been using GBFS as part of their regulatory approach.

Looking forward to seeing this get buttoned up, hopefully for our next release.

provider/README.md
karcass: I think `removed`/`lost` should be either `inactive`/`lost` or `missing`/`lost`.  My interpretation of `removed` is "not in the public right of way because the provider has it".  `lost` is pretty clearly "the provider doesn't have it", and it _may or may not_ be in the public right of way.

provider/README.md
jrheard: In today's MDS call, we talked about marking this /vehicles endpoint as "optional" in 0.4.1 and likely promoting it to "required" in 0.5.0. I'd be curious to see how this could best be expressed in this diff. Right now, this line says "All providers should" provide this endpoint - how should this wording be modified?

Would modifying that portion of the line to read something more like "Providers MAY provide an endpoint that..." be sufficient, or should this documentation also clearly call out the plan to mark this endpoint as required in 0.5.0? I don't have any particular preference re: wording, I just want to make sure that we write this down somewhere so that people who weren't on the call can get the full picture!

jfh01:
Documenting a few notes from the [discussion on 12/19](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2019.12.19). Corrections/edits welcome.

*On-trip vehicles:*
Agreed that they would be marked as on-trip in the end-point. Location would be the start location of the trip, not the current, actual location.

*Jurisdiction:*
When a vehicle is moving across jurisdictional boundaries, it was agreed that it would reflect in the feed of the jurisdiction in which the last status change event occurred. So a vehicle that starts in a jurisdiction would stay in that jurisdiction's feed until such time as a new status change occurred in a new jurisdiction.

*Missing/lost vehicles:*
Was agreed that this concept needs to be formalized in `/vehicles` and in other places in the spec.

*Querying:*
Request that `/vehicles` support REST-style querying by vehicle ID.

*Data update speed / latency:*
Question was raised as to whether the spec should provide guidance as to how up-to-date the `/vehicles` endpoint should be.  My notes do not indicate whether consensus was reached, though it was noted that GBFS does not define a maximum (or minimum) latency in the spec.

jfh01:
Need to ask a missing/lost status. Some notes here: https://github.com/openmobilityfoundation/mobility-data-specification/issues/67

hunterowens:
For now, no need to support REST style querying.

karcass:
I will open a PR for Agency 0.4.1 to add the `missing` status.  We can hash out the transitions.  Will post a link here and make sure that Provider and Agency are aligned well.

jiffyclub:
Hi all, this PR didn't come up on the last provider call so I wanted to check in about its status. What's needed here, is anything blocking this at this point?

provider/README.md
thekaveman: [Decided not to support REST-like querying for now](https://github.com/openmobilityfoundation/mobility-data-specification/pull/376#issuecomment-575301414)

provider/README.md
thekaveman: Including this field outside of `location` is consistent with how `status_changes` represents the same information.

provider/README.md
thekaveman: I think our use of words like MAY/SHOULD and MUST has been fairly consistent and we would continue with that approach here. But you raise a good point that visibility of this changing requirement is important. Maybe it would be useful to add a line to each endpoint description indicating if it is required or not:

```diff
Endpoint: /vehicles
+ Required/Optional: Optional
Method: GET
Schema: vehicles schema
data Payload: { "vehicles": [] }, an array of objects with the following structure
```

jfh01: As discussed on the 2/13 call, we may want to add a note about the future development path. e.g:

     Required/Optional: Optional (may become required in a future major release)

jfh01:
Discussed latency on 2/13 call. 

Strong agreement that /vehicles should provide as close to real-time as possible data. For reference, GBFS refers to "real-time or semi-real-time" data and "at this moment" status of vehicles.

Agreed that the feed should be updated at least every 5 minutes with the best available information about vehicle status at the time of update.

provider/README.md
```diff
@@ -336,8 +336,51 @@ When multiple query parameters are specified, they should all apply to the retur
...
+| `device_id` | UUID | Required | A unique device ID in UUID format, should match this device in Provider |
+| `vehicle_id` | String | Required | The Vehicle Identification Number visible on the vehicle itself, should match this device in provider |
+| `vehicle_type` | Enum | Required | see [vehicle types](#vehicle-types) table |
+| `propulsion_type` | Enum[] | Required | Array of [propulsion types](#propulsion-types); allows multiple values |
+| `last_event_time` | [timestamp][ts] | Required | Date/time when last status change occurred. See [Event Times](#event-times) |
+| `last_event_type` | Enum | Required | Event type of most recent status change. See [event types](#event-types) table |
+| `last_event_type_reason` | Enum | Required | Event type reason of most recent status change, allowable values determined by [`event type`](#event-types) |
+| `last_event_location` | GeoJSON [Point Feature][geo]| Required | Location of veicle's last event |
+| `current_location` | GeoJSON [Point Feature][geo] | Required | Current location of vehicle if different from last event, and the vehicle is not currently on a trip |
```
thekidder: Should this be `Required if Applicable`? Is there value in including this if it is a duplication of the `last_event_location` in many cases?

thekaveman: @thekidder [some earlier comments](https://github.com/openmobilityfoundation/mobility-data-specification/pull/376/files/15bad5cbb43abdcc2beb5e6171658a3510e27594#r335158972) were made about the possible distinction here.

provider/README.md
thekaveman: On the 2/13 call we decided to punt `removed`/`lost` to the Agency/Provider alignment work in #271 slated for the next major MDS release.

thekaveman:
@hyperknot [brought up an important question](https://github.com/openmobilityfoundation/mobility-data-specification/pull/376#issuecomment-542303887) earlier in the comment stream that didn't come up on the 2/13 call discussion of this proposal:

> What vehicles to include. This is the most important question.
> * Currently it states: Only vehicles that are currently in an `available`, `unavailable`, or `reserved` states should be returned in this payload.
>
> ...
> => I'd propose a simplified approach, which solves all these problems:
> Include all vehicles seen in the last n days (say 3 or 7), showing their last known state.

@karcass [showed support](https://github.com/openmobilityfoundation/mobility-data-specification/pull/376#issuecomment-543232388) for this simplified approach, and I also think this makes the most sense. 

Let's try a vote: 👍 this comment to support the notion of including all vehicles with an event in the last N days (let's say 3 days for starters). 👎 this comment to support the notion of including all vehicles in specific `event_type` states (the current wording of the proposal).

billdirks:
I would be great if some operators would vote on this since it was proposed by an operator to help solve a pain point for them. Right now I'm giving a tenative :thumbsdown: until I understand their needs better.

I believe the purpose of this PR is to include a real-time snapshot of the vehicles on the PROW. This allows one to count the number of vehicles in real-time in a way that is simpler than playing back status changes. Including vehicles that are known to be no longer on the PROW (ie removed, lost, and stolen vehicles) seems to be at odds with the stated objective since a consumer will now have to process the returned payload before they can understand what is on the PROW. The problem this suggestion asserts to solve is the non-inclusion of out-of-service vehicles. However, I believe this use case is already addressed since vehicles marked "unavailable" with reason "maintenance" are these out-of-service vehicles.

I think this could also complicate the implementation. The implementer needs to know about vehicles which were removed days earlier in addition to the current world view.

provider/README.md
```diff
@@ -336,7 +336,7 @@ When multiple query parameters are specified, they should all apply to the retur
 [Top][toc]
 
 ## Realtime Data
-All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in an available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should generally reconcile with data from the status changes enpdoint.
+All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should reconcile with data from the status changes enpdoint. This data should be refreshed at most every 5 minutes.
```
jfh01: So be a little pedantic, I think "at most" in the context of time can be ambiguous. "at most every 5 minutes" could be interpreted as "no more frequently than once every 5 minutes" rather than the intended interpretation of "at least once every five minutes."

I would suggest instead:


```suggestion
All providers should provide an endpoint that displays the current status of a providers vehicles on the PROW. Only vehicles that are currently in available, unavailable, or reserved states should be returned in this payload. Data in this endpoint should reconcile with data from the status changes endpoint. The data returned by this endpoint should be as close to realtime as possible, but in no case should it be more than 5 minutes out-of-date.
```

thekaveman:
@bhandzo I'm doing some work on the JSON Schema and notice that your branch is a ways out of date with `dev` here... mind if I rebase clean up future merge conflicts?

provider/README.md
thekaveman: Updated the language to reflect your suggestion @jfh01

thekaveman:
@bhandzo please sign the CLA as soon as possible.

provider/README.md
thekidder: Sorry, I neglected to follow up on this and that link is now broken. Is that discussion archived anywhere?

thekaveman: Made this update in the language and schema

thekaveman:
> ...
> Let's try a vote: 👍 this comment to support the notion of including all vehicles with an event in the last N days (let's say 3 days for starters). 👎 this comment to support the notion of including all vehicles in specific `event_type` states (the current wording of the proposal).

Decided on the call to leave language as-is (vehicles should show up on this endpoint if they are in one of the given states). Updating the schema to reflect this.

thekaveman (approved):
I'll plan to merge this later today (2020/02/27) unless I hear back from anyone. Please take a look if you have time, particularly at the JSON Schema.

@bhandzo @johnpena @hyperknot @ascherkus @thekidder @karcass @billdirks @rf- @jrheard @whereissean

quicklywilliam:
Just making note that there is a request for an "authenticated GBFS feed" with similar functionality in GBFS now that stable vehicle IDs have been removed. I believe several cities already require something like this, and several operators offer it.

I'd love to see a single standard get adopted here.
https://github.com/NABSA/gbfs/issues/237

PR #380, opened 2019-10-16
add associated_ticket. Fixes #269

### LINKED ISSUE #269: Add external_case_id to /status_changes endpoint for linking MDS and 311 data
One of the use cases for `provider` is to help cities monitor compliance with requirements around issue response. For example, if a city reports an misparked vehicle or other safety issue to a provider, the city will want to know if/when the issue was resolved.

Eventually, there will be a larger project to create all the hooks to facilitate full integration between city 311/issue tracking systems and mobility providers. But as a short-term step, would propose adding an optional `external_case_id` field to the `/status_changes` endpoint. This would allow for simple data linkages between 311 systems and provider issue management workflow.

For example, a city reports a problem to a provider (digitally or manually) and includes a case id for the related 311 issue. As the provider moves through its own workflow to resolve the issue, the 311 case id could continue to be tracked. When the provider makes a vehicle status change is made related to that case (such as relocating the misparked vehicle), the 311 case id would be provided back to the city via the `external_case_id` field on the `/status_changes` endpoint.

I'd welcome perspectives on the relative value of creating a lightweight integration hook now versus just waiting until we can properly design a full set of integrations that allow for more seamless digital workflows between 311 systems and mobility providers.

hunterowens:
### Explain pull request

In #269, it is brought up that having a 311 or CRM case id to associate with particular status changes may be useful. This PR allows a status change to have a CRM case ID / associated ticket with it. 
### Is this a breaking change

 * No, not breaking - optional field. 
 
### `Provider` or `agency`

 * `provider`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -300,7 +300,7 @@ Schema: [`status_changes` schema][sc-schema]
 | `event_location` | GeoJSON [Point Feature][geo] | Required | |
 | `battery_pct` | Float | Required if Applicable | Percent battery charge of device, expressed between 0 and 1 |
 | `associated_trip` | UUID | Required if Applicable | Trip UUID (foreign key to Trips API), required if `event_type_reason` is `user_pick_up` or `user_drop_off`, or for any other status change event that marks the end of a trip. |
-| `associated_ticket` | String | Optional | An associated ticket inside a 311 or CRM system that a City can tie an event to. | 
+| `associated_ticket` | String | Optional | Identifier for an associated ticket inside an Agency-maintained 311 or CRM system. | 
 
 ### Event Times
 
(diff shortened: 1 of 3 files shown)


thekaveman:
I'm still not clear on the data flow here - how does the Provider get this value from the Agency? I know @jfh01 had some early thoughts on this, I'm wondering if any Providers or Agencies could weigh in on if they have engaged in the type of interactions that would populate this value (i.e. are Agencies currently contacting Providers about individual vehicles in such a way that would facilitate this exchange?)

hunterowens:
From LA's end, providers have access to the backend of MyLA311, which means they might generate an event for a particular compliant ticket.

hunterowens:
so, company _X_ has access to 311, has a CRM ID, and knows which status_change was used to solve said CRM ID

PR #388, opened 2019-10-21
Add a section clarifying currency units; specify behavior when no currency is given

thekidder:
### Explain pull request

This PR fixes a couple of bugs introduced in #379.

Firstly, the schema only allows integral costs; this is insufficient for many currencies that contain a fractional part (including USD). This PR changes the types of the cost fields to `number` instead of `integer`. Since `number` is a superset of `integer`, this should not break backwards compatibility.

Additionally, the wording does not specify the behavior when the `currency` field is not present. This change clarifies the language to use the old definition (cents of USD) to keep backwards compatibility.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `provider`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -258,7 +258,7 @@ Schema: [`trips` schema][trips-schema]
 | `parking_verification_url` | String | Optional | A URL to a photo (or other evidence) of proper vehicle parking |
 | `standard_cost` | Integer | Optional | The cost, in the currency defined in `currency`, that it would cost to perform that trip in the standard operation of the System (see [Costs & Currencies](#costs--currencies)) |
 | `actual_cost` | Integer | Optional | The actual cost, in the currency defined in `currency`, paid by the customer of the *mobility as a service* provider (see [Costs & Currencies](#costs--currencies)) |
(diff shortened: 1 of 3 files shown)


thekaveman:
@thekidder #379 didn't necessarily introduce a bug of integer-only costs; this has been in the spec/schema since the beginning as the cost fields were defined as USD cents.

Your proposal here does highlight some ambiguity. Allowing fractional costs only works if we're defining the cost at the "dollar" level, vs. the "cents/centavos/etc." level. The currency code carries no information about the scale of the value.

The other way to do this is to define the cost fields in terms of the 1/100 equivalent ("cent") for the identified currency code. I don't know enough about international currencies to know if this would handle all cases.

thekidder:
> The currency code carries no information about the scale of the value.

 If no scale is given, I would assume the major unit is implied (i.e. dollars for USD). MDS could specify a different scale, but given that different currencies may have different minor units, this is less clear to me.

There is an exponent defined in ISO 4217, which defines (in powers of 10) how many minor units of a currency make up a major unit. But using this seems more likely to lead to edge cases - this field does not exist for all currencies.

thekaveman:
The Stripe example you referenced sounds like a good approach - do you want to add that wording into the descriptions?

thekidder:
Yes, agreed. Will update this shortly.

hunterowens:
@thekidder any update on the the update?

thekidder:
Updated the wording here, using Stripe's behavior as an example.

There may still be some ambiguity here - I am not a currency expert - but I think this is an improvement.

PR #400, opened 2019-10-31
Small typo in policy spec

TonioGarcia07:
### Explain pull request

This is simply to fix a typo in the example provided for the policy.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

It will have no impact on the APIs

### DIFF
--- a/policy/examples.md
+++ b/policy/examples.md
@@ -144,10 +144,10 @@ This Policy allows scooters and bikes can be in the public right-of-way for up t
       "geographies": [
         "12b3fcf5-22af-4b0d-a169-ac7ac903d3b9"
       ],
-      "statuses": [
-        "unavailable",
-        "trip"
-      ],
+      "statuses": {
+        "unavailable": [],
+        "trip": []
+      },
       "vehicle_types": [
         "bicycle",
         "scooter"


karcass:
Oof, good catch.  Thanks!

thekaveman:
Thanks @TonioGarcia07!

This points to a question I had with the Policy README and the field list for [Rules objects](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/policy#rules). I was a little confused about the Type of `statuses`. It seems like from the examples that these are related to `Agency`; @karcass can you weigh in? And if so, maybe @TonioGarcia07 can update this PR so that field links back to the appropriate place in the `Agency` spec to make it more clear.

karcass:
Policy is intended to work with either Provider or Agency (neither is explicitly required).  The examples given are all Agency-focused, because they are all directly lifted from what's deployed by LADOT, but that should not be taken as Agency-specificity.

The `statuses` field in the Rule description refers to the status of the vehicle, and that status is described in Provider in the [event types table](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/provider#event-types) and in Agency in the last column of the [vehicle events table](https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/agency#vehicle-events). 
 
Agency has additional statuses (`trip`, `inactive`, `elsewhere`) but the Agency statuses are a superset of Provider's.  

A `Rule` can be applied to not just a status (e.g. `available`) but to an event that put the vehicle into that status, and in this case you have to craft the rule according to whether you're using Agency or Provider.  So if you want a rule that applies to _just_ Provider drop-offs, for Provider you would specify the `statuses` field as

```
  "statuses": {
    "available": ["service_start", "rebalance_drop_off", "maintenance_drop_off"]
   }
```

in Agency it would be:

```
  "statuses": {
    "available": ["provider_drop_off"]
   }
```

In subsequent releases of MDS we plan to make the statuses and transitions to align as closely as possible, to hopefully eliminate this difference.

hunterowens:
thanks for the last minute catch @TonioGarcia07 !

thekaveman:
@karcass Thanks for that clarification. I guess I wasn't asking so much from the "how do I implement this" point of view, but the much simpler "what does this data type refer to, elsewhere in the spec", the purely documentation point of view. 

Since it sounds like the answer is *both*/*either* I think it would be helpful to link to the corresponding event tables in `Agency` and `Provider` to clear up any future confusion. And agreed that we can likely clean this all up by resolving differences in statuses.

karcass:
@thekaveman I know I was over-answering a bit, but I wanted to make sure there weren't any lingering questions.  And agree that links would be helpful.  Would you like me to make that addition prior to 0.4.0, or is waiting to 0.4.1 reasonable?

thekaveman:
@karcass no worries, I love verbosity! I just submitted a PR #402 can you give a quick review and make sure it's correct?

PR #402, opened 2019-10-31
Add clarity on rule status field

thekaveman:
### Explain pull request

Linking to relevant status tables in Agency and Provider for reference clarity.

### Is this a breaking change

 * No, not breaking

### `Provider` or `agency`

 * `policy`


karcass:
@thekaveman I'm too late!  :)  All good.

Issue #386, opened 2019-10-18
Moving MDS from the City of Los Angeles to the Open Mobility Foundation


jfh01:
Hi, I’m Jascha Franklin-Hodge, the Executive Director of the Open Mobility Foundation.

In the next few weeks, ownership of the `mobility-data-specification` repository, along with `mds-core` and `mds-compliance-mobile` will be [transferred](https://help.github.com/en/articles/transferring-a-repository) from the @CityOfLosAngeles organization on GitHub to the @openmobilityfoundation organization. This reflects the donation of the MDS codebase to the Open Mobility Foundation by the City of Los Angeles. The transfer will take place after the release of MDS 0.4.0.

The success of MDS has only been possible through the engagement and contributions of the MDS community. OMF and LADOT want to build on this great work, and bring the community along as the OMF takes on the role of shepherding future development. By moving the MDS codebase from LADOT to OMF, we aim to allow the existing MDS community to continue to thrive while providing new, multistakeholder governance that allows this work to grow as a true open source project. You can read more about the Open Mobility Foundation here: [http://www.openmobilityfoundation.org/](http://www.openmobilityfoundation.org/).

Below are some details about how this code transfer will work. **Please respond to this issue with feedback, suggestions, and questions.** You can also email jascha@openmobilityfoundation.org.

---

**_Technical considerations_**
When the repositories move, all existing history and work-in-progress will transfer over. GitHub will automatically redirect links and forks based on the old repository location. However, GitHub recommends updating any local clones to point to the new repository URL. See [here](https://help.github.com/en/articles/transferring-a-repository) for details.

Some issue and PR assignments may disappear as part of the repository transfer. Assignments are not widely used in the current repositories, so we do not anticipate any significant impacts.

**_Process considerations_**
[...]

PR #404, opened 2019-10-31
Switch CODEOWNERS to use OMF teams

jfh01:
As part of the code transfer from LADOT to OMF (#386), we are switching CODEOWNERS to use OMF's GitHub Teams, rather than individually naming maintainers. This PR should be merged as part of the code transfer process.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,10 +1,13 @@
-## Below is the MDS CODEOWNERS file, which dictates who is required for review on any given file. 
+## Below is the MDS CODEOWNERS file, which dictates who is required for review on any given file.
 
-## All MDS approvals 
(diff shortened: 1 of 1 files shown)


hunterowens:
This should probably have OMF admin group have write on changes to the groups along with the maintainers. 

ie, 
`/agency/*  @openmobilityfoundation/city-services-wg-maintainers @openmobilityfoundation/admin` so that a city service wg approval isn't required for merges in agency. 

This is useful for when you're madly having maintainers making PRs that they themselves cannot merge during the run up to a release.

jfh01:
> This should probably have OMF admin group have write on changes to the groups along with the maintainers.

All set. Thanks.

PR #409, opened 2019-11-04
Add missing rule_id to rule in Policy spec

antoinewg:
### Explain pull request

Hello, it seems that the property `rule_id` is missing in the description of the Rule. It makes sense to have one, and the examples have a rule_id.

### Is this a breaking change

No, not breaking

### Impacted Spec

 * `policy`


TonioGarcia07:
@hunterowens could you recheck please, I've added a commit to modify the examples as well. Cheers !

PR #411, opened 2019-11-07
Rename car to carshare to match top-level spec description

avatarneil:
### Explain pull request

In this commit https://github.com/openmobilityfoundation/mobility-data-specification/commit/f720a092397d9a12bcf346a03a6707b5e12f9f5f the top-level name of `car` was switched to `carshare`. However, I noticed that change did not seem to bubble down to the agency and provider spec level. This changes the `vehicle_types` enum to have `carshare` instead of `car`. This would be a breaking change, however I find the discontinuity between the types in the top level compared to agency and provider to be pretty strange... If this isn't intended to be a rename that goes down to the data-type level, feel free to close the PR!

### Is this a breaking change

 * Yes, breaking

### Impacted Spec

 * `agency`
 * `policy`
 * `provider`

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -264,11 +264,11 @@ A standard point of vehicle telemetry. References to latitude and longitude impl
 
 ### Vehicle Type
 
-| `type`    |
-| --------- |
-| `bicycle` |
-| `car`     |
-| `scooter` |
+| `type`     |
+| ---------- |
+| `bicycle`  |
+| `carshare` |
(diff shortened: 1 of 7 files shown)


thekaveman:
I don't know that this spec-level change is needed, especially as it would be breaking. The vehicle is a `car` after all, it just happens to be a car in a car-sharing fleet.

avatarneil:
Sounds good to me! I'll nuke the PR.

PR #412, opened 2019-11-08
Update example to use "maximum" instead of "limit"

antoinewg:
### Explain pull request

Hello, just a small PR to update the example (Idle time) as it has `limit` where it should be `maximum` to configure the rule.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `policy`


thekaveman:
I think this can go to 0.4.1 right? Really a typo fix.

PR #413, opened 2019-11-14
change CLEVR url

valclevrmobility:
### Explain pull request

changed CLEVR mds url

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `provider`


valclevrmobility:
it's just to update the URLs which we already submitted before but for some reason that change was reverted and currently master again contains wrong URLs. 

so it's repeated request about the same

valclevrmobility:
@hunterowens can you please approve this request?

hunterowens:
yes sorry, had to change the base branch.

PR #414, opened 2019-11-15
fix CLEVR mds url

valclevrmobility:
### Explain pull request

fix CLEVR mds url

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `provider`


thekaveman:
@valclevrmobility I am assuming this is a follow-up to #413?

thekaveman:
@valclevrmobility Any feedback here? We're preparing the 0.4.1 release and I don't want this to get missed if it should go in. Also I don't want this to go in if it is a duplicate/superseded.

valclevrmobility:
Hi Kegan,

as far as I see our request was approved, so we have no issues with it.
Thanks !

// Val
CLEVR Mobility

thekaveman:
Yes, I approved it. The question is whether this PR replaces the (already merged) #413 or if it was a duplicate. 

To confirm, this PR will be merged and the `mds_api_url` for CLEVR will change as follows:

```diff
-https://portal.clevrmobility.com/api/mds/
+https://portal.clevrmobility.com/api/mds
```

PR #416, opened 2019-11-25
Add "moped" vehicle type

pvanliefland:
### Explain pull request

This is a first attempt at adding the "moped" vehicle type. The company I'm working at (https://www.lab-box.com/) has, among others, a multimodal service (https://poppy.be/en/) that offers shared electric mopeds.

There are at least two other electric mopeds providers in Brussels (https://scootysharing.be/ and https://felyx.com/).

### Is this a breaking change

I'm not sure

### Impacted Spec

 * `agency`
 * `policy`
 * `provider`

### Additional context

Not sure about the way to implement this properly - I used https://github.com/openmobilityfoundation/mobility-data-specification/pull/219 as a basis.

There was also a mention somewhere of using `generate_schemas.py` to add a mode, but the "car" PR didn't follow that route - let me know if you prefer another implementation.


jfh01:
We might consider borrowing from SAE J3194 for this or other future vehicle types https://saemobilus.sae.org/content/J3194_201911/

pvanliefland:
> [CLA assistant check](https://cla-assistant.io/openmobilityfoundation/mobility-data-specification?pullRequest=416)
> All committers have signed the CLA.

Please note that I am an independent consultant, with a contract for a company that is not a member of the OMF - I was distracted and thought I could already sign the individual agreement before the company in question signs the entity agreement. I think this is a mistake.

I have asked a representative of the company to sign the ECLA.

pvanliefland:
> > [CLA assistant check](https://cla-assistant.io/openmobilityfoundation/mobility-data-specification?pullRequest=416)
> > All committers have signed the CLA.
> 
> Please note that I am an independent consultant, with a contract for a company that is not a member of the OMF - I was distracted and thought I could already sign the individual agreement before the company in question signs the entity agreement. I think this is a mistake.
> 
> I have asked a representative of the company to sign the ECLA.

The company I'm working for (Lab-Box) has signed the ECLA.

thekaveman:
Off-hand comment, but it seems like we should get away from having to update the README every time a new `vehicle_type` comes online. Can we use a more generic phrase there like "shared mobility devices" (instead of re-listing the enum values explicitly)?

HenriJ:
Looks good to me for the Agency API.
@hunterowens your opinion for Provider ?

@jfh01 indeed, it would be interesting to use a more detailed classification for the vehicles, it would be a breaking change though. Maybe we can open a dedicated issue to replace or enrich the existing vehicle field ?

jfh01:
> @jfh01 indeed, it would be interesting to use a more detailed classification for the vehicles, it would be a breaking change though. Maybe we can open a dedicated issue to replace or enrich the existing vehicle field ?

Agreed. Opened #426 to address the bigger concern.

jfh01:
@thekaveman @sarob @HenriJ Any reason this is on the Future milestone? Seems like this could go into 0.4.1. Mopeds (Revel, in particular) are rapidly expanding and this seems like a useful addition for now, regardless of where we go with this long-term.

thekaveman:
@jfh01 

> Any reason this is on the Future milestone? Seems like this could go into 0.4.1.

This change could be considered breaking in that existing code with enumerations for these vehicle types would need to be adjusted to include the new value. But it's an easy enough change and I'm definitely not against putting it in 0.4.1. Will put it on the next agenda and make sure the larger group sees no issues.

jfh01:
Brought up on the city services working group call and raised the breaking change issue. No one spoke up against putting this in 0.4.1, but looking forward to discussing next week in Provider Services.

thekaveman:
Also brought this up on the Provider Services WG call on 1/30 and no one raised the breaking change issue. Approving for 0.4.1.

PR #420, opened 2019-12-20
Add Bolt Technology as MDS provider

artem-vereshchaka:
### Explain pull request

This PR adds new MDS provider

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * no one


artem-vereshchaka:
@thekaveman, this is the root GBFS endpoint. for each city GBFS feed is different, and URL also differs.
I saw when the root GBFS URL respond with available cities but didn't see it in the specification directly.

Issue #417, opened 2019-11-26
Confusion about Reserved state


concept47:
### Is your feature request related to a problem? Please describe.

I was looking at [the provider event types spec](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/provider#event-types) as I was getting ready to implement the `reserved` event update. I was curious about why in the spec for the `reserved` event_type it says in the description

"A customer reserves a device **(even if trip has not started yet)**"

but then an associated_trip (UUID) is required because the `reserved` event maps to the event_type_reason `user_pick_up`

### Describe the solution you'd like

associated_trip (UUID) not required for `reserved` event_type

### Is this a breaking change

 * No, not breaking

### Impacted Spec

For which spec is this feature being requested?

 * `provider`

thekaveman:
I would be more inclined to update the spec language to remove the *even if trip has not started yet* piece. Associated trip UUID for the trip-specific events seems like something we would want to continue requiring.

concept47:
I'm fine with that too. Although I am curious as to what's your rationale for that is @thekaveman?

It seems odd to be to require a UUID for a trip that has not started, but then again I'm an absolute newb to all this, so I'm just trying to understand where you are coming from.

Thanks!

geobir:
I am not sure to understand perfectly but for the "A customer reserves a device **(even if trip has not started yet)**"
If the provider let the user reserved vehicle before they start the trip.

For the UUID I understand your concern giving the trip UUID before the trip start can be interesting. Even more if the reservation is canceled.

[associated_trip](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/provider#status-changes)
> associated_trip | UUID | Required if Applicable | Trip UUID (foreign key to Trips API), required if event_type_reason is user_pick_up or user_drop_off, or for any other status change event that marks the end of a trip.

We may want to switch user_pick_up event type by another or remove it from the applicable one.

concept47:
> For the UUID I understand your concern giving the trip UUID before the trip start can be interesting. Even more if the reservation is canceled.

exactly my concern @geobir. Thanks for stating that better than I did 😅 

> We may want to switch user_pick_up event type by another or remove it from the applicable one.

This is what I was hoping was possible ...

thekaveman:
[...]

PR #421, opened 2020-01-02
event_type reserved now explicitly means taking possession

dirkdk:
### Explain pull request

The status_change event_type `reserved` has caused confusion (see https://github.com/openmobilityfoundation/mobility-data-specification/issues/417). More and more providers support reservation as in future hold for a vehicle. It seemed that `reserved` was actually not meant for this use case, but just for taking possession. The requirement to have a trip_id hints at this. This PR more explicitly states the intent of the event_type `reserved` to be only when a user takes physical possession of the vehicle, not to reserve for future use. 

### Is this a breaking change

Although the API definitions strictly remain the same, both API clients and implementations may have to be adjusted.

 * I'm not sure

### Impacted Spec

 * `provider`

### Additional context

Agency has `trip_start` and `reserved`, this means that Provider `reserved` is comparable to Agency `trip_start`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -367,11 +367,11 @@ Without an `event_time` query parameter, `/status_changes` shall return a `400 B
 | `event_type` | Description | `event_type_reason` | Description |
 | ---------- | ---------------------- | ------- | ------------------ |
 | `available` | A device becomes available for customer use | `service_start` | Device introduced into service at the beginning of the day (if program does not operate 24/7) |
-| | | `user_drop_off` | User ends reservation |
+| | | `user_drop_off` | User ends possession |
 | | | `rebalance_drop_off` | Device moved for rebalancing |
 | | | `maintenance_drop_off` | Device introduced into service after being removed for maintenance |
 | | | `agency_drop_off` | The administrative agency (ie, DOT) drops a device into the PROW using an admin code or similar | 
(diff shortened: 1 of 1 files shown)


thekidder:
What happens when a device is reserved but not yet in a customer's possession? Should the vehicle remain in the `available` state, despite the fact that it is not rentable?

I think it would be helpful to add a new event_type_reason for the `unavailable` state to address this.

A proposal to clarify this, though it is a breaking change:
* Rename the existing `reserved` event_type to `on_trip` or similar
* Add a new `reserved` event_type_reason to the unavailable state.

This accurately captures the distinction here.

Retzoh:
> What happens when a device is reserved but not yet in a customer's possession? Should the vehicle remain in the `available` state, despite the fact that it is not rentable?
> 
> I think it would be helpful to add a new event_type_reason for the `unavailable` state to address this.
> 
> A proposal to clarify this, though it is a breaking change:
> 
> * Rename the existing `reserved` event_type to `on_trip` or similar
> * Add a new `reserved` event_type_reason to the unavailable state.
> 
> This accurately captures the distinction here.

This would re-conciliate the provider and agency APIs as far as trip and reservation are concerned.

karcass:
@Retzoh `trip` would match Agency.  I think this would be a good change for 0.5.0.

dirkdk:
closing this for PR https://github.com/openmobilityfoundation/mobility-data-specification/pull/439

PR #429, opened 2020-01-16
Adds a "missing" status to the "agency" specification.

wellorder:
### Explain pull request

This PR creates a `missing` status as part of the agency specification. It is often the case that scooters go missing for a short period of time due to GPS issues, connectivity issues, etc. Under the current spec, this is supposed to be handled by deregistering the scooter and setting its state to `removed`, but this makes it difficult to distinguish scooters which have been removed by our operations people and scooters which have simply gone missing, which causes confusion when someone comes across a scooter in the wild which shows up as `removed`. What's more, deregistering and re-registering a scooter is more work to implement than changing the state to and from `missing` would be.

### Is this a breaking change

 * Yes, breaking

As written, this removes the `missing` transition to `removed`, which would break existing implementations that use this transition.

### Impacted Spec

 * `agency`

### Additional context

This follows up discussion in https://github.com/openmobilityfoundation/mobility-data-specification/issues/67 and https://github.com/openmobilityfoundation/mobility-data-specification/pull/376#issuecomment-575302567 . Mark said he would work on such a PR, but since I had already started on this one I figured I could post it.  I was holding off until I had figured out `provider` changes, but I think he was right and that can be put in a separate PR.


karcass:
Thanks @wellorder !

karcass:
Propose we fold this into the Agency/Provider reconciliation project #271

thekaveman:
`missing` was included as an `event_type` with #506. Thanks for the work on it @wellorder!

Issue #432, opened 2020-01-24
Policy: inconsistent types for timestamps (string vs. number)


ascherkus:
**Describe the bug**
The examples contained in the policy README.md have all timestamps declared as strings, but the examples.md uses numbers.

**To Reproduce**
Steps to reproduce the behavior: N/A

**Expected behavior**
I believe for consistency across the MDS specification that timestamps are milliseconds represented as numbers.

**Additional context**
Came across this while prototyping a tool for #423.

Happy to send out a PR - just want clarification before doing so!

karcass:
Yes, we should be consistent.  Good catch @ascherkus.  Please send a PR while I look at the other two issues you filed.  :)

ascherkus:
Will take a crack!

schnuerle:
This issue is closed with PR #523.

PR #435, opened 2020-01-24
Policy: use a consistent definition for timestamps.

ascherkus:
Aligns the Policy text specification for timestamps to the Provider and Agency
APIs, which use integer milliseconds from Unix epoch.

Fixes issue #432.

### Explain pull request

Fixes mismatch of string vs. number for timestamps.

### Is this a breaking change

Technically, yes, but in reality I'm not sure.

The `mds-core` reference implementation already defined timestamps as numbers:
https://github.com/openmobilityfoundation/mds-core/blob/09930c89ea4c71684140aa0046ba9727be40bd79/packages/mds-types/index.ts#L144

That being said, if other implementations exist that were using strings it's possible there will be breakage.

### Impacted Spec

 * `policy`

### Additional context

N/A

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -62,6 +62,10 @@ The response to a client request must include a valid HTTP status code defined i
 - **404:** Not Found: Object(s) do not exist.
 - **500:** Internal server error.
 
+#### Timestamps
+
+As with the Provider API, `timestamp` refers to integer milliseconds since Unix epoch.
+
 #### Error Responses
 
 ```json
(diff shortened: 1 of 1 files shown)


schnuerle:
This looks straightforward enough to me to merge to dev for 1.0.0.  Will check with @thekaveman and get plan for starting 1.0.0 PR merges.

avatarneil:
@schnuerle I'm surprised that Github is even allowing a PR from a branch that has been deleted... What do you think about capturing these changes, closing this PR, and cutting a new PR with the same changes?

schnuerle:
@avatarneil I think if @ascherkus or we can't resolve it in the next week, that's what I'll do.

schnuerle:
@ascherkus @avatarneil I'm going to close this PR tomorrow and open a new one with these changes since the ascherkus:Policy-timestamps branch is no longer there.

schnuerle:
Closing this PR and instead using the duplicate PR #523.

PR #442, opened 2020-02-05
Add stop references to trips and status_changes

### LINKED ISSUE #428: Modify /trips to represent docked start/end locations

avatarneil:
### Explain pull request
Resolves #428 following @jfh01's suggestion to use GeoJSON Foreign Members. Relates pretty closely to the open PR #427, and is dependent on that being merged first.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `provider`

### DIFF
--- a/general-information.md
+++ b/general-information.md
@@ -108,6 +108,8 @@ All String fields, such as `vehicle_id`, are limited to a maximum of 255 charact
 
 ## Stops
 
+**Stops** describe vehicle trip end locations in a pre-designated physical place. They can vary from docking stations with our without charging, corrals with lock-to railings, or suggested parking areas marked with spray paint.  **Stops** are used in both [Provider](/provider#stops) (including routes and event locations) and [Agency](/agency#stops) (including telemetry data).
+
 | Field                  | Type                                                        | Required/Optional | Description                                                                                  |
 |------------------------|-------------------------------------------------------------|-------------------|----------------------------------------------------------------------------------------------|
 | stop_id                | UUID                                                        | Required          | Unique ID for stop                                                                           |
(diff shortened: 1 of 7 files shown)


thekaveman:
Two comments:

1. I know @jfh01 suggested the use of Foreign Members, and I don't see a specific issue there. I just wonder if there is any reason not to extend our existing use of the `properties` object already present in GeoJSON Features?

2. Regardless of the answer to 1, the [schema](https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/schema/generate_schemas.py#L130) will need to be updated to handle the additional field.

ascherkus:
Chiming in ... I'm also a bit hesitant on relying on GeoJSON foreign members vs. properties (similar to the hesitancy mentioned in #441).

Given:
  - The [GeoJSON spec cautions against it](https://tools.ietf.org/html/rfc7946#section-6.1)
  - We're already using `properties`
  - Other sources of GIS data seen in this space (notably GeoJSON produced from Shapefiles such as [Santa Moncia's drop zones](https://gis-smgov.opendata.arcgis.com/datasets/shared-mobility-drop-zones)) map attributes to properties

... I'm inclined to suggest sticking with `properties` for Feature-specific metadata, unless @jfh01 has some additional insight/experience into foreign members :)

Alternatively start/end stop ids could be expressed higher in the hierarchy alongside `start/end_time` -- the tradeoff there would being unable to express a trip that stopped at multiple stops (unclear to me if such a trip would exist, or if that would be represented as two trips with separate start/stop pairs). If I had to pick, I think sticking with Feature properties with schema validation accomplishes the same goal while retaining flexibility vs. adding more named fields at the top-level `trips` object.

jfh01:
No particular insight here. I'm happy with either direction. Thanks!

avatarneil:
Updated the PR to move the `stop` definitions to the top level of the`trips` and `status_changes` objects.

thekaveman:
@rgangopadhya 

> Hmm, I do think this could be useful, though. I must've missed it on the call but did we agree to remove this?

I think we decided to move it nearer to the geometry, inside the `properties` object of relevant GeoJSON Feature (`status_change.event_location`, the first/last point in `trip.route`). This PR has been updated to reflect the changes for `trip.route` but it looks like we still need something for `status_changes.event_location` @avatarneil

avatarneil:
Hi @thekaveman, I've updated the `event_location` definition. Sorry for the delay!

thekaveman:
I rebased on the latest `dev` to incorporate Reconciliation changes. Let's wait to merge this after #427 so we can update references to Stops in the Provider README.

PR #443, opened 2020-02-05
Add missing/lost event to provider spec

jiffyclub:
### Explain pull request

This is a companion pull request to #429 to add a "missing"/"lost" event to the provider spec. My hope is to unblock #376 (`/vehicles` endpoint). I'm not at all sure I've done this correctly, so pointers appreciated.

It might also be worth noting that this event type would be changing to "unknown" in the future per @thekidder's suggestion in https://github.com/openmobilityfoundation/mobility-data-specification/issues/271#issuecomment-582243658.

### Is this a breaking change

 * No, not breaking

This only adds a field so I don't think it's breaking, but not 100% sure on the conventions here.

### Impacted Spec

 * `provider`


jiffyclub:
Closing per discussion on the OMF Provider call today (2020-02-13), we'll skip adding missing/lost at this point since it'll be included in the overhaul of vehicle status/event type coming in 0.5.0 (see discussion in #271).

Issue #422, opened 2020-01-06
Add a "Versioning" section to the Agency spec as in the Provider spec


Retzoh:
### Is your feature request related to a problem? Please describe.

There is no specification on how versioning works for the Agency API :
A provider cannot specify the version used for the data he is trying to push.

### Describe the solution you'd like

For consistency, I propose to use the [provider API versioning system](https://github.com/openmobilityfoundation/mobility-data-specification/blob/0.4.0/provider/README.md#versioning), where the version used is specified in the request Accept header.

If we re-use the `application/vnd.mds.provider+json` media-type  or rename it to `application/vnd.mds+json` in the provider spec, we could easily centralise the versioning spec in one place and expand it to further APIs.

### Is this a breaking change

A breaking change would require consumers or implementors of the API to modify their code for it to continue to function (ex: renaming of a required field or the change in data type of an existing field). A non-breaking change would allow existing code to continue to function (ex: addition of an optional field or the creation of a new optional endpoint). 

 * I'm not sure

We can decide to make it breaking or not :

Adding implicit support for the previous minor version like in the provider spec makes it non-breaking :

> Since versioning was not added until the 0.4.1 release, if the Accept header is not set as specified above, the Agency API must respond as if version 0.3 was requested.

This would stay consitent with the [two concurrent version support](https://github.com/openmobilityfoundation/mobility-data-specification/blob/80e628debc5e438625b094673665471edc2351fd/ReleaseGuidelines.md#ongoing-version-support) from the release guidelines.

Enforcing this retro-actively would make it breaking.

### Impacted Spec

For which spec is this feature being requested?

 * `agency` : yes

 * `policy` : we could add it there too, so a provider requesting the API can ask for a specific version in return

### Describe alternatives you've considered

The classic alternative would be to enforce a version number in URLs.
But this would be inconsistent with the provider spec and so far, MDS does not put constrains on API URLs outside from the endpoint and query-parameters. 

### Additional context

None

Retzoh:
Hi @sarob, this issue should be labeled "Agency"

karcass:
Agreed, I'll fix it in 0.4.1.  Thanks @sarob.

Retzoh:
@karcass have you made any progress? I would be happy to help.

karcass:
@Retzoh I have not, I have been heads-down on the Agency/Provider naming reconciliation.  If you want to take a swing at it, please do!

Retzoh:
[...]

PR #444, opened 2020-02-06
add versioning for Agency and Policy API

Retzoh:
### Explain pull request

This is a pull request attempting to solve #422 without introducing any breaking changes to the `provider` API.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

* `agency`: adding versioning section
* `policy`: adding versioning section
* `provider`: moving the versioning section

### Additional context

None

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -27,9 +27,7 @@ The following information applies to all `agency` API endpoints. Details on prov
 
 `agency` APIs must handle requests for specific versions of the specification from clients.
 
-Versioning must be implemented as specified in the [`Provider API versioning section`][Provider/versioning].
-
-Since versioning was not added until the 0.4.1 release, if the Accept header is not set as specified above, the Agency API must respond as if version 0.3 was requested.
+Versioning must be implemented as specified in the [`General information versioning section`][general-information/versioning].
 
 ## Authorization
 
@@ -316,4 +314,4 @@ A vehicle may have one or more values from the `propulsion`, depending on the nu
 [Top][toc]
 
 [toc]: #table-of-contents
-[provider/versioning]: /provider/README.md#versioning
+[general-information/versioning]: /general-information.md#versioning
--- a/agency/README.md
+++ b/agency/README.md
@@ -315,4 +315,5 @@ A vehicle may have one or more values from the `propulsion`, depending on the nu
 
 [Top][toc]
 
+[toc]: #table-of-contents
 [provider/versioning]: /provider/README.md#versioning
--- a/agency/README.md
+++ b/agency/README.md
@@ -29,6 +29,8 @@ The following information applies to all `agency` API endpoints. Details on prov
 
 Versioning must be implemented as specified in the [`Provider API versioning section`][Provider/versioning].
 
+Since versioning was not added until the 0.4.1 release, if the Accept header is not set as specified above, the Agency API must respond as if version 0.3 was requested.
+
 ## Authorization
 
 When making requests, the Agency API expects `provider_id` to be part of the claims in a [JWT](https://jwt.io/)  `access_token` in the `Authorization` header, in the form `Authorization: Bearer <access_token>`. The token issuance, expiration and revocation policies are at the discretion of the Agency.
--- a/agency/README.md
+++ b/agency/README.md
@@ -7,6 +7,7 @@ This specification contains a collection of RESTful APIs used to specify the dig
 
 ## Table of Contents
 
+* [General Information](#general-information)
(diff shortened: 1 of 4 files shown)


agency/README.md
```diff
@@ -18,6 +19,18 @@ This specification contains a collection of RESTful APIs used to specify the dig
...
+
+The following information applies to all `agency` API endpoints. Details on providing authorization to endpoints is specified in the [Authorization](#authorization) section.
+
+### Versioning
+
+`agency` APIs must handle requests for specific versions of the specification from clients.
+
+Versioning must be implemented as specified in the [`Provider API versioning section`][Provider/versioning].
+
+Since versioning was not added until the 0.4.1 release, if the Accept header is not set as specified above, the Agency API must respond as if version 0.3 was requested.
```
Retzoh: Alternatives would be :
1. Copy the content of the provider versioning section.
2. Create a separate `Versioning` file somewhere and reference it from the Provider, the Agency and the Policy specs.

Option 1 seems hard to maintain to me.
Option 2 would be a "cleaner" way of doing this but it requires to change the provider spec too.

agency/README.md
```diff
@@ -299,3 +312,7 @@ A vehicle may have one or more values from the `propulsion`, depending on the nu
 | `error`             | String   | Error message string   |
 | `error_description` | String   | Human readable error description (can be localized) |
 | `error_details`     | String[] | Array of error details |
+
+[Top][toc]
+
+[provider/versioning]: /provider/README.md#versioning
```
Retzoh:
```suggestion
[toc]: #table-of-contents
[provider/versioning]: /provider/README.md#versioning
```

agency/README.md
Retzoh: The same will then be done for the Policy spec, I just want to get to a consensus first.

thekaveman: I like Option 2, seems like we might run into more of these use-cases and a `general-information.md` or similar file would be useful.

Retzoh:
@thekaveman @jrheard I have updated the PR to move the provider Versioning section to a separate general-information file and link it from the other APIs. Let me know what you think.

general-information.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Table of Contents
+
+* [Versioning](#versioning)
+
+## Versioning
+
+MDS APIs must handle requests for specific versions of the specification from clients.
+
+Versioning must be implemented through the use of a custom media-type, `application/vnd.mds.provider+json`, combined with a required `version` parameter.
```
Retzoh: We could remove the `.provider` part of the header since it is now available for all the APIs but it would make the changes breaking for provider.

This could be done later in the 0.5 timeframe if we want.

provider/README.md
```diff
@@ -21,45 +21,7 @@ Currently, the provider API is implemented for dockless scooter and bikeshare. T
...
-```
-
-The response will include the most preferred supported version in the `Content-Type` header. For example, if only `0.3` is supported:
-
-```http
-Content-Type: application/vnd.mds.provider+json;version=0.3
-```
-
-The client can use the returned value verbatim as a version request in the `Accept` header.
+Versioning must be implemented as specified in the [`General information versioning section`][general-information/versioning].
```
Retzoh: I just moved the `versioning` sub-section to the `general-information` file since this is the critical part for making the agency and policy API usable.

The other sub-sections of the provider `General information` section could be moved later.

general-information.md
```diff
@@ -0,0 +1,48 @@
...
+Clients must specify the version they are targeting through the `Accept` header. For example:
+
+```http
+Accept: application/vnd.mds.provider+json;version=0.3
+```
+
+> Since versioning was not available from the start, the following APIs provide a fallback version if the `Accept` header is not set as specified above:
+> - The `provider` API must respond as if version `0.2` was requested.
+> - The `agency` API must respond as if version `0.3` was requested.
+> - The `policy` API must respond as if version `0.3` was requested.
```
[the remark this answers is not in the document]
Retzoh: Thanks for catching this, I have updated the PR.

general-information.md
thekaveman: Seems like a good idea, want to open an issue for tracking purposes?

Retzoh:
@thekaveman this has been validated on the Agency side too. Can you merge it ?

general-information.md
Retzoh: I have done it with #455

karcass:
LGTM from the Agency/Policy side

PR #449, opened 2020-02-18
Add the currently supported MDS versions

sarob:
### Explain pull request

This is a minor update to the release guidelines

### Is this a breaking change

 * No, not breaking

### Impacted Spec

N/A

### Additional context

This is a documentation update


thekaveman:
@sarob Just a quick follow-up, that Wiki section you have linked is exactly what I had in mind as the place to advertise this (maybe with a slightly different heading like *Supported MDS Releases* or similar). My concern was with re-listing supported versions in this document as well, and having to update this document (in addition to the Wiki, etc.) for each release.

PR #456, opened 2020-02-27
Making required/optional endpoints more explicit

thekaveman:
### Explain pull request

As discussed on the [2020-02-13 Provider Services call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.02.13-(Provider-Services-WG)) we'd like to more easily identify which endpoints are required, which are optional, and any anticipated changes to this status.

### Is this a breaking change

* No, not breaking

### Impacted Spec

* `provider`


babldev:
+1 to indicate trips + status changes as the only required APIs for now

PR #461, opened 2020-03-04
Clarify success response for telemetry update

Retzoh:
# MDS Pull Request

## Explain pull request

The telemetry spec only says that when new telemetry are successfully received, the server should:

> Responds with number of successfully written telemetry data points and total number of provided points.

But the spec does not specify how those two numbers should be written.

## Is this a breaking change

* Yes, breaking

## Impacted Spec

* `agency`

## Additional context

None

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -209,7 +209,7 @@ Body Params:
 
 | Field     | Type                           | Field Description                                                                                       |
 | --------- | ------------------------------ | ------------------------------------------------------------------------------------------------------- |
-| `result`  | String                         | Responds with number of successfully written telemetry data points and total number of provided points. |
+| `result`  | String                         | Responds with number of successfully written telemetry data points and total number of provided points. Format to use: `{success}/{total}`, both amounts as integers. |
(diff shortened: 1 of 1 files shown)


Retzoh:
This PR would be rendered obsolete by #462

schnuerle:
Waiting to merge to see if we move forward on #462 for the 1.0.0.  It needs some decisions and a PR.

agency/README.md
```diff
@@ -209,7 +209,7 @@ Body Params:
 
 | Field     | Type                           | Field Description                                                                                       |
 | --------- | ------------------------------ | ------------------------------------------------------------------------------------------------------- |
-| `result`  | String                         | Responds with number of successfully written telemetry data points and total number of provided points. |
+| `result`  | String                         | Responds with number of successfully written telemetry data points and total number of provided points. Format to use: `{success}/{total}`, both amounts as integers. |
```
karcass, line 212: How about adding `successes` and `total` as integers, in addition to a `result` string, or dropping `result` entirely?

Retzoh:
@karcass @schnuerle I just updated the PR, you can review it

PR #463, opened 2020-03-04
Added Louisville link to open data

schnuerle:
# MDS Pull Request

## Explain pull request

Adding Louisville, KY as another city example of how to process MDS data into open data using mostly MySQL code.  This is transparency around our process that can be replicated in other cities.  Kansas City, MS is using this code now, for example.

## Is this a breaking change

Only a ReadMe change.

* No, not breaking

## Impacted Spec

* `provider` (simply uses provider, does not impact)

## Additional context

Includes code and detailed methodolgy.


schnuerle:
From an account that is no longer active so CLA can't be signed.  Will add same info to a new PR.

PR #468, opened 2020-03-13
Define "beta" features and replace the optional/required designation for endpoints

jfh01:
## Explain pull request

This pull request defines the idea of a "beta" feature and replaces the optional/required designation (#456) on the `provider` API endpoints with a beta yes/no designation.

This is a proposed change to the 0.4.1 release candidate that emerged from review by the OMF Technology Council. The concern identified by the Council is that "optional/required" does not clearly communicate if, when, and how an endpoint should be used. For example, is it OK for a regulator to ask a provider to implement an optional endpoint? For what purposes is it appropriate to use an optional endpoint?

The new "beta" language attempts to remove this ambiguity by honestly communicating what users can expect in terms of the completeness, quality, and stability of newly added features. Users can then make an informed decision about when and how to use beta features.

## Is this a breaking change

* No, not breaking 

## Impacted Spec

Which spec(s) will this pull request impact?

* `provider`

Issue #455, opened 2020-02-26
Use a generic versioning header


Retzoh:
### Is your feature request related to a problem? Please describe.

In #444 we moved the versioning spec to one file, used by all the APIs.
To avoid breaking changes in the Provider API, the `Accept-content` header was kept as `application/vnd.mds.provider+json`.

### Describe the solution you'd like

With the 0.5.0 milestone, we can remove the `.provider` part of the header since it is now available for all the APIs.

The header would become `application/vnd.mds+json`.

### Is this a breaking change

A breaking change would require consumers or implementors of the API to modify their code for it to continue to function (ex: renaming of a required field or the change in data type of an existing field). A non-breaking change would allow existing code to continue to function (ex: addition of an optional field or the creation of a new optional endpoint). 

 * Yes, breaking

### Impacted Spec

For which spec is this feature being requested?

 * `agency`
 * `policy`
 * `provider`

### Describe alternatives you've considered

Leave the header as it is.

### Additional context

None

karcass:
Sounds good to me

schnuerle:
It looks like this change is now in `dev`, correct @Retzoh @karcass?  Can we close this issue now?

karcass:
I believe so.

thekaveman:
[Yes it is in there](https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/general-information.md#versioning).

Closing.

Issue #447, opened 2020-02-11
Policy spec doesn't state whether or not clients should specify an MDS version through the Accept header


jrheard:
### Describe the problem

According to https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/policy#rest-endpoints ,

> Responses must set the Content-Type header, as specified in the [Provider versioning](https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/provider/README.md#versioning) section.

This seems pretty clear: I take it to mean that responses to requests for Policy endpoints must include a `Content-Type` header that indicates the version of MDS associated with the response.

What's not clear to me is: does the **rest** of the Provider spec's versioning section's behavior need to be implemented as well for Policy endpoints, or is this responses-must-set-a-`Content-Type`-header portion the **only** http-header-related behavior that's required in Policy-land?

---

Specifically, I'm curious about whether or not the Policy endpoints should implement these versioning-related behaviors described in the Provider spec:
* "Clients must specify the version they are targeting through the Accept header."
* "If an unsupported or invalid version is requested, the API must respond with a status of `406 Not Acceptable`."
* "A client negotiates available versions using the `OPTIONS` method to an MDS endpoint."

---

Also, the Provider spec's versioning section includes this line:

> Since versioning was not added until the 0.3.0 release, if the Content-Type header is not set as specified above, version 0.2 must be assumed.

Does some form of this line (presumably with a different default version number) apply to the Policy spec?

### Describe the solution you'd like

I think that the Policy spec should be modified to answer these questions.

If the `Content-Type` response header is the **only** versioning-header-related behavior required by the Policy spec, then the spec should be modified to explicitly say so.

If the Policy endpoints must implement the full set of behaviors described in https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/provider/README.md#versioning , then the spec should say so, and should be clear about any Policy-specific changes that might be necessary (eg using 0.4 as the default MDS version instead of 0.2, etc).

### Is this a breaking change

A breaking change would require consumers or implementors of the API to modify their code for it to continue to function (ex: renaming of a required field or the change in data type of an existing field). A non-breaking change would allow existing code to continue to function (ex: addition of an optional field or the creation of a new optional endpoint). 

* I'm not sure

### Impacted Spec

For which spec is this feature being requested?

 * `policy`

Retzoh:
Hi, this is something I have already started tracking in #422.
[...]

PR #473, opened 2020-04-07
rename general-information.md

sarob:
It was discussed in a past wg-ops meeting to change this file to be topic specific rather general

## Is this a breaking change

* No, not breaking

## Impacted Spec

N/A

## Additional context

documentation change

### DIFF
--- a/api-versioning.md
+++ b/api-versioning.md
@@ -1,12 +1,4 @@
-# Mobility Data Specification: **General information**
-
-This document contains specifications that are shared between the various MDS APIs such as `agency`, `provider` and `policy`.
-
-## Table of Contents
-
-* [Versioning](#versioning)
-
-## Versioning
+# API Versioning
 
 MDS APIs must handle requests for specific versions of the specification from clients.
 


thekaveman:
As a slight aside, I couldn't find the minutes for the wg-ops discussion for this change, am I missing something? Would you mind giving a little more context around the discussion?

The idea for the `general-information.md` file came out of discussion in a number of issues/PRs and calls, see: #422, #444, #447, #455. I'm wondering if the smaller group in wg-ops gives enough chance for public discussion to decisions like these? In particular, I feel there should at least be an Issue before a PR, to have a chance for broader input / async input for those that can't participate in calls (I would say the same for agency/policy/provider changes as well).

jfh01:
This PR conflicts with #468, which extends `general_information.md` beyond just versioning. Agree with Kegan that this needs more discussion if we're going to pursue this change.

sarob:
This must have come from a separate conversation concerning governance documentation since I cannot find the reference in the wg-ops meeting notes either. Definitely was not meant to be an end run around the working groups.

schnuerle:
Closed, no longer relevant.

PR #482, opened 2020-04-17
Rearrange the Policy API's speed limit example

jrheard:
Rearrange the Policy API's speed-limit examples from most specific to least specific, per the API's ["order of operations"](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/policy#order-of-operations) system.

## Explain pull request

Right now, these two examples are in (least specific, most specific) order, which should mean that the "Venice Beach on weekend afternoons" Rule **never** applies to any vehicle, which seems like probably a mistake!

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`


Retzoh:
@karcass can you take a look at it and see if we should include wording clarifications to the order of operation spec?

Issue #472, opened 2020-04-02
Policy: Add fees


brianellin:
### Is your feature request related to a problem? Please describe.

Many policies that govern mobility programs are related to fees. For example, a city may have a $0.25 per-trip fee, fees related to particular kinds of usage, or fees related to vehicle permitting.

The problem: MDS Policy currently enables agencies to communicate policy rules to providers and update those rules over time, however does not provide a way to describe fees and fees associated specific policies.

### Describe the solution you'd like

A solution that's able to accurately represent real-world agency fees in the `policy` API. While there are many ways fees could be represented, here's an initial proposal for how it could implemented with a small number of non-breaking changes:

**Additions to Rule Types**

Add the following Rule Type: 

`fee`: Fees based on regions. Rule amount refers to the fee amount in the currency defined in the Policy fee_currency.

**Additions to the Rule object**

Add the following new fields on `Rule` to support the new `fee` Rule Type:

- `amount` *(Integer, optional)*: The amount of the fee in the currency defined in the Policy fee_currency

**Additions to the Policy object**

Add the following field on `Policy`:

- `fee_currency` *(String, optional)*: An ISO 4217 Alphabetic Currency Code representing the currency of the fee, as defined in the provider/trips endpoint.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

For which spec is this feature being requested?

 * `policy`

### Describe alternatives you've considered

None at this time.

### Additional context

Example uses of fee type rules:

- A city-wide trip surcharge of $.25 cents applied when a trip starts

      "name": "City Wide Trip Surcharge",
      "rule_type": "fee",
      "amount": 25,
      "geography": CITY_WIDE_GEOGRAPHY,
      "statuses": {
            "reserved": []
          }

- A right-of-way surcharge that is charged when a vehicle is deployed in a given area – $.25 cents downtown and $.5 cents in a historically underserved neighborhood

      "name": "Downtown Right of Way Surcharge",
      "rule_type": "fee",
      "amount": 25,
      "geography": DOWNTOWN_GEOGRAPHY,
      "statuses": {
            "available": ["service_start"]
          }

karcass:
I have a different design in mind, which is to put fines/fees/etc. in a separate API called Enforcement.  I will do a writeup.

quicklywilliam:
Based on feedback from the City Services call today, we will follow up with a PR with detailed changes and additional real-world examples from fee rules in use in cities today.

[...]

PR #484, opened 2020-04-22
Policy: add "rates" (fees + subsidies)

quicklywilliam:
## Context for this PR
Many policies that govern mobility programs are related to fees. These fees can be simple or complex, including geographic or time-based restrictions. Note that we distinguish **fees** from **fines**, which are punitive in nature and are not addressed with this proposal.

While most cities are using MDS data to assess fees, the fees themselves are not currently represented in MDS. `Policy` enables agencies to communicate policy `rules` to providers and update those rules over time, but it does not yet provide a way to describe fees. This proposal adds the ability to fully represent real-world agency fees in the policy API. It supports three types of fees based on the fee's `recurrence`:

- Fees charged `once` per event. For example, per-trip "Street Use fees" are charged for each trip. Cities using these types of fees include Austin, Hoboken, Oakland, Portland and others.
- Fees charged once for `each_time_unit`.  For example, "Right of Way fees" that are charged once for each day a vehicle is deployed. Cities using these types of fees include Portland and Santa Monica.
- Fees that are charged `per_complete_time_unit`. For example, "Metered fees" that are charged for each hour a vehicle is parked. One city using this type of fee for micromobility is Charlotte. These types of fees may be more common for moped, car-sharing etc.

Lastly, this proposal generalizes from describing fees to describing `rates`- which includes fees but also subsidies. As micromobility evolves, particularly under COVID, subsidies are a growing area of discussion and are addressed in this proposal simply by allowing the `rate_amount` to be a negative.

## Is this a breaking change
No.

## Impacted Spec

`policy`


quicklywilliam:
This pull request addresses #472

Retzoh:
Mike McDonald, Charles Noling, @karcass please add comments about your use-cases, concerns and remarks.

mike-mcdonald:
## Need for fees in MDS
:thumbsup:

We (Portland Bureau of Transportation) currently charge shared electric scooter operators a daily fee for parking in different areas of the city to provide some relative incentive for not placing all shared electric scooters in what is essentially our downtown area. That use case is fully addresed by this pull request. As a larger variety of vehicles use MDS, we want to express our parking regulations and policies such as parking meter zones and other curb regulations. The policy API and this pull request seem to be able to express those regulations to a high degree.
### Creating parking garage functionality
My comment in the city services working group call last week (April 30, 2020) discussed creating more complex pricing structures than what is documented here. Thinking over what I want to express some more, I think what I was trying to say amounts to expressing parking garages, at least a concept of them, and their fee structures within the policy API. The [current rates for SmartPark garages](https://www.portlandoregon.gov/transportation/article/254736) in Portland don't fall neatly within the examples here, although they are close. Going further, I think that our curb space could be analagous to a parking garage for scooters, where some curbs are preferred and follow a different fee structure than other less preferred curbs, and having a way to express more complex fee structures may be a way to promote good behaviors, especially when used to subsidise or waive parking in preferred areas.

schnuerle:
@mike-mcdonald It would be great to have this PR address your use cases for the 1.0.0 release if the changes are minor.  Or at least the next release.  Can you explain more what it would take to align the proposal to your needs?  Would you use it in its current form?

mike-mcdonald:
@schnuerle For most if not all of our current usage fees for e-scooters, we can use this addition.

schnuerle:
Some discussion on the call today about this.  Agreed it's useful and close to being ready as is.

Discussed the possibility of flagging this as a [beta feature](https://github.com/openmobilityfoundation/mobility-data-specification/blob/master/general-information.md#beta-features), or adding a sentence to clarify that an agreement on usage of rates should happen between cities and providers prior to using and relying on it.  

We will discuss at our next [Provider WG](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/MDS-Provider-Services-Working-Group) call.

schnuerle:
Final discussion on this is during this week's City Services call.

@quicklywilliam Can you add some language to clarify that an agreement on usage of rates should happen between cities and providers prior to them using and relying on it? 

Also can you flag this as a beta feature, and link to that description with "Beta feature: Yes (as of 1.0.0)".   See the [vehicles endpoint](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/provider#vehicles) for an example.

policy/README.md
```diff
@@ -238,6 +242,7 @@ An individual `Rule` object is defined by the following fields:
 | `count` | Fleet counts based on regions. Rule `max`/`min` refers to number of devices.                                  |
 | `time`  | Individual limitations on time spent in one or more vehicle-states. Rule `max`/`min` refers to increments of time in [Rule Units](#rule-units). |
 | `speed` | Global or local speed limits. Rule `max`/`min` refers to speed in [Rule Units](#rule-units).                  |
+| `rate`  | **[Beta feature](/general-information.md#beta-features):** Yes (as of 1.0.0). Fees or subsidies based on regions and time spent in one or more vehicle-states. Rule `rate_amount` refers to the rate charged according to the [Rate Recurrence](#rate_recurrence). Agencies and Providers must agree on terms of use prior to utilizing the `rate` rule type.     |
```
jfh01, line 245:
```suggestion
| `rate`  | **[Beta feature](/general-information.md#beta-features):** Yes (as of 1.0.0). Fees or subsidies based on regions and time spent in one or more vehicle-states. Rule `rate_amount` refers to the rate charged according to the [Rate Recurrence](#rate_recurrence). As this is a beta feature, agencies must communicate with providers about how the `rate` rule type will be used prior to implementation.     |
```

jfh01:
Suggested some changes to the language regarding communication between agencies and providers on the use of rates here:

https://github.com/openmobilityfoundation/mobility-data-specification/pull/484/files#r445699437

Would love to get input from cities and providers on the specific verbiage.  

@mike-mcdonald @bhandzo @dirkdk

schnuerle:
@quicklywilliam Can you rebase your branch against the OMF `dev` branch since we've pushed a lot of changes this week for 1.0.0 and there are now conflicts?  If you can do this by Monday that would be ideal since that's the day we are making the release candidate for approval.

policy/README.md
bhandzo: I prefer the earlier language, but maybe agree gives providers more say than cities are willing to give. If "communicate" was changed to "consult" that makes it feel more like the needs of providers should be considered when implementing.

jfh01: "Consult" sounds fine to me. We'll highlight this language for the Board too when they do their review of the release.

Issue #496, opened 2020-05-16
Add Policy Author API


janedotx:
### Is your feature request related to a problem? Please describe.

One pain point for cities is managing TNCs and micromobility providers. For example, some places have been inundated with scooters cluttering up the right of way, or traveling at speeds that are unsafe for nearby pedestrians. We have the `Policy` spec to define machine-readable policies that can be communicated to TNCs and micromobility providers, but since it's a read-only set of endpoints, we'd like to propose another spec that can manage writing, editing, and deletion.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

This would complement the Policy spec, but without affecting it.

### Describe alternatives you've considered

It's possible for agencies to continue to serve Policies as flat files, but in case they wanted to be able to implement some kind of CRUD app to manage policy instead of say, hand-editing JSON files themselves.

janedotx:
See https://github.com/openmobilityfoundation/mobility-data-specification/pull/497

schnuerle:
See the notes from the [last WG call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.01.07-(Joint-Working-Group)#agenda).

- Way to update policies - create/add, draft, and deactivate policy
- Providers could get early access to review, heads up beforehand, part of conversation (not written that way now)
- Adds scope and access control levels to changes
- Maybe useful, as is doesn’t have connection between providers and agencies
- Where should this live, in its own MDS Admin new repo? WGSC conversation.
- Sherie Tan San Jose - wants it to be fair to vendors, could be helpful to have a testing area for providers. 
- Explore it's use on the vendor side. Dirk says it could be useful to them

PR #497, opened 2020-05-16
Add Policy Author API spec

janedotx:
See https://github.com/openmobilityfoundation/mobility-data-specification/issues/496


schnuerle:
See the notes from the [last WG call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.01.07-(Joint-Working-Group)#agenda).

- Way to update policies - create/add, draft, and deactivate policy
- Providers could get early access to review, heads up beforehand, part of conversation (not written that way now)
- Adds scope and access control levels to changes
- Maybe useful, as is doesn’t have connection between providers and agencies
- Where should this live, in its own MDS Admin new repo? WGSC conversation.
- Sherie Tan San Jose - wants it to be fair to vendors, could be helpful to have a testing area for providers. 
- Explore it's use on the vendor side. Dirk says it could be useful to them

Issue #480, opened 2020-04-15
Policy-Driven Trip Events


quicklywilliam:
### Is your feature request related to a problem? Please describe.

Cities want to be able to perform Policy compliance monitoring (both real-time and historical) for vehicles in motion. Currently, it is not possible to meet this need without requiring precise trip location data.

### Describe the solution you'd like

The idea is that vehicles would emit events whenever they enter or leave a geography defined in the Policy endpoint. Then, instead of sharing the exact vehicle location, these events would simply give the UUID for the geography. This would be accomplished via the following changes:

**Rename events for entering and leaving municipal areas so they apply to all Policy geographies.**

- `enter_municipal_area` ⇒ `enter_geography`
- `leave_municipal_area` ⇒ `leave_geography`

**Add the geography information to the event.**

- Add a new `event_geography` field containing the geography UUID to `Provider`. This field is required if `event_location` is not present.
- Add a new `event_geography` field containing the geography UUID to `Agency`. This field is required if `telemetry` is not present.

**Remove requirement for precise location data for vehicles during trips.**

- Make `event_location` in Provider optional in the case of `vehicle_state:in_use` events
- Make `telemetry` in agency optional in the case of `vehicle_state:in_use` events

### Is this a breaking change
 * Yes, breaking

### Impacted Spec

For which spec is this feature being requested?

 * `agency`
 * `provider`

### Describe alternatives you've considered

None at this time.

### Additional context

Note that this proposal assumes/builds on the already proposed changes that unify and simplify Policy and Agency events: 
https://github.com/openmobilityfoundation/mobility-data-specification/issues/271

This issue is also related to a proposal to unify the way municipal boundaries and Policy geometries are expressed:
https://github.com/openmobilityfoundation/mobility-data-specification/issues/474

Retzoh:
Discussed during WG call:

- How to manage the entrance into several geographies at once?
- Is it possible to leverage this idea to let planning team identify the most used routes in the city?
- Could events be sent in real time with only geography, and then updated later with exact location once real-time privacy concerns are over?

johnclary:
At the City of Austin, we have been hesitant to adopt the policy API because of the required exchange of telemetry data. This approach would provide a needed and welcome alternative. I'm offering our strong support for this change.

karcass:
I like this idea!  The draft of the PR for #271 is ready for an initial review too:

https://github.com/lacuna-tech/mobility-data-specification/pull/7

[...]

PR #499, opened 2020-05-16
Extract Geography API from the Policy API

janedotx:
Originally the Policy API included the Geography API, because Policy needed Geography objects.
As other proposed services (e.g. Jurisdiction API) are introduced, Geography really should be its own thing.


schnuerle:
From the [Working Group call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.09.10-(Provider-Services-WG)) today.

- **/geographies use cases**. Is it needed?
   - Caching - so providers and third parties can download it all at once.
   - Searching - so you can see all geographies at once and how they relate before asking for details on the one you need from another API.
   - Adding geoJSON fields would make it too unwieldy. 
   - Agreed for now it's useful and keep as is.
- **Previous Geographies**
   - this is written now so these never become 'inactive', they just may not be referenced any more.
   - Warrants a new section that elaborates on the nuance and implementation here.
   - Need to check on city capacity to implement this and update
      - noted that Geography is already in Policy, so if a city is implementing Policy, they can handle the new Geography endpoint.
      - complexity comes from the prev_geographies - could be hard to keep things up to date
      - WGSC and OMF can do some outreach to cities and Policy Webinar attendees to ask about challenges
      - third parties can alleviate these concerns. Eg, in Louisville one employee just managed existing GIS files and one spreadsheet, and Stae took care of translating and publishing to Policy API.
- **Type of Geography** field
   - Currently is meant to just be the definition of an area to be used elsewhere
   - Use can always be found elsewhere in other APIs
   - Use cases could be to pass in a variable and return just those on /geographies, find specific geographies like city boundaries or parking spots, Metrics to get relevant areas
   - Discussion if it exists to make an enumerated list of value, or maybe some kind of tagging system, or freeform.
   - For now leave it out, but could be added later if other endpoints need it
- **Public feed**
   - Could be made public (non-authenticated) since there is no sensitive info here
   - Along with Policy, could be very useful for third party apps, new operators in a city
   - Needs to be accurate and stay up to date

geography/README.md
```diff
@@ -0,0 +1,150 @@
...
+            // GeoJSON 1
+        },
+        {
+            // GeoJSON 2
+        }
+    ]
+}
+```
+
+_Note: A simple tool to validate `geographies.json` will be contributed to the Open Mobility Foundation._
```
schnuerle, line 78: Should this note be included now? How close to being ready is this tool? It may take some time to contribute to a tool to the OMF.

geography/README.md
```diff
@@ -0,0 +1,150 @@
+# Mobility Data Specification: Geography
+
+This specification contains a collection of RESTful APIs used to read Geographies (descriptions of geographical information, e.g. multi-polygons, currently represented via GeoJSON).
```
schnuerle, line 3: Are geographies always single or multi polygons?  They will never be a line or point right?  Either way we should clarify this somewhere in the spec.  For an example, #480 assumes only polygons.

karcass: `FeatureCollection` is the highest-level concept.  You can include lines and points but they are no-ops because they can't contain points.  We will revisit once we have a curb spec.

geography/README.md
karcass: We should discuss.  I don't have a good answer.  Perhaps it's trivial now that @thekaveman made a schema?

schnuerle: I think just cut it for now so we can get this into the feature branch in OMF.  The schema helps, and I'm not sure what this simple tool would be otherwise.

README.md
```diff
@@ -31,6 +31,8 @@ The Mobility Data Specification (**MDS**), a project of the [Open Mobility Found
 
 * The [`policy`][policy] API endpoints are intended to be implemented by regulatory agencies and consumed by mobility providers. Providers query the Policy API to get information about local rules that may affect the operation of their mobility service or which may be used to determine compliance. It was first released in October 2019. Development takes place under the guidance of the OMF's City Services Working Group.
 
+* The [`geography`][geography] API endpoints are intended to be implemented by regulatory agencies and consumed by mobility providers. Providers query the Policy API to get information about geographical regions for regulatory and other purposes. It was first released in October 2019, originally included as part of the Policy specification. Development takes place under the guidance of the OMF's City Services Working Group.
```
jschirrmacher, line 34: With the new endpoint, the title line should now be "MDS is currently comprised of **four** distinct components"

karcass: Yes.

schnuerle:
Merging this to the feature branch so we can look at it holistically, make edits and suggestions, and incorporate other related issues/PRs before we agree on the spec and merge to dev.

schnuerle:
PR #499 is merged to a 'feature branch' in the MDS repo so we can see how it fits into the larger spec, allow the community and staff to do PRs against it, merge other related PRs into it, and see how it will fit into dev when it's ready with PR #582.  

See the '[feature-geography](https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-geography/geography/README.md)' branch for details.

PR #503, opened 2020-05-21
Geography-Driven Events

quicklywilliam:


quicklywilliam:
This is an initial attempt at implementing #480. Note that because #271 has not landed yet I had to be a bit interpretive in how to work things into the current event structure. Happy to clean things up as that issue moves to implementation.

agency/README.md
```diff
@@ -173,8 +173,9 @@ Body Params:
 | ----------- | ----------------------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
 | `event_type` | Enum                         | Required          | see [Vehicle Events](#vehicle-events)                                                                                                           |
 | `event_type_reason` | Enum                  | Required if Available | see [Vehicle Events](#vehicle-events)                                                                                                           |
-| `timestamp`  | Timestamp                    | Required          | Date of last event update                                                     |
-| `telemetry`  | [Telemetry](#telemetry-data) | Required          | Single point of telemetry                             |
+| `timestamp`  | Timestamp                    | Optional          | Date of last event update                                                     |
```
billdirks, line 176: I believe `timestamp` is suppose to be `Required` and `telemetry` is suppose to be `Optional`?

provider/README.md
```diff
@@ -341,6 +342,8 @@ Without an `event_time` query parameter, `/status_changes` shall return a `400 B
 | | | `maintenance_drop_off` | Device introduced into service after being removed for maintenance |
 | | | `agency_drop_off` | The administrative agency (ie, DOT) drops a device into the PROW using an admin code or similar |
 | `reserved` | A customer reserves a device (even if trip has not started yet) | `user_pick_up` | Customer reserves device. Note that this event_type is deemed ambiguous right now: it can mean reserved as in possession, or reserved as in future hold. This issue will be remedied in a future version with an extra event_type. For now, different providers have implemented this in either way. |
+| | | `enter_geography ` | Device enters one or more Geographies managed by Policy during an active trip. |
```
billdirks, line 345: nit: There's an extra space after enter_geography and leave_geography.

jfh01:
How do folks feel about trying to get this in for 1.1.0? @quicklywilliam @schnuerle @billdirks

schnuerle:
Feel good about it for 1.1. Milestone marked.

karcass:
Probably want to spell out some use cases for `trip_enter_geography` and `trip_leave_geography`.  E.g. if you have, say, council districts, would transitioning from one to another generate two simultaneous events (leave, enter)?  Would leaving the municipal boundary potentially generate two events (trip_leave, trip_leave_geography)?

quicklywilliam:
@karcass these are interesting questions. I'm wondering if there might be a way that we could unify the events, so that there's just a single `trip_leave` event issued (with both the jurisdiction and policy geos in `event_geographies`) instead of a `trip_leave_jurisdiction` and a `trip_leave_geography`.

As for uses cases, one that has came up when we started talking to cities about this was for assessing fees. For example, suppose a city wants to assess a 25 cent fee for trips entering or starting downtown during rush hour.

provider/README.md
```diff
@@ -260,7 +260,8 @@ Unless stated otherwise by the municipality, this endpoint must return only thos
 | `event_types` | Enum[] | Required | [Vehicle event(s)][vehicle-events] for state change, allowable values determined by `vehicle_state` |
 | `event_time` | [timestamp][ts] | Required | Date/time that event occurred at. See [Event Times][event-times] |
 | `publication_time` | [timestamp][ts] | Optional | Date/time that event became available through the status changes endpoint |
-| `event_location` | GeoJSON [Point Feature][geo] | Required | See also [Stop-based Geographic Data][stop-based-geo] |
+| `event_location` | GeoJSON [Point Feature][geo] | Optional | See also [Stop-based Geographic Data][stop-based-geo] |
+| `event_geographies` | UUID[] | Required | Array of Geography UUIDs consisting of every Geography that contains the location of the status change. Required if `event_location` is not present. |
```
karcass, line 264: For symmetry I would propose that we modify `trips` to _either_ send the route (FeatureCollection with lat/lon) _or_ an array of the Properties fields.  The Properties currently contains timestamp (and optionally stop_id) so you'd also add `list-of-geographies` as the "telemetry" points, dropping `geometry`.  My read of the GeoJSON spec is that Features without `geometry` are not supported, so this would be a not-strictly-GeoJSON object.  

Example:

```js
"route": [
    {
      "timestamp": 1529968782421,
      // Required for Trips starting at a Stop
      "stop_id": "95084833-6a3f-4770-9919-de1ab4b8989b",
      "geography_ids": [
        "4ad32463-2c14-4979-9e7f-ec0f21a643e3",
        "99fc6c35-394c-4651-bfb7-9883d3c96d4a"
      ]
    },
    {
      "timestamp": 1531007628377,
      // Required for Trips ending at a Stop
      "stop_id": "b813cde2-a41c-4ae3-b409-72ff221e003d",
      "geography_ids": ["439ec866-ad7b-468f-969b-6d7e912ee5a5"]
    }
  ]
```

karcass (commented):
Added some suggestions about not conflating telemetry (movement of vehicle in same state) with events (change of vehicle state).  

Also we will need to be super clear on how Geographies are communicated, such that the Providers' notions of currently relevant Geographies is in sync with the regulating Agency.  Geographies themselves do not have start-dates or end-dates; applicability is for Policies or Jurisdictions or Metrics to specify.   Solving that problem cleanly blocks this proposal IMHO.

agency/README.md
```diff
@@ -183,7 +183,8 @@ Body Params:
 | `vehicle_state` | Enum                          | Required | see [Vehicle States][vehicle-states] |
 | `event_types`   | Enum[]                        | Required | see [Vehicle Events][vehicle-events] |
 | `timestamp`     | [timestamp][ts]                     | Required | Date of last event update |
-| `telemetry`     | [Telemetry](#telemetry-data)  | Required | Single point of telemetry |
+| `telemetry`     | [Telemetry](#telemetry-data)  | Optional | Single point of telemetry. |
+| `event_geographies`  | UUID[] | Required          | Array of Geography UUIDs consisting of every Geography that contains the location of the event. Required if `telemetry` is not present.|
```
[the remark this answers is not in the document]
quicklywilliam: This seems logical to me. As an aside, I'm having fun imagining deanonymization attack techniques to derive a route from speed data alone.

general-information.md
```diff
@@ -283,6 +283,8 @@ Vehicles can enter the `unknown` state to and from any other state with the foll
 | `unknown` | `on_trip`   | `comms_restored`        | The vehicle transmitted status information after a period of being out of communication. |
 | `unknown` | `on_trip`   | `unspecified`        | The provider cannot definitively state how a vehicle started a trip. |
 | `on_trip` | `elsewhere`   | `trip_leave_jurisdiction` | A vehicle on a trip left the jurisdiction |
+| `on_trip` | `on_trip `   | `trip_enter_geography` | The vehicle enters one or more Geographies managed by Policy during an active trip. |
+| `on_trip` | `on_trip `   | `trip_leave_geography` | The vehicle leaves one or more Geographies managed by Policy during an active trip. |
```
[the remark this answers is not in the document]
quicklywilliam: It seems like we need to define an event type in order to effect a status event for crossing geometry boundaries, right?

But I see your point about how these new events I've added are different because they represent internal transitions. What do you think about having a separate table of internal transition events? I'm imagining there may be other possible useful internal events we might some day consider:
- a stationary vehicle updates its location due to acquisition of a better GPS fix
- a parked vehicle is picked up and carried to a new location
- a parked vehicle is charged
- an in motion vehicle experiences a possible crash event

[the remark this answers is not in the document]
karcass: What I propose is that we make Policy responsive to telemetry as well as events.  We need that anyway to enforce `speed` rules.  Let me put together a specific (hypothetical) example that expresses your "any trip that intersects downtown" use-case.

provider/README.md
quicklywilliam: I agree that there is a symmetric change to make in the Trips field. We started fleshing one out, but we decided not to include it in this PR for two reasons
- The use cases around the trips endpoint are very different
- There are some subtleties in the changes required, and these depend on the use cases. For example, one possible use case might be assessing speed limit data. It's possible to imagine adding a max speed to the properties list for this purpose, but this seems very specific and in want of some kind of generalization where the required data was specified at the Policy level. 

What do you think about starting a new issue and having a conversation around the use cases for this change?

general-information.md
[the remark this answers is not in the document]
quicklywilliam: I'm unclear on how that would work for MDS Provider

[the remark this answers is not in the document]
karcass: See below, it would require using Trip data.

provider/README.md
karcass: I'm not sure how to separate them.  Do you want to put this on the agenda for the WG meeting this coming Thursday?  Or maybe a quickie zoom session beforehand?

I appreciate wanting to keep the PR surface area low.

general-information.md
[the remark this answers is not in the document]
quicklywilliam: Ah, now I understand the thinking – thanks! I think this makes logical sense and is a good change regardless of whether we add internal state transitions.

The downside here is that it means increased latency – geography transitions would only be reported when the trip is complete. One of the goals of this PR is to allow cities to make realtime Policy decisions.

The other argument I'd make is that we are really just generalizing the existing pattern of firing events when municipal boundaries are crossed. As we know municipal boundaries can be quite complex and multi-layered. The way I think about is simple: _we communicate an event when a vehicle passes into an area where different rules apply_. That could be a Policy region under the same jurisdiction, a new jurisdiction – or potentially both.

schnuerle:
We reviewed this PR as part of the second OMF Working Group Steering Committee release [Checkpoint](https://github.com/openmobilityfoundation/governance/blob/main/technical/ReleaseGuidelines.md#checkpoints).  Both WGSCs had some feedback and I'm documenting it here for discussion.  

Note: we definitely want to flag this as a 'beta' feature in the release, so leaving a reminder here.  

We were trying to think how geographies can best be aligned to this, how do you know which geography is used for an event?

And how can this idea be addressed in the simplest way initially, with minimal changes or dependencies on other things like Jurisdictions.

Idea 1)  Maybe just use all geographies and return data for all of them by default, then the city can decide which geographies are of value.  As part of it being beta, recommend cities talking to providers first to provide a list of geographies it applies to.   

Idea 2) There could be a new rule type in policy. One that is 'event' and you would define the geography and details there.  This may be a bit duplicative, since you would also use that geography in another rule type.

Idea 3) Events could be an optional flag in policy that says that this policy needs to be treated as an event. 

What do you all think of these ideas? I like Idea 3, but I'm not sure it meets all of your use cases @quicklywilliam.

quicklywilliam:
Thanks @schnuerle, and sorry we missed the last meeting – had our team offsite that day and I didn't realize this was part of the agenda!

Idea 1 makes sense to me, though I'm not sure it solves the right problems. As I see it, the problems are with Operators getting behind on Geometries - issuing events for old Geometries or failing to issue events for new Geometries. Idea 1 seems like it would be more helpful in a case where Operators are ahead – issuing events for Geometries that the city doesn't know about yet, which doesn't seem like it should be possible.

To me the issue here is that the spec technically allows a new Geometry to be published with an effective_date of now – or even in the past. This makes it infeasible or impossible for a provider to issue events for this Geometry without backfilling events, which many MDS provider implementations cannot do. A few options here:

1. Require event backfilling for MDS provider implementations (at the risk of operators not implementing this anytime soon)
2. Change spec to require that the effective_date be sufficiently far from the initial publishing date to allow providers a reasonable amount of time (say, 2 hours) to poll the Geometries endpoint and begin producing events.
3. Leave effective_date alone and instead document that Events for new Geometries must begin being issued within 2 hours (or whatever) of the publish_date or the effective_date, whichever is later.
4. Introduce a new property on Rules called 'compliance_required_by_date', which would allow cities to specify at the Policy level how long Operators have to begin issue events and complying. Perhaps it could default to the 2 hours.

Of these, I like 4 for being the most flexible and explicit. It seems valuable apart from Policy-driven events for a City to be able to set a compliance_required_by_date (and thereby establish an effective 'grace period' after a Policy's effective_date), and it seems reasonable to insist that this date be the same date at which operators are required to start issuing events (ie they need to know about the Policy to comply with it, and they should be ready to issue events by the time they know about it).

As for Ideas 2 and 3, I think maybe they are similar in intent to what I just suggested? Not sure I am understanding them correctly though. Could you explain? In particular, not understanding how these two ideas help us ensure that Providers are issuing up-to-date events for new Geometries?

provider/README.md
quicklywilliam: Good call, def think it would be easier to flesh this out on Zoom. I will follow up via email.

quicklywilliam:
Following up on my last message regarding stale Geometries. @karcass @avatarneil and I caught up on this and agree that the best thing for this PR is to simply require that effective_date to be at least X (say 10 minutes, or an hour) after publish_date. This is just to give providers a generous amount of time to poll the endpoint and start generating events.

We're planning to meet again to resolve the other outstanding question around policy driven events and state transitions.

schnuerle:
Note that for 1.1.0 we have [merged](https://github.com/openmobilityfoundation/mobility-data-specification/pull/582) the new [Geography API](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/geography) to the 'dev' branch. Please update this pull request with the latest code, resolve any conflicts, and make references to the Geography API where appropriate, e.g. with UUIDs.

schnuerle:
The primary topic for this week's Working Group call is Policy Driven Events.

https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.10.22-(Joint-Working-Group-Provider-Services)

@quicklywilliam Do you think you could incorporate the Geography changes from dev into this by then?  It would be good to add any other finishing touches to the PR, and have some discussion topics you need some community feedback on.

general-information.md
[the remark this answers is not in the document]
quicklywilliam: Spoke about this today, we've decided to go ahead with the proposal to include these new event types. There is some consideration of collapsing them into one event instead of two, that's a detail something we hope to discuss and resolve at tomorrow's WG.

agency/README.md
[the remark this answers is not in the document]
quicklywilliam: We decided to go forward with this as an event. Here's a relevant issue on that for the trips endpoint – it seems like we should have an effort in the future to add the richness of telemetry data to endpoints throughout Provider.

provider/README.md
quicklywilliam: With the decision to create new event types, the work to add geographies data to the trips endpoint be captured via a future issue, if it is needed.

quicklywilliam:
All set! Have also updated the agenda with details of what needs to be covered.

schnuerle:
From the Working Group call last week, here is a summary of the discussions and action items that should be done in the next week or two to make this release:

https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.10.22-(Joint-Working-Group-Provider-Services)#minutes

- Rename the proposal to ‘Geography Driven Events’ (GDE) to reflect that for now the actions are triggered when vehicles enter or leave a geometry, and geometry IDs are returned.
- If you use GDEs, then cities are required to make policy and geography public and communicate the URLs to provide public transparency in the policy and geographic areas, and reduce authentication burden on providers. Add that language.
- Ensure this is structured to be a non-breaking change for 1.1.0, and so it can be tested with minimal disruptions.
- Agreed that 2 fields (enter and leave) are better than one (cross), so leave as is.
- Added note that new policies can’t be expected to be implemented until 20 minutes after publishing to prevent frequent or real time changes.  20 min delay is good for some more emergency policy situations, but most policy will be communicated days or weeks ahead of time. Ensure this is clear.
- Add the beta label and link in the appropriate place to allow cities and providers to try it out and give feedback.
- Beta designation will also also providers to see what their technical burden is to implement and run in production
- Language needed to clarify that policy driven events only apply when you opt into this and discuss with provider and city.
- Can be run in parallel with all existing MDS if desired by cities. So you can turn on GDEs and other telemetry, or turn on GDEs and turn off telemetry, or just use telemetry as it is in 1.0.0. Ensure this is clear.
- Could be issues with GPS in precision float and boundary riding, returning too many events. May need smoothing of data. Could be worth adding some language to spec to account for this.

schnuerle:
There's about a week left to get these changes into the PR for this release @quicklywilliam. If you can do that we can save some time on the last WG call Nov 12, though that day is filling up with some other items now like Metrics and k-values, so Nov 5 may be better (though it's of course much sooner!).

quicklywilliam:
@schnuerle I'll post an update on this shortly so that we can discuss and resolve most, if not all, of the remaining points.

quicklywilliam:
Here's what has been resolved:
- [x]  mark as beta
- [x]  clarify which geometries are communicated in in the case of an enter/leave. event_geographies
- [x]  add language clarifying that event_location is required if geometries not present
- [x]  add language around events being only required for beta users
- [x]  change title to "geography driven events"

And here's the last bit remaining:
- [ ]  add some overall language somewhere describing intent/purpose of "geography driven events"

quicklywilliam:
One thing I want to call attention to is that in resolving the issues above I decided to eliminate the `trip_leave_geography` event in favor of a single `crossed_geography_boundary`. This was something we discussed at the last work group. We saw benefits to both approaches, but no strong preference emerged.

However, I came to a new conclusion in considering how to best clarify which geometries are communicated with these events. The inclusion of the leave event was actually adding complexity here. For every event besides leave, `event_geographies` simply consists of every Geography that contains the location of the event. It is implied that the location given is of the vehicle in its new state – not the one it has just transitioned from. However, for a leave event it is unclear – it would seem that event_geographies should contain the geographies the vehicle just left. Removing this event resolves this ambiguity and makes the behavior of `event_geographies` consistent. If needed, the list of geographies left can be inferred.

karcass:
Name seems a little wordy.  :)
How about `changed_geographies`?

schnuerle:
William this looks great, thanks for incorporating all the feedback!

Will you be available to present on your changes at tomorrow's meeting?  No need to do an overview of the whole GDE proposal, but just what's been updated, and the decision to merge the leave/enter geographies.

Also I noticed a few things in the PR that could be update before tomorrow. If you give me edit access ([#2 here](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md#mds-pull-request)) I can make them, otherwise here they are:

- [ ] 1. 'geometry' driven events is written in a number of places. Surely just a typo, one I've made before too!
- [ ] 2. can you add DS Store back to the .gitignore for the repo?
- [ ] 3. a new request: could you add an example of GDEs in the /examples folder readme?

policy/README.md
```diff
@@ -180,6 +181,31 @@ The optional `end_date` field applies to all policies represented in the file.
 }
 ```
 
+[Top][toc]
+
+## Geography-Driven Events **[Beta feature](/general-information.md#beta-features):** *Yes (as of 1.1.0)*
```
schnuerle, line 186: Since this info is relevant to both Agency and Provider, I think it should be moved to the general-information.md file and linked to from both APIs as needed.

quicklywilliam: Hmm, I considered this but thought that it made sense to put it in Policy for now because Geography-Driven Events is exclusive to Policy. I was imagining it could move to General Information if/when Geography-Driven Events become usable in other ways (such as with non-Policy geographies and/or Metrics).

karcass: I'm leaning General because that's where it's going to end up anyway.  IMHO.

quicklywilliam: Happy to move it if that's the consensus here. If so, I think it would probably be good to add a bit of extra language making it clear that Geography-Driven Events are currently only issued-for and used-with Geometries that are tied to one or more active Policy rules.

policy/README.md
```diff
@@ -203,7 +203,9 @@ Here's how it works in practice:
 
 
 
-Agencies that wish to use Geography-Driven Events do so by requiring a new `event_geographies` field in status events. This field acts in lieu of location or telemtry data. When an Agency is using Geography-Driven Events, Providers must emit a new `changed_geographies` status event whenever a vehicle in a trip enters or leaves a Geography managed by a Policy.
+Agencies that wish to use Geography-Driven Events do so by requiring a new `event_geographies` field in status events. When an Agency is using Geography-Driven Events, Providers must emit a new `changed_geographies` status event whenever a vehicle in a trip enters or leaves a Geography managed by a Policy.
+
+During the Beta period for this feature, location and telemtry data remain required fields. This allows Aggencies to test Geometry-Driven Events, measuring its accuracy and efficacy against regulatory systems based on precise location data. After the beta period, if Geometry-Driven Events is deemed by OMF to be accurate and effective, the specification will evolve to allow cities to use Geography-Driven Events in lieu of location or telemtry data.
```
quicklywilliam: I wrote 'Geometry' on purpose here just to maintain the long running joke. Changed it back to 'Geography' in the subsequent commit.

quicklywilliam: I for one thought this was much funnier than modifying my Git history

schnuerle (approved):
Looks good thanks for the changes based on community and city feedback. Will make other minor changes during release candidate preparation (eg like typos with 'telemtry' spelling).

Issue #491, opened 2020-05-08
Add Jurisdiction spec (Need a way to describe the attributes of a regulating agency)


janedotx:
### Is your feature request related to a problem? Please describe.

It is necessary for cities and transportation agencies to be able to designate certain geofenced areas as being legally significant. The most obvious example is delineating the boundaries of a city. We used to fulfill that need through the use of service areas, but those are being deprecated, as they mingled too many concerns together. A plain Geography is not enough, even though it is possible to annotate such an object with metadata, because the metadata can be arbitrary, and it is desirable to keep each object type as atomic as possible.

### Describe the solution you'd like

A new Jurisdiction type and service need to be defined.

### Is this a breaking change

 * No, not breaking

### Impacted Spec
This is a call for a brand new spec.

### Describe alternatives you've considered

It's possible to annotate certain Geographies using GeographyMetadata, but that gives us a less clean ontology.

janedotx:
Pull request here: https://github.com/openmobilityfoundation/mobility-data-specification/pull/492

quicklywilliam:
Can you clarify how this relates to Policy and the /geometries endpoint in particular? It's confusing to me that geometries is a part of Policy but Jurisdiction is a separate thing that references geometries.

karcass:
@quicklywilliam 

Policy and Geography are presently together in the same spec because Policy was the only service that used Geography.  We will be explicitly making Geography stand-alone in an upcoming PR.  Apologies for not doing that first.

jrheard:
It’s not clear to me why this problem is best solved with a new Jurisdiction spec. It seems to me that the same goal could be achieved by amending the Geographies spec. It's very possible that I'm just not understanding the actual problems that this issue seeks to solve, because there's some stuff (eg `effective`) in the corresponding PR that isn't described in the issue.

---

[The proposed Jurisdiction spec’s README](https://github.com/openmobilityfoundation/mobility-data-specification/blob/eafc876242639617c480c457c3f86043888ff148/jurisdiction/README.md) shows that one of the main purposes of this spec is to associate the string fields `agency_key`, `agency_name`, and `description` with a Geography.

Why not add those fields to the Geography spec, rather than making a new+separate Jurisdiction spec?

The issue says that:
> A plain Geography is not enough, even though it is possible to annotate such an object with metadata, because the metadata can be arbitrary, and it is desirable to keep each object type as atomic as possible.

[...]

PR #504, opened 2020-05-22
Provider: make Trip 'route' field optional for privacy reasons.

vperron:
**Explain pull request**
For many regulatory and privacy-oriented reasons, it's possible that cities or providers do not want external actors to access the `route` geolocalized data of the trips throughout the city.

We propose to make this `route` field optional to better reflect that choice.

The distance, cost and duration fields, all available in the general Trip payload,  still
contain valuable and aggregated information and should thus be accessible, even if the route is not.

**Is this a breaking change**

No, not breaking. A mandatory field is now optional.

**Which spec(s) will this pull request impact?**

provider

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -199,7 +199,7 @@ A trip represents a journey taken by a *mobility as a service* customer with a g
 
 The trips endpoint allows a user to query historical trip data.
 
-Unless stated otherwise by the municipality, the trips endpoint must return all trips with a `route` which [intersects](#intersection-operation) with the [municipality boundary](#municipality-boundary).
+Unless stated otherwise by the municipality, the `route` (if provided) of the trips endpoint must [intersect](#intersection-operation) with the [municipality boundary](#municipality-boundary).
 
 **Endpoint:** `/trips`  
 **Method:** `GET`  
@@ -218,10 +218,10 @@ Unless stated otherwise by the municipality, the trips endpoint must return all 
 | `trip_id` | UUID | Required | A unique ID for each trip |
 | `trip_duration` | Integer | Required | Time, in Seconds |
 | `trip_distance` | Integer | Required | Trip Distance, in Meters |
-| `route` | GeoJSON `FeatureCollection` | Required | See [Routes](#routes) detail below |
 | `accuracy` | Integer | Required | The approximate level of accuracy, in meters, of `Points` within `route` |
 | `start_time` | [timestamp][ts] | Required | |
 | `end_time` | [timestamp][ts] | Required | |
+| `route` | GeoJSON `FeatureCollection` | Optional | See [Routes](#routes) detail below |
 | `publication_time` | [timestamp][ts] | Optional | Date/time that trip became available through the trips endpoint |
 | `parking_verification_url` | String | Optional | A URL to a photo (or other evidence) of proper vehicle parking |
 | `standard_cost` | Integer | Optional | The cost, in the currency defined in `currency`, that it would cost to perform that trip in the standard operation of the System (see [Costs & Currencies](#costs--currencies)) |
@@ -247,6 +247,8 @@ For the near-ish real time use cases, please use the [events](#events) endpoint.
 To represent a route, MDS `provider` APIs must create a GeoJSON [`FeatureCollection`](https://tools.ietf.org/html/rfc7946#section-3.3), which includes every [observed point][geo] in the route, even those which occur outside the [municipality boundary](#municipality-boundary).
 
 Routes must include at least 2 points: the start point and end point. Routes must include all possible GPS or GNSS samples collected by a Provider. Providers may round the latitude and longitude to the level of precision representing the maximum accuracy of the specific measurement. For example, [a-GPS](https://en.wikipedia.org/wiki/Assisted_GPS) is accurate to 5 decimal places, [differential GPS](https://en.wikipedia.org/wiki/Differential_GPS) is generally accurate to 6 decimal places. Providers may round those readings to the appropriate number for their systems.
+
+Routes are optional, as they may contain sensitive information and local policies could forbid their availability.
 
 ```js
 "route": {


schnuerle:
About the city limit change, I think that should be removed from this and put in a different PR.  Additionally there are cases where a city needs to know about routes outside of their city limits/jurisdiction.  See Issue #491 for some discussion.

vperron:
Done with #505  @schnuerle

quicklywilliam:
We support this change. It's easy to meet perform basic tasks like getting trip counts via the trips endpoints, which is why many cities use it (the provider/trips endpoint was **the most commonly used** endpoint in the MDS Maturity survey). It would be great if we could meet such needs without requiring access to route data.

Another thing that we've noticed is that compared to status events, the trips endpoint tends to be more consistent from operator to operator in terms of the data it returns. Hence, a secondary need for a routeless /trips endpoint is to help catch and diagnose issues with status feeds.

quicklywilliam:
This change is complimentary to #480, which makes location and telemetry data optional.

thekaveman:
I think it should be a little clearer that *optional* here means it is up to the consuming city/agency. There are many consumers of Provider that continue to rely on `route` objects, including for purposes of accurate cap counting via a blending of status and trips.

Maybe clarity around optional fields is a broader need that should be addressed in the [General Information](https://github.com/openmobilityfoundation/mobility-data-specification/blob/master/general-information.md) document? Tagging @jfh01 for visibility.

schnuerle:
It seems the main issue that is trying to be addressed with this PR is that cities may not want trip line data. I’d like some cities to chime in to see if this is just a ‘possible’ scenario, or a current need.

Another way that this could be solved if needed is to create a way to make this and other fields optional via the API. An example would be the inclusion of a new parameter, say `exclude`, that could have a list of fields the city does not want to have returned in the response.  If a city didn’t want trips, then they can add `exclude=trips` to the API call.  If they don’t want trip or fee data, they could add `exclude=trips,fees`.  This list of possible options can be defined.  

This would also solve future issues like this where cities don’t want/can’t receive certain data. And has a benefit of getting only the data you need for the task at hand, and reducing the returned data file size and processing on the provider’s side.  

I propose we move this discussion over to the new #507 issue, so we can talk about solutions separate from the details of a PR.

vperron:
@thekaveman I second that !

@schnuerle great idea, I also answered on #507.

quicklywilliam:
@thekaveman makes a good point about using routes for cap counting and such. I think those needs could be met by just having start/end location but no other route data? 

Arguably sharing start/end locations would be equivalent to this PR in terms of privacy impact, because start/end data is already exposed via status changes. So, perhaps this issue could be solved while retaining the use cases @thekaveman mentions via either:
- a simple language tweak that makes it clear that sharing 2-location (start/end) routes is an acceptable practice
- start/end location fields that can be provided in lieu of the route field

thekaveman:
In my prior experience implementing cap counting in Santa Monica, having just the start/end location would not be enough. The [method we used](https://github.com/CityofSantaMonica/mds-provider-services) looks at each point of the route within city limits, and uses the earliest/latest timestamps to reconstruct the window of time the vehicle spent inside the boundary.

vperron:
@thekaveman That is very much in line with the point I'm raising, I think.

Of course, if the routes are available (beacuse the city / agency decided it), any client app may then use those to implement accurate information such as heatmaps, etc.

If they are not, advanced features may not be usable, but that would not discard all of the info provided by the `/trips` endpoint (distance, cost, vehicles involved, etc) and a simpler, best-effort version can be implemented.

Right now, apps couldn't use the endpoint at all if for any reason the routes are not made available.

@quicklywilliam start/end may indeed help, but I'm not entirely sure that every city/agency would be willing to provide them, for the same privacy reasons.

quicklywilliam:
@vperron are these cities using /trips without the /status endpoint, then?

vperron:
@quicklywilliam I don't know, that was purely hypothetical ! My point being that if privacy is the reason the routes are not provided within trips, replacing them with starting and ending points maybe does not change much that issue. 

So my first thought would be that they probably would have to be optional as well, even if I entirely agree, those start/end points  could also be found in the /status endpoint, maybe just not as easily.

So we see 4 options now:
- Do nothing: the routes are mandatory and no trip info is given without it
- Make the route optional: all the trip metadata is accessible to apps, no privacy issues with /trips
- Replace route by start/end points in some cases: better info, but maybe still has privacy issues
- Allow payloads to be parametrized: extremely flexible, but probably tough to implement for all providers

quicklywilliam:
Ah, thank you for clarifying! Apologies, I think I might have misunderstood your original use case.

Could you clarify what kinds of external actors you have in mind for this use case? If the use case involves sharing trips data widely beyond the agency, I think even removing route data might not suffice to address potential privacy issues.

vperron:
The sensitive data from the point of view of cities&agencies we've contacted is the geolocation, combined with a timestamp, of a particular vehicle, especially if it becomes easy to determine frequent routes from point A to B (commuting, for instance)

If this `route` information is stripped, the `/trips` payload still contains duration, vehicle, distance and cost information, which can be used for city-wide mobility analysis and trends but can not identify any place, portion or district specifically. 

My point is, we should consider making this basic information accessible even if the cities or agencies are not willing to expose `route` information, thus making the `route` "optional".
On the other hand, if the `route` info is present, further analysis (district-to-district balancing, etc) becomes possible.

quicklywilliam:
Got, thank you for clarifying! For this use case I think I would advocate for sharing via a different means. I am concerned that in raw format trip data can be attacked, even without route data. For example, if I know duration along with the exact times that a trip ends and begins then it is likely I can attack a GBFS feed by comparing timestamps and durations with when various vehicles disappear at a given location. From there, it is likely I could then establish an exact trip O/D for a portion of trips.

schnuerle:
Since the original use case by @vperron is about sharing the feeds with other entities, and removing routes from trips may not be enough to protect privacy, and #480 is now implemented in 1.1.0, and we are discussing this more broadly in #507, I think we should close this specific PR.

vperron:
Agreed.

schnuerle:
This has been addressed even more explicitly with #646 and the ability for cities to exclude route data from provider endpoints.

PR #505, opened 2020-05-22
provider: Fix a contradiction about the route localization

vperron:
**Explain pull request**
I understand that choosing which option is still an ongoing discussion, in  #491 for instance.

Nevertheless, for now the current specification is contradictory:
- routes must be within city limits at some point
- route may not be within city limits 10 lines further

This PR enforces the most restrictive access for now: the route should be within a city boundary, as stated higher up in the document.

**Is this a breaking change**
Not really, since the specification is currently contradictory, it can't be implemented.

**Which spec(s) will this pull request impact?**
Provider

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -244,7 +244,7 @@ For the near-ish real time use cases, please use the [events](#events) endpoint.
 
 ### Routes
 
-To represent a route, MDS `provider` APIs must create a GeoJSON [`FeatureCollection`](https://tools.ietf.org/html/rfc7946#section-3.3), which includes every [observed point][geo] in the route, even those which occur outside the [municipality boundary](#municipality-boundary).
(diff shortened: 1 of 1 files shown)


thekaveman:
It's not clear where you are reading the contradiction? Are you referring to this line:

> Unless stated otherwise by the municipality, the trips endpoint must return all trips with a route which intersects with the municipality boundary.

This line speaks to *which routes* must be included (all routes that have any intersection with the municipality boundary). The line changed in this PR speaks to *which points* must be included in a route (all points, both inside and outside the boundary).

This change does not look correct to me.

vperron:
I think you're right, the change is incorrect. Some colleagues and myself did indeed misread it. 

I'll try and add some clarification to better reflect your clarification.

schnuerle:
@vperron is there a different change you'd like to propose here, or is it OK to close this PR in favor of a future PR aligned with the latest MDS release code?

vperron:
As I stated in my latest comment this PR is actually invalid, I'll close it immediately.

PR #523, opened 2020-06-19
Policy: use a consistent definition for timestamps (recreation)

schnuerle:
Recreated the #435 PR which fixes issue #432.

-----

Aligns the Policy text specification for timestamps to the Provider and Agency
APIs, which use integer milliseconds from Unix epoch.

Fixes issue #432.

### Explain pull request

Fixes mismatch of string vs. number for timestamps.

### Is this a breaking change

Technically, yes, but in reality I'm not sure.

The `mds-core` reference implementation already defined timestamps as numbers:
https://github.com/openmobilityfoundation/mds-core/blob/09930c89ea4c71684140aa0046ba9727be40bd79/packages/mds-types/index.ts#L144

That being said, if other implementations exist that were using strings it's possible there will be breakage.

### Impacted Spec

 * `policy`

### Additional context

N/A

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -73,6 +73,10 @@ The response to a client request must include a valid HTTP status code defined i
 - **404:** Not Found: Object(s) do not exist.
 - **500:** Internal server error.
 
+#### Timestamps
+
+As with the Provider API, `timestamp` refers to integer milliseconds since Unix epoch.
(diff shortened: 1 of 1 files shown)


schnuerle:
Please review and approve for merger to dev.

PR #506, opened 2020-05-26
Provider/Agency reconciliation

karcass:
Reference documents:

1. [Github Issue](https://github.com/openmobilityfoundation/mobility-data-specification/issues/271)
2. [Spreadsheet Comparison](https://docs.google.com/spreadsheets/d/1rs6pGI3LudXIFQfZLkqE5sJ4A6_WfM1439hG4IK6B0Y/edit#gid=0)
3. [Notes](https://docs.google.com/document/d/1-KggZnPtdmiDvJJOzb4NWa8v7MfdNK88quWc0y7dzyo/edit#heading=h.j69uqwsassh7)
4. [State machine diagram](https://docs.google.com/presentation/d/1Ar2-ju8YlddSsTATvQw4YjsSa5108XtidtnJNk-UAfA/edit#slide=id.p)
5. [Presentation Slides](https://docs.google.com/presentation/d/1tBMli8dGP6d101UvYnBEqQMCDnSq5xUvBKEMLDwKjEo/edit?usp=sharing)
6. [Video of Presentation](https://us02web.zoom.us/rec/share/xpRzCJTI1npOGY3PuBn0ebEhXdTieaa82yUf8vsMyh5k7t_WQ5m_tol_pJS6bLY8)  Password: 1B&1*2E?

PR #534, opened 2020-06-26
Reconciled JSON schemas

### LINKED ISSUE #518: Add 'other' option to vehicle_type

Would like to align vehicle types between MDS and GBFS.  

Currently [GBFS v2.1rc has 5 types](https://github.com/MobilityData/gbfs/blob/master/gbfs.md#free_bike_statusjson), and [MDS 0.4.1 has 4 types](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/provider#vehicles).  

They are the same except GBFS has an `other` type.

Would like to add the `other` to vehicle_types.  It seems valuable so you don't have to make up your own type or pigeon hole into one of the 4, for example if you want to do use MDS with a new vehicle type like dockless cars, drones, etc.  Can be updated later away from `other` by providers/cities when the new vehicle type is added to the spec.

If possible, I'd like to include this in `1.0.0`.  Can be implemented in the Reconciliation PR #506.

### Is this a breaking change

 * No, non-breaking

### Impacted Spec

 * `agency`
 * `policy`
 * `provider`

Do nothing, leave the feature specified as-is.

### LINKED ISSUE #526: Update JSON Schemas for reconciled state machine
After #506 is merged, we'll have a reconciled state machine between Agency and Provider. A follow-up task is to update the JSON Schemas with the new set of `vehicle_state` and `event_type`. 

If possible, we'd also like to indicate which combinations of those values are considered valid under the state machine as is done now within the [`status_changes.json`](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/provider/dockless/status_changes.json#L324) Schema. This could be difficult because the new model allows an array of `event_type` per `vehicle_state`, and intermediate states are "hidden" by the final valid event transition. [See discussion on #271 for more details](https://github.com/openmobilityfoundation/mobility-data-specification/issues/271#issuecomment-624170186).

### LINKED ISSUE #531: Reconcile more Agency and Provider fields

This is a follow-up item from #506 and related to the work for #526.

In updating the JSON Schemas for the Reconciliation work in #506, it has become apparent that there are more areas of Agency and Provider that "almost" overlap. Since we're taking this opportunity in 1.0.0 to introduce breaking changes to the state machine, I propose the following additional breaking changes to a few additional fields. These will allow for more reuse in the schema generation process and generally further improve alignment between Agency and Provider.

#### 1. Vehicle model

Agency `GET /vehicles` and `POST /vehicles` uses the following fields to model a vehicle (omitting some for brevity):

```
provider_id
device_id
vehicle_id
type
propulsion
```

While in all of the Provider endpoints, the vehicle model looks like:

```
provider_id
device_id
vehicle_id
vehicle_type
propulsion_type
```
**Proposal:** rename Agency fields `type` and `propulsion` to match their Provider corollaries. The field `propulsion_type` will be renamed to `propulsion_types` to more clearly indicate that it allows an array of values.

#### 2. Conditional Trip references

In Agency `POST /vehicles/{device_id}/event` we have the field `trip_id`:

> `trip_id`: UUID provided by Operator to uniquely identify the trip. Required for `trip_start`, `trip_end`, `trip_enter_jurisdiction`, and `trip_leave_jurisdiction` event types

And in Provider `/status_changes` we have the field `associated_trip`:

> `associated_trip`: Trip UUID (foreign key to Trips API), required if `event_type` is `trip_start` or `trip_end`, or for any other status change event that marks the end of a trip.

**Proposal:** rename Provider `associated_trip` to `trip_id` to align with Agency, and add the event `trip_cancel` to the list of `event_types` that make this trip reference required in either case.

### Is this a breaking change

 * Yes, breaking

### Impacted Spec

 * `agency`
 * `provider`

thekaveman:
Following up with #506, this is an update to the JSON Schema generation process and schemas for the reconciled state machine.

Work is ongoing and almost there. Marking as DRAFT for now.

Fixes #518
Fixes #526
Fixes #531


thekaveman:
See [my comment](https://github.com/openmobilityfoundation/mobility-data-specification/issues/526#issuecomment-651393739) on the Issue for a trade-off I had to make in validating the vehicle state machine.

PR #539, opened 2020-06-30
Release Candidate 1.0.0 Cleanup

schnuerle:
Cleanup for the 1.0.0 release candidate. Minor edits in formatting, links, arrangement, and language to clean up the recent changes across the repository.

List of changes (to be updated as they are made):

- Move geography to the higher level general information file out of Provider
- Add 'other' vehicle option
- Update cities using MDS pull request link
- Update TOC links with more detail
- Removed beta and date from Agency header
- Removed reference to Geography API
- Made rule units required in policy
- Updated pull request guidance for conflicts

PR #540, opened 2020-06-30
make rule-units required per #472

karcass:


schnuerle:
Going to resolve this in PR #539 with some other tweaks.

PR #545, opened 2020-07-01
Update /geographies to clearly encode municipality boundary

### LINKED ISSUE #474: Policy: Formalize "Municipality Boundary" discovery
Several parts of MDS depend on a precise definition of the municipality boundary, however the spec doesn't offer guidance on *where* exactly to define or discover this important geography. 

Here's the current language from the [provider spec](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/provider#municipality-boundary):

> Municipalities requiring MDS Provider API compliance should provide an unambiguous digital source for the municipality boundary. This boundary must be used when determining which data each provider API endpoint will include.
>     
> The boundary should be defined as a polygon or collection of polygons. The file defining the boundary should be provided in Shapefile or GeoJSON format and hosted online at a published address that all providers and provider API consumers can access and download.

This can lead to inefficiencies and confusion when communicating the boundary details from agencies to providers and vice-versa. Additionally, the Policy spec states that:

> No part of the geographical data should be outside the municipality boundary.

For both those reasons, formalizing the discovery of the municipality boundary seems desirable. 

Since the MDS Policy API already does the job of communicating important geography definitions from agencies to providers, that seems like a natural home for the municipality boundary.

Below is a proposal to represent the boundary within the existing `/geographies` REST endpoint and `geographies.json` flat file:

```
{
    "version": "0.4.0",
    "updated": "1570035222868",
    "data": {
        "geographies": [
            {
                // GeoJSON Feature 1
            },
            {
                // GeoJSON Feature 2
            }
        ],
        "municipality_boundary": [
            {
                // GeoJSON Feature 1
            }
        ]
    }
}
```

An alternate approach might be to include the `municipality_boundary` in the `geographies` dictionary directly. Would welcome other ideas/suggestions for how to formally define.

### Is this a breaking change
 * No, not breaking

### Impacted Spec

 * `policy`

n/a

### Additional context
n/a

jrheard:
Fixes #474 .

This is my first nontrivial PR for this repository, so please let me know if I've missed anything!

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -113,7 +113,7 @@ Policies will be returned in order of effective date (see schema below), with pa
 
 Endpoint: `/geographies/{id}`  
 Method: `GET`  
-`data` Payload: `{ geographies: [] }`, an array of GeoJSON `Feature` objects.
+`data` Payload: `{ geographies: geojson_feature_array, municipality_boundary: geography_id }`, where `geojson_feature_array` is an array of GeoJSON `Feature` objects and `geography_id` is the ID of the Geography that represents the [Municipality Boundary](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/provider#municipality-boundary).
 
 ##### Query Parameters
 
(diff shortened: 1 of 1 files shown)


schnuerle:
I'm not sure if this is the best way to do this, adding the municipality ID to geometry.  It seems we may want to instead add a geography ID to the proposed Jurisdictions API?

Also note we are synthesizing Geography work so I'm going to point this PR to the new geography feature branch.  This will move it out of Policy exclusively so you will have to change this PR to align to the new top-level Geography endpoint.

schnuerle:
Note that for 1.1.0 we have [merged](https://github.com/openmobilityfoundation/mobility-data-specification/pull/582) the new [Geography API](https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/geography) to the 'dev' branch. Please update this pull request with the latest code, resolve any conflicts, and make references to the Geography API where appropriate, e.g. with UUIDs.

jrheard:
I'm just going to close this PR, I think it's generally outdated/obsolete. Time flies!

PR #544, opened 2020-07-01
Release Candidate 1.0.0

schnuerle:
### Summary

The Release Candidate for MDS `1.0.0` has been submitted: <https://github.com/openmobilityfoundation/mobility-data-specification/releases/tag/1.0.0-rc1>

Please use this pull request to track Technology Council and OMF Board feedback and/or requested changes.

### Action Item

This pull request will be merged by an OMF maintainer after OMF Board Approval following the [Release Guidelines](https://github.com/openmobilityfoundation/governance/blob/main/technical/ReleaseGuidelines.md#making-a-release).

PR #546, opened 2020-07-06
Adding Provider mention in Readme

schnuerle:
I'd like to add a mention of the number of providers that use MDS (based on the MDS providers.csv file) in the main readme under the 'Cities That Use MDS" section, and linking over to the root [providers.csv file](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/providers.csv) for more details and a full list.

Instead of listing them out I think the link works.  Though maybe providers could opt in or make a PR to get themselves listed here?

Also thought about this for third parties.  If this idea is good we could explore in another pull request.


schnuerle:
I added this simple update to the #544 PR.  If people would like to have providers opt in and be enumerated with links in this section, please open an issue, discuss on calls or forums.  If we do have links to providers' websites it would be great to also link to FAQ or docs online on their websites that talk about the use of MDS.  Same with third parties.

PR #553, opened 2020-07-13
Add Circ to Providers.csv

bhandzo:
@schnuerle Discovered that Circ never got a provider id, so adding one for them.


schnuerle:
Added this as part of the 1.0.0 release candidate with PR #544.

PR #560, opened 2020-07-31
Add 'unspecified' events to some schema states

thekaveman:
Also clean up a few duplicate table entries.


schnuerle:
Thanks for the schema update, and removing the duplication.

PR #563, opened 2020-08-18
Clarify trips and status changes response behavior

jiffyclub:
In MDS 0.4.x we request trips and status changes for a given hour. A provider may have no data to report for the requested hour for multiple reasons, including:

- The hour has not yet fully passed
- The provider was not operational in the region during the hour
- The provider has not yet completed processing that hour's data
- No trips or events occurred during the hour

Differentiating the last two reasons is particularly important because it lets consumers of APIs know whether or not they need to re-request the hour at a later time. A successful 200 response with an empty list of data clearly communicates that the provider has nothing to report for that hour and that it's not necessary to try again later. Having this behavior consistent across providers would help us at Populus spend less time requesting data and reduce the number of requests received by providers.


schnuerle:
This seems like a good idea and change. Would like some community feedback on it.

Retzoh:
👍 to this. As an agency, it is a real problem to request a time range and not know if we should move on because there are no events there, or retry later because there is an error / the data is incomplete.

jiffyclub:
Bumping this to get it included in the 1.1.0 release.

schnuerle:
Thanks for the bump this is looking good for 1.1.0 unless there is some opposition to it. Tagging some provider members just in case @bhandzo @dirkdk.

thekaveman:
Updated target branch to `dev`.

schnuerle (approved):

PR #576, opened 2020-09-14
Policy JSON Schema

### LINKED ISSUE #389: JSON Schemas for the Policy Endpoints
**Describe the bug**
not including JSON Schema in #382, so we can iterate faster on the draft during the 0.4.1 cycle. at some point during 0.4.1, need to have that created. 

**To Reproduce**

**Expected behavior**
we generate the JSON schemas for policy by running `python schema/generate_schemas.py`. 

**Additional context**

thekaveman:
JSON Schema describing MDS 1.0 Policy.

The examples have been pulled out into distinct files in a subdirectory and updated to pass schema validation.

Closes #389.

### DIFF
--- a/agency/get_stops.json
+++ b/agency/get_stops.json
@@ -111,11 +111,8 @@
         },
         "devices": {
           "$id": "#/definitions/stop/properties/devices",
-          "type": "array",
-          "description": "List of device_id for vehicles currently at this Stop.",
-          "items": {
-            "$ref": "#/definitions/uuid"
-          }
+          "$ref": "#/definitions/uuid_array",
+          "description": "List of device_id for vehicles currently at this Stop."
         }
       }
     },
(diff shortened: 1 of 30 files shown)


schnuerle:
Beyond all the detailed work on the Policy schema and examples, it also looks like you have some critical updates to the Policy readme. It seems like it would be good to get this into 1.0.0, especially since the schema applies to 1.0.0 and it may change a lot in 1.1.0 based on what's being worked on.

Do you think this should be in the 1.0.0 release, possibly approved this week? Tagging @jfh01 @karcass @hunterowens in case they want to weigh in.

thekaveman:
> Do you think this should be in the 1.0.0 release, possibly approved this week?

That was my hope. The (current) examples for Policy aren't correct under 1.0, so it would great to get this update pushed in time.

schnuerle:
I am planning on merging this to the 1.0.0 release tomorrow.  

The new items are defining a schema for Policy, something that's been missing and known to be needed since its creation.  The changed items are critical updates to links and the spec that were either missing or incorrect.  

If we need further changes or updates to this we can open a new PR and address them in the next release or with a hotfix.

PR #581, opened 2020-09-29
Add optional geography types (per #580)

karcass:
## Explain pull request

See #580  

## Is this a breaking change

* No

## Impacted Spec

Which spec(s) will this pull request impact?

* `geography`

## Additional context

None at moment

Issue #580, opened 2020-09-29
Add optional geography_type field to Geography spec


karcass:
### Is your feature request related to a problem? Please describe.

Per 1.1 Working Group meetings, the idea of clarifying Geography objects via free-form types in an optional `geography_type` was raised as a useful non-breaking addition.

### Describe the solution you'd like

Add an optional string field to the Geography object definition

### Is this a breaking change

 * No, not breaking

### Impacted Spec

For which spec is this feature being requested?

 * `geography`

### Describe alternatives you've considered

We thought about not adding it ... another proposal was to add a list of geography_types when a geography could plausibly be more than one type of area, but so far nobody has identified clear and useful examples of this.

schnuerle:
Added in #582

PR #582, opened 2020-09-30
Geography - Feature Branch

schnuerle:
## Explain pull request

This PR is to see what changes between the Geography feature branch and the dev branch, and prepare it for merger once it's ready.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `agency`
* `policy`
* `provider`

## Additional context

PR #499 is merged to a 'feature branch' in the MDS repo so we can see how it fits into the larger spec, allow the community and staff to do PRs against it, merge other related PRs into it, and see how it will fit into dev when it's ready with this PR.


geography/README.md
```diff
@@ -0,0 +1,148 @@
...
+- 403 - user is attempting to read an unpublished geography, but only has the `geographies:read:published` scope.
+
+**Endpoint**:  `/geographies`
+**Method**: `GET`
+
+Path Params: 
+
+| Name         | Type      | Required/Optional | Description                                    |
+| ------------ | --------- | --- | ---------------------------------------------- |
+| `summary`    | string    | Optional   | Return geographies, minus the GeoJSON in each geography object     |
```
janedotx, line 127: Bit of a nit, but I feel it could be helpful to show an example of a summarized `Geography`.

janedotx (changes requested):

geography/README.md
schnuerle: I'd love to see some examples, maybe in an examples folder like we have with [Policy](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy/examples), or maybe just an examples.md file. Do you want to create that?

schnuerle:
Note: something missing from this is a schema.  We can work on that at a later date.

geography/README.md
```diff
@@ -0,0 +1,148 @@
...
+### Geography Fields
+
+| Name               | Type      | Required/Optional | Description                                                                         |
+| ----------------   | --------- | --- | ----------------------------------------------------------------------------------- |
+| `name`             | String    | Required   | Name of geography                                                                      |
+| `description`      | String    | Optional   | Detailed description of geography                                                                      |
+| `geography_id`     | UUID      | Required   | Unique ID of geography                                                                 |
+| `geography_json`   | UUID      | Required   | The GeoJSON that defines the geographical coordinates.
+| `effective_date`   | [timestamp][ts] | Optional   | The date at which a Geography is considered "live".  Must be at or after `publish_date`.
+| `publish_date`     | [timestamp][ts] | Required   | Timestamp that the policy was published, i.e. made immutable                                             |
```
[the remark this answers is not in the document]
schnuerle: Thanks for the catch, I made the update.

geography/README.md
schnuerle: I created the start of an examples.md file for others to add on to. 

https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-geography/geography/examples.md

geography/README.md
```diff
@@ -0,0 +1,158 @@
+# Mobility Data Specification: Geography
+
+This specification contains a collection of RESTful APIs used to read Geographies (descriptions of geographical information, e.g. multi-polygons, currently represented via GeoJSON).
+
+Geographical data has many applications in the context of mobility, such as the description of municipal boundaries, locations for pick-up and drop-off zones, and areas of temporary closure for special events or emergencies.  This API is intended to support a variety of other APIs, including the Policy API.
+
+Geographical data will be stored as GeoJSON and read from either `geographies.json` or the `/geographies` endpoint, referenced by UUID. Geography data once published through this API shall be treated as immutable, to ensure that any rules or regulations referring to the boundaries cannot be retroactively changed.  A Geography may be deprecated and replaced by updated version with a new UUID.
```
schnuerle, line 7: I wonder if this should be /geography and geography.json, instead of /geographies? The other APIs are singular, and match their API name, and we call this the Geography API. Also changing the `/geography/{geography_id}` endpoint.

schnuerle: Consensus on call today to keep the endpoint as /geographies

geography/README.md
schnuerle: Will move this into a folder like in Policy and will add some real geographic data for some concrete examples and visuals of the GeoJSON.

schnuerle:
Added geography type manually from #580 and #581.

schnuerle:
Added 5 examples and associated files.  Each example has 2 files, one is the MDS JSON, and the other is the GeoJSON part of those files so that Github can make a preview of it in the browser.  

I pulled them all from a real world city, Louisville, KY because 1) we needed a real city, 2) all of these GIS files are documented and available online as open data [here](https://data.louisvilleky.gov/dataset/dockless-vehicles#data-and-resources) and [here](https://data.lojic.org/search?q=dockless), 3) many different geography_types are represented in the available files, and 4) I am familiar with this city and these files.

I'd like to add one more example file which is the location of Stops when it's available from the city online.

schnuerle:
I've added an [example for 'stop'](https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-geography/geography/examples/README.md#stop) which uses Point instead of MultiPolygon.  Are points allowed, or do all geographies have to be MultiPolygons?  Points would seem to affect the ability for some policy actions and policy driven events if used.

geography/examples/README.md
```diff
@@ -0,0 +1,196 @@
...
+- [Stop](#stop)
+
+## All Geographies
+
+All of the examples below will be presented within the single [/geographies](/geography#geographies) endpoint.  
+
+**File with full geometry**: [`geographies.json`](geographies.json)
+
+**GeoJSON sample to view all geometries together in a preview**: [`geographies.geojson`](geographies.geojson)
+Note this file is just for your reference to generate a preview on Githug. It is not part of the spec.
```
thekaveman, line 22:
```diff
-Githug
+Github
```

geography/README.md
```diff
@@ -0,0 +1,254 @@
...
+
+[Top][toc]
+
+### Previous Geographies
+
+Obsoleting or otherwise changing a geography is accomplished by publishing a new geography with the `prev_geographies` field, which is a list of UUID references to the geography or geographies superseded by the new geography. The previous geographies are also published in the `/geographies` endpoint.  Using it allows agencies to look back historically at previously published geographies, for analysis, historic reference, or an auditable change trail.
+
+This field is optional can be omitted by the publishing Agency.  
+
+### Geography Type
```
thekaveman, line 117: No specific suggestions here, but some of these Geography types feel like Policies more than descriptions of physical space. Do we want to mix those concerns?

thekaveman (changes requested):
Fix minor typo.

geography/examples/README.md
schnuerle: That was maybe subconscious.🤗

geography/README.md
schnuerle: Do you mean for Previous Geographies, or Geography Type?  For the types I think they can align and overlap with policies, but not always (eg census).  Many of them are in the examples as geographies that cities define, so we want to call them out and name them here.

thekaveman: For the Types. Trying to think from an implementation perspective, how does a geography with type `no_ride_zone` or `slow_ride_zone` differ from the equivalent Policy document? If an agency publishes a Geography without a corresponding Policy, are companies expected to follow the type of the Geography as an implicit "rule"?

schnuerle: Ah I could see the confusion. Maybe we need some language to clarify that the types do not equal policy, but instead are for organizational and discovery purposes in the Geography API.  They need to be linked to policy or other areas of MDS to be meaningfully used.

thekaveman: Yeah that could work. Another idea is to collapse those down to `policy_zone` or similar, to make it explicit that an associated Policy is needed. Then leave the details to the Policy/Rules. Would also make reusing the Geography across multiple (perhaps totally different) Policies more natural.

schnuerle: "Geography types do not equal policy, but instead are for organizational and discovery purposes within the standalone Geography API. They need to be linked to policy or other areas of MDS to be meaningfully used."

thekaveman: But that isn't necessarily the case for all Geography types. The Agency may wish to publish their official boundary under `/geographies`, not to tie it to Policy but to have a consistent and known publishing mechanism. I think part of the reason this API was moved out from Policy is to not tie it so explicitly to Policy, right?

schnuerle: I think clarifying language like this would work. Geographies could be used in more areas than just policy, and I think that geographies defined for some initial purpose align with how cities work, even if those areas can be used for multiple different things later.

thekaveman (approved):

PR #590, opened 2020-10-23
Add retire_date as an optional field to Geography

alexdemisch:
## Explain pull request

Add a retire_date field to allow cities to specify when a geography is no longer active.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `geography`

## Additional context

As discussed in the [10/20 Provider WG meeting](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.10.08-(Joint-Working-Group-Provider-Services)#minutes), adding a retire_date field will allow cities to indicate when a geography isn't used anymore.  While the prev_geographies field allows this functionality, it requires cities to maintain that relationship, which they may not do.  Additionally, some geographies might be considered inactive without a new geography to replace them (e.g., a special event zone).

### DIFF
--- a/geography/README.md
+++ b/geography/README.md
@@ -105,6 +105,7 @@ Placeholder -- link to schema to be added later.
 | `effective_date`   | [timestamp][ts] | Optional   | The date at which a Geography is considered "live".  Must be at or after `publish_date`. |
 | `publish_date`     | [timestamp][ts] | Required   | Time that the geography was published, i.e. made immutable                       |
 | `prev_geographies` | UUID[]    | Optional   | Unique IDs of prior geographies replaced by this one                                   |
+| `retire_date` | [timestamp][ts] | Optional | The date at which a Geography is no longer considered "live".  Must be after `effective_date`. |
 
 [Top][toc]
 


thekaveman:
I like the idea of an additional signal for stale geographies, like we talked about on the previous call.

The pedant in me needs to point out that this somewhat conflicts with our current language / understanding of geographies as immutable. For example, we say for `publish_date`:

> Time that the geography was published, i.e. made immutable

Since `retire_date` may not be known at the initial publish of a geography, is there a subsequent publish/update implied when that date becomes known? If so, are we concerned with this breaking immutability of the geography?

alexdemisch:
That's a good point @thekaveman.  I could see knowing the retire_date in advance for some geographies, but there would undoubtedly be some where the agency would not know until it happens, and this would require updating the record.  I can't think an easy way to get around this.  We could maybe keep the spatial object separate from its attributes somehow, so that the spatial object is immutable but some or all of the attributes like the retire_date can be modified, but that might be overkill and might not still not achieve immutability.  Anyone have any ideas on this?  I'm personally not concerned with this one field breaking immutability if we provide specific instructions that only this field can be updated.

schnuerle:
I attempted to add this to this [branch](https://github.com/openmobilityfoundation/mobility-data-specification/blob/ms-geography-2/geography/README.md#geography-fields) for discussion on today's call, with some language about how it's allowed to be added without affecting immutability.

schnuerle:
This [commit](https://github.com/openmobilityfoundation/mobility-data-specification/pull/604/commits/b9f5f0419c4b2f1dd82b94f2517b2ac5c370ddff) in the branch made for these and other changes as been updated per the latest WG call.  This PR can be closed in favor of the changes in that one. Thanks!

Issue #555, opened 2020-07-16
Add Stop image to Policy 


schnuerle:
### Is your feature request related to a problem? Please describe.

As parking locations are created or designated by cities, providers and riders need more context to use them properly.  This would add an optional image path URL to the Policy API for Stops.  

### Describe the solution you'd like

This photo or illustration created by the city would be sent along in the Policy API for providers' information and could be communicated visually to riders what a Stop looks like, aiding in parking.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

For which spec is this feature being requested?

 * `policy`

### Describe alternatives you've considered

Something similar could be added to Stops in Agency or Provider, but it's not clear to me what that could be.

### Additional context

Note that Stops do not yet exist in Policy, so this would be included as part of the development of Issue #554.

schnuerle:
We will be discussing this tomorrow at our Joint Working Group meeting, so attend if you are interested.  

https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.09.17-(Joint-Working-Group-City-Services)

schnuerle:
**From the WG call yesterday.**

Could be added as an optional field in Stops.
Should be an easy thing to add. 
Could just be added as part of #554.

karcass:
What form(s) would it take?  URL?  Base-64 encoded with mime type?  Other?

Retzoh:
I think an url is best, to avoid having to fetch hundreds of megabytes of data when you are just looking for the list of existing stops and their location.

schnuerle:
I will make a PR for this since it's pretty straightforward and there seems to be agreement on it.

schnuerle:
Added the optional image URL to this [branch](https://github.com/openmobilityfoundation/mobility-data-specification/blob/ms-stops/general-information.md#stops) and soon to be PR.

Issue #554, opened 2020-07-16
Add Stops across all of MDS


schnuerle:
### Is your feature request related to a problem? Please describe.

Stops has been added to Agency and Provider as a new beta feature in 1.0.0, but is not a part of Policy yet.

### Describe the solution you'd like

Add a way to specify Stops so they can be used by Policy, so that cities can geographically define and communicate details around docking stations, charing stations, parking locations, corrals, etc.  

### Is this a breaking change

 * No, not breaking - would be an addition to the spec

### Impacted Spec

For which spec is this feature being requested?

 * `policy`

### Describe alternatives you've considered

N/A

### Additional context

This is a requested addition that was not ready for the 1.0.0 release.

schnuerle:
Tony Cruz from Long Beach CA mentioned on the Provider call today a use case related to this.  The city defines preferred parking areas for events (temporary mobility hubs) and emails them to each provider a few days before the events.  Thinks defining them with GeoJSON would work for the city.  

If Stops are added in Policy these can be defined here.  Since Policy is currently not meant to be real time, he could then mass email the providers to say that the Policy feed has been updated, and they can ingest the parking locations digitally instead.

avatarneil:
The most straight-forward way I can think about implementing this would be just allowing an array of `stop_ids`. Additionally, depending on the work in https://github.com/openmobilityfoundation/mobility-data-specification/issues/549 we may want to allow policies to specify particular Spaces within a Stop too; that could be represented by having a key/val store as opposed to just an array, so like `{ stops: { stop_id: space_ids } }` (e.g. `{ stops: { 68bbdc6e-8666-416e-88cb-cabbbe5a2085: ['4e8bb731-ebb5-4244-adc0-ad4d09aa0b03', '39f603a2-ee13-4d99-a96d-67d38f392659'] } }` or something.

schnuerle:
We will be discussing this tomorrow at our Joint Working Group meeting, so attend if you are interested.  

https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.09.17-(Joint-Working-Group-City-Services)

schnuerle:
**From the WG call yesterday.**

Vianova, Louisville, and LADOT have existing use cases for this.

[...]

PR #603, opened 2020-11-19
Stops Updates for Release

schnuerle:
Consolidated updates to stops from the community across a few issues and discussions:

- Stops in policy - Add optional stops to rules (like geographies). If stops then geography ID required. #554
- Required stops example addition using current policy rules #554
- Optional stops image #555

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `agency`
* `policy`
* `provider`

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -248,7 +248,6 @@ An individual `Rule` object is defined by the following fields:
 | `days`             | day[]                       | Optional   | Days `["sun", "mon", "tue", "wed", "thu", "fri", "sat"]` when the rule is in effect (default all) |
 | `messages`         | `{ String:String }`         | Optional   | Message to rider user, if desired, in various languages, keyed by language tag (see [Messages](#messages)) |
 | `value_url`        | URL                         | Optional   | URL to an API endpoint that can provide dynamic information for the measured value (see [Value URL](#value-url)) |
-| `stops`      | UUID[]                      | Optional   | List of [Stop](/general-information.md#stops) UUIDs that this Rule applies to. If this field is used then the `geographies` field must be used to identify the locations.  |
 
 [Top][toc]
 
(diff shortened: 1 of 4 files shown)


policy/README.md
```diff
@@ -248,6 +248,7 @@ An individual `Rule` object is defined by the following fields:
 | `days`             | day[]                       | Optional   | Days `["sun", "mon", "tue", "wed", "thu", "fri", "sat"]` when the rule is in effect (default all) |
 | `messages`         | `{ String:String }`         | Optional   | Message to rider user, if desired, in various languages, keyed by language tag (see [Messages](#messages)) |
 | `value_url`        | URL                         | Optional   | URL to an API endpoint that can provide dynamic information for the measured value (see [Value URL](#value-url)) |
+| `stops`      | UUID[]                      | Optional   | List of [Stop](/general-information.md#stops) UUIDs that this Rule applies to. If this field is used then the `geographies` field must be used to identify the locations.  |
```
billdirks, line 251: We talked at the working group meeting today about the potential for errors with having stops and geographies both point to geographical entities. We also discussed if there would be other geographical entities and if that would lead to a proliferation of fields. One proposed solutions is to replace `stops` and `geographies` with a field called `places` which is a map where the keys indicate another API and the values are a list of UUIDS. So for example:
```
{ 'geographies': [UUIDS], 'stops': [UUIDS]' }
```
This would be breaking if we replaced `geographies`. We could add this field instead of `stops` and incorporate 'stops', and maybe 'geographies', as allowed keys here. If we go this route we will want to update the description of the geographies field to indicate that this change is happening.

thekaveman: +1 to adding top-level `places` now instead of `stops`, and duplicating the data from `geographies`. This seems like an OK non-breaking approach which can be further cleaned up in 2.0 timeframe.

schnuerle: I think that for this release we should just pull out this `stops` option, work out the best solution to the larger idea in the next release, and not have to duplicate geographies data now.  

For naming, `places` is now used within Stops, eg, a place to park a vehicle within a Stop area. Naming this places I think would be confusing.  Maybe `area`?

schnuerle:
@thekaveman or @jfh01 can you review and approve changes please?

Issue #569, opened 2020-08-27
Special groups aggregate metrics


dirkdk:
### Is your feature request related to a problem? Please describe.

Cities often request data on how many lower income users we reach with our vehicle service, and how many trips such users take. MDS does not currently support user segmentation. We would oppose attaching any user data to the Trips endpoint as it would involve information close to Personally Identifiable Information (PII) and would make it fairly easy to identify individuals by trip route and user segment. Instead, we propose the  solution of providing aggregate trip data by user segment.

### Describe the solution you'd like

We propose that Providers provide aggregated data on trips by special user segments, using the new [Metrics API](https://github.com/openmobilityfoundation/mobility-data-specification/pull/486). By using aggregation, it should be impossible to trace back this data to individuals. This does mean we need to set meaningful minimals for certain metrics so that the aggregated data has k-anonymity.

The new Metrics API specifies parameters for name, since, interval, dimensions that we will assume these metrics support.

### Proposed metrics for special groups:

|Metric | Description|
|-- | --|
| special_groups | array of names of special groups as served by Provider. This list will be fairly static. Example values are low_income or students |
| active_users[special_group] | total number of users for given group with at least 1 trip in interval |
| trips[special_group] | count of trips by users of given special group during interval |
| trips.average_duration[special_group] | average duration in seconds of trips by users of given special group during interval |
| trips.median_duration[special_group] | median duration in seconds of trips by users of given special group during interval |
| trips.std_duration[special_group] | standard deviation of duration  in seconds of trips by users of given special group during interval |
| trips.average_distance[special_group] | average distance in meters of trips by users of given special group during interval |
| trips.median_distance[special_group] | median distance in meters of trips by users of given special group during interval |
| trips.std_distance[special_group] | standard deviation of distance in meters of trips by users of given special group during interval |

### Overall aggregate statistics
For overall usage we can do what is listed below. Please note that this data should be derivable from MDS trip data, minus active users.

| Metric | Description |
| -- | -- |
| active_users | total number of all users with at least 1 trip in interval |
| trips | count of trips by all users during interval |
| trips.average_duration | average duration in seconds of trips by all users  group during interval |
| trips.median_duration | median duration in seconds of trips by all users  during interval |
| trips.std_duration | standard deviation of duration in seconds of trips by all users  during interval |
| trips.average_distance | average distance in meters of trips by all users  during interval |
| trips.median_distance | median distance in meters of trips by all users  during interval |
[...]

PR #606, opened 2020-12-10
Provider Reports - API

schnuerle:
## Explain pull request

This is a simplified, beta implementation of the #569 special groups aggregate metrics issue.  It adds a /reports endpoint within the Provider API that handles returning monthly trip counts of special groups, in this case just low income groups.

The structure is similar to the Metrics API, but reduced in scope (no duration specifications, no measures besides counts of trips).  

## Is this a breaking change

* No, not breaking (new feature)

## Impacted Spec

* `provider`

## Additional context

Measures and durations could be added back into the PR. See PR #607 for a static version of this solution.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -23,7 +23,7 @@ This specification contains a data standard for *mobility as a service* provider
 * [Reports](#reports)
   * [Reports - Response](#reports---response)
   * [Reports - Response Schema](#reports---response-schema)
-  * [Reports - Query Parameters](#reports---query-parameters)
+  * [Reports - Query Parameters](#reports---parameters)
   * [Reports - Dimensions](#reports---dimensions)
   * [Data Redaction](#data-redaction)
 * [Realtime Data](#realtime-data)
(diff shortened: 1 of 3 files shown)


provider/README.md
```diff
@@ -291,6 +297,162 @@ Without an `event_time` query parameter, `/status_changes` shall return a `400 B
...
+## Reports
+
+Reports are information that providers can send back to agencies containing aggregated data that is not contained within other MDS endpoints, like counts of special groups.
+
+The authenticated reports endpoint allows a user to pass in some parameters and get trip counts in the response.
+
+### Reports - Response
+
+**Endpoint:** `/reports`  
+**Method:** `GET`
```
thekaveman, line 309: Method is `GET` here, but down below the example shows `POST`.

Any reason this needs to be `POST`?

schnuerle: No this probably should get Get. The POST is a left over.

schnuerle:
Created a static version of this solution with PR #607 based on feedback from [WG call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.12.10-(Joint-Working-Group)#minutes).

schnuerle:
Closing this in favor of the static version in #607.

PR #630, opened 2021-03-04
Align propulsion_types in provider/status_changes spec and .ReadMe

ezmckinn:
---
name: Emmett McKinney
about: Suggest changes to MDS
title: Aligning "propulsion_types" in provider/status_changes spec with ReadMe.

---

## Explain pull request

The [ReadMe](https://github.com/openmobilityfoundation/mobility-data-specification/tree/release-1.0.0) for provider/status_changes.json 1.0.0 and the [spec](https://github.com/openmobilityfoundation/mobility-data-specification/blob/release-1.0.0/provider/status_changes.json) appear to conflict on the deprecated 'propulsion_type' field.   We suggest renaming the `propulsion_type` object to `propulsion_types` (i.e., adding an 's'), replacing the `$id` reference with an enum[], and removing the existing `propulsion_types` object.

## Is this a breaking change

* Yes, breaking

## Impacted Spec

* `agency`
* `provider`

## Additional context

The competing fields begin at Line 127 of provider/status_changes.json and Line 138, respectively. This caused some confusion for us while validating our Provider feeds.

### DIFF
--- a/provider/status_changes.json
+++ b/provider/status_changes.json
@@ -124,25 +124,16 @@
         }
       }
     },
-    "propulsion_type": {
-      "$id": "#/definitions/propulsion_type",
+    "propulsion_types": {
+      "$id": "#/definitions/propulsion_types",
       "type": "string",
-      "description": "The type of propulsion",
+      "description": "Array of propulsion types, allowing multiple values",
       "enum": [
         "combustion",
         "electric",
         "electric_assist",
         "human"
-      ]
-    },
-    "propulsion_types": {
-      "$id": "#/definitions/propulsion_types",
-      "type": "array",
-      "description": "Array of propulsion types, allowing multiple values",
-      "items": {
-        "$id": "#/definitions/propulsion_types/items",
-        "$ref": "#/definitions/propulsion_type"
-      },
+      ],
       "uniqueItems": true
     },
     "vehicle_type": {
--- a/provider/status_changes.json
+++ b/provider/status_changes.json
@@ -126,7 +126,7 @@
     },
     "propulsion_types": {
       "$id": "#/definitions/propulsion_types",
-      "type": "string",
+      "type": "array",
       "description": "Array of propulsion types, allowing multiple values",
       "enum": [
         "combustion",


provider/status_changes.json
```diff
@@ -124,25 +124,16 @@
         }
       }
     },
-    "propulsion_type": {
-      "$id": "#/definitions/propulsion_type",
+    "propulsion_types": {
+      "$id": "#/definitions/propulsion_types",
       "type": "string",
```
natealmeida, line 129:
```suggestion
      "type": "array",
```

thekaveman:
Thank you for the feedback @ezmckinn. A few issues with the proposal:

First and foremost, any changes to the JSON Schemas need to originate in the templates and/or generator code, which are found under the [`schema/`](https://github.com/openmobilityfoundation/mobility-data-specification/tree/1.0.0/schema) directory. The schema files sitting in top-level `agency/`, `provider/` etc. directories are generated files and should not be edited directly. 

These proposed changes were made against the common _definitions_; the actual `status_changes` schema is a few more lines down and (correctly) references the `propulsion_types` definition, matching the README: [here in the list of required properties](https://github.com/openmobilityfoundation/mobility-data-specification/blob/1.0.0/provider/status_changes.json#L246) and [here in the properties](https://github.com/openmobilityfoundation/mobility-data-specification/blob/1.0.0/provider/status_changes.json#L278).

We split "item" and "array" definitions in a few other places, including UUID/UUID array, `vehicle_type`/`vehicle_types`, `vehicle_event`/`vehicle_events` ... again these are all in the _definitions_, to be composed by the schema generator code into the actual schemas in various ways.

Can you give a little more context about the trouble you were having validating your Provider feeds? We may need to update these definitions! Some clarity around the issues would help us target the right place to make the updates (templates, generators, etc.)

schnuerle:
Since this looks to be breaking, and requires some more clarification and discussion, I'm going to mark this as a future 2.0 release item, unless things change in the next few weeks for 1.2.

ezmckinn:
Hi Michael,

Thanks — this seems fine to me.

Best,
Emmett

schnuerle:
Hi could you all leave your thoughts on this PR now to see if we can work this in 2.0 ASAP? It may be ok to work into the Modes work.

thekaveman:
@schnuerle IMHO this PR should be closed / rejected. The change being proposed is not correct. The spec and the README do align, this person was confused reading the JSON Schema.

schnuerle:
> @schnuerle IMHO this PR should be closed / rejected. The change being proposed is not correct. The spec and the README do align, this person was confused reading the JSON Schema.

Ah I see what you mean, agree. 

@ezmckinn we will be upgrading the JSON Schema to OpenAPI for 2.0 as well, so this change isn't needed.

PR #634, opened 2021-03-23
Adding SHARE NOW As A New Data Provider 

cemsiyok:
SHARE-NOW is added as a new mobility data provider
---
name: updates providers.csv file
about: SHARE-NOW is added as a new mobility data provider
title: Adding SHARE NOW As A New Data Provider 
---


schnuerle:
Thank you for your contribution. I've added this not to dev, but to the latest 1.1.0 release which is about to come out. You can see the change here:

https://github.com/openmobilityfoundation/mobility-data-specification/blob/release-1.1.0/providers.csv

Issue #608, opened 2020-12-16
[Requirements] Ability to express data sharing requirements in Policy


quicklywilliam:
### Is your feature request related to a problem? Please describe.

It has come up in a number of contexts that it would be useful to allow agencies to express data sharing requirements in MDS. I'm probably forgetting a few use cases, but here are some I have heard:

- In #607, allow agencies to state which geographies they would like aggregate metrics or reports for
- In #507, allow certain fields or endpoints to become "optional" in the sense of not being required by the spec, but still allow them to be "required" in the sense of the city explicitly requiring them of providers
- In #503, allow agencies to explicitly opt int to new optional or beta endpoints
- In #585, allow agencies to invite transparency and public accountability around the data they collect and why
- Improve data security by allowing operators to allocate access properly so that a city that does not request a given field/endpoint does not in fact have access to it
- Improve system efficiency by allowing operators to know in advance which fields/endpoints need to be stored and/or computed in advance

### Describe the solution you'd like

I think of MDS Policy as the digital representation of a City's mobility program policies – we are digitalizing their program PDF, in a sense. Data sharing requirements, including the requirement to provide MDS, are common in nearly any such policy. Hence, I think this belongs in MDS and in Policy specifically. One possibility that might meet some of the uses cases above would be adding a new rule type or set of rule types within the Policy endpoint that describe a city data sharing need.

### Is this a breaking change

I think this could probably be done in a non-breaking way.

### Impacted Spec

Policy.

### Describe alternatives you've considered

There has also been discussion of this living in a new endpoint. I'm open to that, but I like the simplicity of having a single (public!) endpoint return everything about a city's policy. In addition, there are often direct relationships between rules and data sharing needs (cities need data to assess compliance with a rule). I think these relationships might be made more clear and explicit through a carefully considered augmentation of MDS Policy.

karcass:
I have given some thought to putting this information in the Jurisdictions API, but I'm not strongly attached to that.

schnuerle:
See [notes from the WG Call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2020.01.07-(Joint-Working-Group)#minutes) today. Some good interest in this concept for the next release.

- Discovery for what cities require - APIs, endpoints, optional fields, version supported
- Can be tied to a multi-jurisdictional thing
- Like a specification guide for cities and MDS - digital representation of what’s in the data section of the permit (as it relates to MDS).
[...]

Issue #639, opened 2021-04-15
Retrieve operational zones from operators 


tybaltspark:
### Is your feature request related to a problem? Please describe.

Some permits contain defined operational zone (i.e. London), some others don't (i.e. Milan). In the case where cities dont impose an operational zone, then it's up to the provider to propose one (if at all).  
In most cases, cities are in the dark to know exactly what is the operational zone of each operator. One can guess it using vehicles positioning. However this only remains a guess and it is then difficult to track how these operational zones are evolving. 
The objective for cities is to better understand the coverage of each operator, and why some parts of the city do not show strong activities. Is it because the need is simply not there or because the vehicles are restricted from going in this part of town. 

### Describe the solution you'd like

We would propose that provider give their operational zone within the Provider API as part of an `info` endpoint. 
This information should be updated on a daily basis. 

### Is this a breaking change

No, it's not a breaking change. It's additive

### Impacted Spec

 * `provider`

### Describe alternatives you've considered

What we have done so far is asking the geographies directly from the providers to expose that on the Vianova platform. 
So this is done manually and needs recurrent updates. 

### Additional context

Three European cities have asked for this feature (Italy, UK and Germany).

schnuerle:
The way this is currently handled in MDS is the agency dictates this to the providers using the Policy API.  You are proposing that the providers send back a geojson boundary as part of one of the Provider APIs, so that the city/third parties can confirm that these align?

I have noticed in my city that opening an operator app will show the operating boundaries, and they are 1) different for each operator and 2) do not align to the operating areas in our city policy permit document.  I know our city's coordinator has tried to have operators update this but is not always successful, and it's not egregious enough to withdraw their permit.  Do you think this could help solve for an issue like this?

tybaltspark:
Some European cities (and probably US, and others) do not provide operational zones as part of the permit. Hence there is no communication through the Policy API to be done, as there is no enforcement as part of the permit.  

Operators do have different operating zones whether there is an operational zone provided by the authority as part of the permit or not. Four cities have expressed strong desire to know what these operational zones are in order for them to better understand service coverage & devices deployment. 
We see operators providing the operational zones to cities and local authorities as part of the provider API. Please let me know if you see any counter-indication or other experiences to complement this thread.

tybaltspark:
@schnuerle Is there a possibility that this could be discussed at the next meeting ? 
[...]

Issue #507, opened 2020-05-27
Method to Exclude some Provider Fields from Response


schnuerle:
### Is your feature request related to a problem? Please describe.

It is possible that cities or third parties may not want or need certain fields returned.

### Describe the solution you'd like

Add an option through the API request that allows some fields to be removed from the response.

An example solution could be the inclusion of a new parameter, e.g. `exclude`, that could have a list of fields the city does not want to have returned in the response.  If a city didn’t want trips, then they can add `/trips?exclude=route` to the API call.  If they don’t want trip or fee data, they could add `/trips?exclude=route,standard_cost,actual_cost`.  This list of possible options can be defined in the spec.

Not sure if the resulting fields would be completely removed, or be returned empty, or with a special value.  I prefer the first option.

This also has a benefit of getting only the data you need for the task at hand, reducing the returned data file size, and reducing processing on the provider’s side.  

### Is this a breaking change

A breaking change would require consumers or implementors of the API to modify their code for it to continue to function (ex: renaming of a required field or the change in data type of an existing field). A non-breaking change would allow existing code to continue to function (ex: addition of an optional field or the creation of a new optional endpoint). 

 * Yes, breaking

### Impacted Spec

For which spec is this feature being requested?

 * `provider` - for now let's just focus on Provider

### Describe alternatives you've considered

This is a proposed solution and discussion point to PR #504.

### Additional context

I have seen this in other APIs, so there is some precedent here.

vperron:
@schnuerle I think you meant `routes` ? trips is a whole endpoint, I'm not sure it can exclude itself.

I support this idea, even though I'm a little worried this would only land in a quite remote MDS version instead of a first version clarifying which fields are "optional" and what it means, which could get us going with 1.0 until this is settled.

schnuerle:
@vperron Yes that's what I meant.  Updated the original to clarify, more like this: `/trips?exclude=route,standard_cost,actual_cost`

Retzoh:
@schnuerle if I understand correctly, this means there would be two factors determining the fields included in the responses:
- providers may choose to include or exclude optional fields depending on data availability or other reasons
- cities may ask for some fields, optional or required, to be excluded from the responses

Is that correct? 

The way I understood #504 is to make it possible for providers to implement the Provider API without having to transmit routes at all if the data is judged too sensitive. The approach here is a bit different since an agency may or may not ask for routes. But if it does, the routes would still have to be transmitted.

[...]

PR #646, opened 2021-05-20
Policy Requirements

schnuerle:
This pull request is written to fulfill the needs of the discussions started in Issue #608 around agencies expressing data sharing requirements more clearly. 

## Explain pull request

This would add a new endpoint Requirements under the Policy API that allows agencies to digitally express only the data they need for their jurisdiction and operating permit. 

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`

## Additional context

The initial PR is a draft that will require feedback and edits from the MDS community before becoming part of the release, including external guidance documents and a new OMF repo to host requirement files for some agencies.


policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+#### Requirement Metadata
+
+Contains metadata applicable to the agency and at the top of its [Requirement](#requirement) data feed in the `metadata` section. 
+
+| Name                         | Type            | Required / Optional | Description              |
+| ---------------------------- | --------------- | -------- | ----------------------------------- |
+| `mds_release`                | text            | Required | Release of MDS that the requirements data feed aligns to, based on official MDS releases. E.g. "1.2.0" |
+| `version`                    | integer         | Required | Version of this file. Increment 1 with each modification. E.g. "3" |
+| `last_updated`               | [timestamp][ts] | Required | When this file `version` was last updated. E.g. "1611958740" |
+| `max_update_frequency`       | integer         | Required | The expected maximum frequency with which this file could be updated. E.g. "P1D" |
```
karcass, line 434: `P1D` is not an integer

policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+| `agency_uuid`                | UUID            | Required | UUID of the agency this file applies to. Must come from [agencies.csv](/agencies.csv) file. E.g. "737a9c62-c0cb-4c93-be43-271d21b784b5" |
+| `agency_name`                | text            | Required | Name of the agency this file applies to. E.g. "Louisville Metro" |
+| `agency_time_zone`           | text            | Required | Timezone used for dates and times across all MDS endpoints. E.g. "America/New_York" |
+| `agency_currency`            | text            | Required | Currency used for all monetary values across all MDS endpoints. E.g. "USD" |
+| `agency_policy_website_url`  | URL             | Required | URL of the agency's transportation policy page. E.g. "https://www.cityname.gov/transporation/shared-devices.htm" |
+| `agency_policy_document_url` | URL             | Optional | URL of the agency's operating permit rules that mention data requirements. E.g. "https://www.cityname.gov/mds_data_policy.pdf" |
+| `gbfs_required`              | Enum            | Required | yes/no. Is public GBFS required explicitly by providers? E.g. "yes" |
```
karcass, line 443: why not `boolean`?

policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+| Name                         | Type           | Required / Optional | Description              | 
+| ---------------------------- | -------------- | -------- | ----------------------------------- | 
+| `version`                    | text           | Required | Version number of an official MDS release | 
+| `provider_ids`               | UUID[]         | Required | Array of providers that apply to this part of the requirements | 
+
+[Top][toc]
+
+#### Requirement MDS APIs
+
+For each combination of MDS version and provider list, you can specify the MDS APIs, endpoints, and optional fields that are required per your agency's policy. This is an array within the [Requirement MDS Version](#requirement-mds-version) `mds_apis` section in the [Requirement](#requirement) data feed.
```
karcass, line 481: I think we will want support for `modes` and some SLA fields such as latency.  E.g. "/events must arrive within 30 seconds."

policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+
+Contains metadata applicable to the agency and at the top of its [Requirement](#requirement) data feed in the `metadata` section. 
+
+| Name                         | Type            | Required / Optional | Description              |
+| ---------------------------- | --------------- | -------- | ----------------------------------- |
+| `mds_release`                | text            | Required | Release of MDS that the requirements data feed aligns to, based on official MDS releases. E.g. "1.2.0" |
+| `version`                    | integer         | Required | Version of this file. Increment 1 with each modification. E.g. "3" |
+| `last_updated`               | [timestamp][ts] | Required | When this file `version` was last updated. E.g. "1611958740" |
+| `max_update_frequency`       | integer         | Required | The expected maximum frequency with which this file could be updated. E.g. "P1D" |
+| `omf_review`                 | Enum            | Required | yes/no. Was this file reviewed by OMF Staff for accuracy? E.g. "yes" |
```
janedotx, line 435: Sorry if I'm misunderstanding the original impetus for this, and I know that this field is being removed for now, but would have some kind of official OMF-sanctioned validation tool for the requirements endpoint payload suffice instead?

policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+| `omf_review_date`            | [timestamp][ts] | Optional | If `omf_review`, add timestamp. E.g. "1611958749" |
+| `agency_uuid`                | UUID            | Required | UUID of the agency this file applies to. Must come from [agencies.csv](/agencies.csv) file. E.g. "737a9c62-c0cb-4c93-be43-271d21b784b5" |
+| `agency_name`                | text            | Required | Name of the agency this file applies to. E.g. "Louisville Metro" |
+| `agency_time_zone`           | text            | Required | Timezone used for dates and times across all MDS endpoints. E.g. "America/New_York" |
+| `agency_currency`            | text            | Required | Currency used for all monetary values across all MDS endpoints. E.g. "USD" |
```
janedotx, line 440: Feels a bit superfluous since `currency` is specified in each policy, and I don't think it's used anywhere else.

policy/README.md
```diff
@@ -123,24 +130,34 @@ Policies will be returned in order of effective date (see schema below), with pa
...
 
+### Requirements
 
-### Flat Files
+Endpoint: `/requirements/`  
+Method: `GET`  
+`data` Payload: `{ requirements: [] }`, JSON objects that follow the schema [outlined here](#requirement).
+[Beta feature](/general-information.md#beta-features): *Yes (as of 1.2.0)*.
+
+See [Policy Requierment Examples](/policy/examples/requirements.md) for how this can be implemented.
```
jrheard, line 156:
```suggestion
See [Policy Requirement Examples](/policy/examples/requirements.md) for how this can be implemented.
```

policy/README.md
```diff
@@ -296,6 +313,8 @@ An individual `Rule` object is defined by the following fields:
 
 ### Geography
 
+**Depreciated:** see the new [Geography API](/geography#transition-from-policy) to understand the transisiton away from this endpoint, and how to support both in a MDS 1.x.0 release.
```
jrheard, line 316:
```suggestion
**Deprecated:** see the new [Geography API](/geography#transition-from-policy) to understand the transition away from this endpoint, and how to support both in a MDS 1.x.0 release.
```

policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
 
 The internal mechanics of ordering are up to the Policy editing and hosting software.
 
+[Top][toc]
+
+### Requirement
+
+The agency Policy Requirement file ennumerates all of the parts of MDS that an agency requires from providers, including APIs, endpoints, and optional fields, as well as information for providers about the APIs the agency is hosting. The requirements are specific to the needs and use cases of each agency, and ensures there is clarity on what data is being asked for in operating policy documents from providers, reducing the burden on both. This also allows additional public transparency and accountability around data requirements from agencies, and encourages privacy by allowing agencies to ask for only the data they need.
```
jrheard, line 391:
```suggestion
The agency Policy Requirement file enumerates all of the parts of MDS that an agency requires from providers, including APIs, endpoints, and optional fields, as well as information for providers about the APIs the agency is hosting. The requirements are specific to the needs and use cases of each agency, and ensure there is clarity on what data is being asked for in operating policy documents from providers, reducing the burden on both. This also allows additional public transparency and accountability around data requirements from agencies, and encourages privacy by allowing agencies to ask for only the data they need.
```

policy/README.md
```diff
@@ -365,6 +384,130 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+
+This endpoint it not authenicated (ie. public), and allows the discovery of other public APIs like Geography, Policy, and Jurisdiction. The agency can host this as a file or API on their servers, on a third party server, or the OMF can host on behalf of an agency in the [agency requirements repo](#). See this [hosting guidance document](#) for more information.  This requirements file can be [referenced directly](https://github.com/openmobilityfoundation/governance/blob/main/technical/OMF-MDS-Policy-Language-Guidance.md) in an agency's operating permit/policy document when discussing program data requirements.
```
jrheard, line 393:
```suggestion
This endpoint is not authenticated (ie. public), and allows the discovery of other public APIs like Geography, Policy, and Jurisdiction. The agency can host this as a file or API on their servers, on a third party server, or the OMF can host on behalf of an agency in the [agency requirements repo](#). See this [hosting guidance document](#) for more information.  This requirements file can be [referenced directly](https://github.com/openmobilityfoundation/governance/blob/main/technical/OMF-MDS-Policy-Language-Guidance.md) in an agency's operating permit/policy document when discussing program data requirements.
```

policy/examples/requirements.md
```diff
@@ -0,0 +1,118 @@
...
+This file presents a series of example [Policy documents](../README.md#requirement) for Agencies to use as templates.
+
+## Table of Contents
+
+- [Trips Only](#trips-only)
+- [Vehicles Only](#vehicles-only)
+
+## Trips Only
+
+Version 1.1.0 for 2 providers requiring only Provider `/trips` with the optional `parking_verificaiton_url` field.
```
jrheard, line 12:
```suggestion
Version 1.1.0 for 2 providers requiring only Provider `/trips` with the optional `parking_verification_url` field.  
```

jrheard (commented):
just some typo fixes!

policy/examples/requirements.md
schnuerle: Fixed thank you!

policy/README.md
schnuerle: Updated thank you!

policy/README.md
schnuerle: Thank you for the catches. Fixed!

policy/README.md
schnuerle: Corrected, thanks!

policy/README.md
schnuerle: Got it, thank you!

policy/README.md
schnuerle: I think it's good for an agency to specify the currency it wants here at a high level so it's clear, in case elsewhere in the spec a value is left off.

policy/README.md
schnuerle: Yes that would also work, but building and maintaining a tool would be a much bigger undertaking, and may still have the same implications here, eg. "this feed was validated by the OMF tool", which could still be a liability issue. Not sure so removing for now. Instead it could just be a "validator" text field that says who or what validated the feed, like 'Ride Report' or 'Populus' or 'OMF Validator', etc.

policy/README.md
schnuerle: I think SLA's and what they entail may come in a the next version of this. Unless you have some specific proposals that we can look at and agree on. I see a lot of variability here that could be challenging.

But your mention of modes makes me think maybe modes could be listed for each Release/Provider/state/end combination too, so different modes could apply to the same provider, or different MDS versions...

policy/README.md
schnuerle: Not sure what's best. I don't think we have any boolean fields in MDS so this would be the first.  Could go 0/1 or yes/no. We should talk about it.

policy/README.md
schnuerle: Yes, updated, thanks!

policy/README.md
schnuerle: Thinking about the SLA some more... Maybe there could be a field for 'expected_latency' at the endpoint level which specified an ISO interval that is the minimum expected time that an agency would like to get the data. Eg 5 minutes if they want near real time data, but could be 24 hours if they only have use cases that require it to be updated once per day? Not sure how that sits with certain endpoint latency as required in the spec already, and if this would be allowed to override that.

schnuerle:
Finished a round of updates and adding examples based on last week's WG conversations. Please review and provide additional thoughts of your own and on the comment that are already here.

policy/README.md
jiffyclub: Definitely would prefer a boolean value here. JSON supports `true`/`false`, so no need to come up with alternate schemes.

schnuerle: Changed to boolean in the spec and examples

policy/README.md
schnuerle: I think SLA should wait for a future version of this, so let's put a pin in it.  I did add `vehicle_types` [here](https://github.com/openmobilityfoundation/mobility-data-specification/commit/6c6a89d7ed24f220eaa6607caf0efcd7a853aa4e) so you can optionally specify an array of types that apply to the specific version and providers in question. Does this solve your mode idea?

policy/README.md
schnuerle: So we can add a 'feed_validator' text field which would just be the free form name of the organization that validated the tool. No need to enumerate values.  And a second 'feed_validator_date' field that is the ISO date (eg 2021-06-02) of when the most recent validation was done, useful to compare the last updated date of the feed to this date.

schnuerle:
We will be talking about this and recent updates on tomorrow's [working group call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2021.07.08-(Joint-Working-Group)).

policy/examples/requirements.md
```diff
@@ -579,3 +580,74 @@ Version 1.1.0 for 2 providers requiring Provider `/status_changes` with the mini
...
 [Top](#table-of-contents)
+
+## GBFS Only
+
+Since Requirements allows the GBFS versions and optional endpoints and fields to be defined, an agency could use it to only require public GBFS feeds, and not require MDS at all.
+
+```json
+{
+  "metadata": {
+    "mds_release": "1.2.0",
```
quicklywilliam, line 591: I think this field got removed from this level?

policy/examples/requirements.md
```diff
@@ -579,3 +580,74 @@ Version 1.1.0 for 2 providers requiring Provider `/status_changes` with the mini
...
+    "version": "2",
+    "last_updated": "1611958740",
+    "max_update_interval": "T1M",
+    "agency_uuid": "737a9c62-c0cb-4c93-be43-271d21b784b5",
+    "agency_name": "Louisville Metro",
+    "agency_timezone": "America/New_York",
+    "agency_language": "en-US",
+    "agency_currency": "USD",
+    "agency_website_url": "https://www.cityname.gov/transporation/",
+    "url": "https://mds.cityname.gov/policy/requirements/1.2.0"
```
quicklywilliam, line 601: url path should be 2?

policy/README.md
```diff
@@ -365,6 +387,231 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+  },
+  "mds_versions" [ 
+    // Requirement MDS Versions
+  ]
+}
+```
+
+| Name                         | Type            | Required / Optional | Description              |
+| ---------------------------- | --------------- | -------- | ----------------------------------- |
+| `mds_release`                | text            | Required | Release of MDS that the **requirements data feed** aligns to, based on official MDS releases. E.g. "1.2.0" |
```
quicklywilliam, line 481: would suggest mds_requirements_version or something that makes it more clear that this applies to the requirements file itself, not the APIs it is describing

policy/examples/requirements.md
schnuerle: What do you mean by "2"?

policy/README.md
schnuerle: Yes good point on clarifying the intent of this version of this file. Will adjust.

policy/examples/requirements.md
quicklywilliam: ie https://mds.cityname.gov/policy/requirements/2 – I was imagining the URL path would include the version of the file rather than the version of the requirements spec

schnuerle: I think the path should be static and not change with each edit, though it could change with each MDS version. That's why there is a 'file_version' field and 'last_updated' field, and why OMF will host files in GitHub for a built in version history. Even if it did change with each edit, I think it would be `https://mds.cityname.gov/policy/requirements/1.2.0/2`. Even so, I'd rather see the latest version stay at the same URL, and maybe older version can have their URLs updated in a consistent way, eg v1 would show up at `https://mds.cityname.gov/policy/requirements/1.2.0/1` when v2 is made. But really, this is just an example, and agencies can do what they wish.

Maybe we need some additional language about this, and also recommend agencies host on a system that tracks versions publicly?

policy/README.md
schnuerle: Updated to 'file_version' which I think is clear.

policy/README.md
schnuerle: It think we are going to leave this off for the initial release. I'd like to wait and see what sort of adoption we get before building validating tools.

policy/examples/requirements.md
quicklywilliam: Got it. Sounds like I was making the wrong assumption, which is that Agencies would host the old versions of their requirements files.

policy/README.md
```diff
@@ -365,6 +387,266 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+| `url`                | URL   | Optional | Location of API endpoint url. Required if the API is unauthenticated and public, optional if endpoint is authenticated and private. E.g. "https://mds.cityname.gov/geographies/geography/1.1.0"  | 
+| `available_fields`   | Array | Optional | Array of optional field names provided by the agency. Can be omitted if none are required. Use dot notation for nested fields. See **special notes** below. | 
+
+**Special notes about `required_fields`.** 
+
+- All fields marked 'Required' in MDS are still included by default and should not be enumerated in `required_fields`. They are not affected by the Requirements endpoint.
+- Fields in MDS marked 'Required if available' are still returned if available, and are not affected by the Requirements endpoint.
+- To reference fileds that are lower in a heirarchy, use [dot separated notation](https://docs.oracle.com/en/database/oracle/oracle-database/18/adjsn/simple-dot-notation-access-to-json-data.html#GUID-7249417B-A337-4854-8040-192D5CEFD576), where a dot between field names represents one nested level deeper into the data structure. E.g. 'gps.heading' or 'features.properties.rules.vehicle_type_id'.
```
janedotx, line 648: `fields`

policy/README.md
```diff
@@ -123,24 +133,34 @@ Policies will be returned in order of effective date (see schema below), with pa
 
 `provider_id` is an implicit parameter and will be encoded in the authentication mechanism, or a complete list of policies should be produced. If the Agency decides that Provider-specific policy documents should not be shared with other Providers (e.g. punitive policy in response to violations), an Agency should filter policy objects before serving them via this endpoint.
 
-#### Geographies
+### Geographies
 
-**Note:** see the new [Geography API](/geography#transition-from-policy) to understand the transisiton away from this endpoint, and how to support both in the MDS 1.1.0 release.
+**Deprecated:** see the new [Geography API](/geography#transition-from-policy) to understand the transistion away from this endpoint, and how to support both in a MDS 1.x.0 release.
```
janedotx, line 138: typo: `transition`

policy/README.md
schnuerle: Fixed. Thanks for finding typos!

policy/README.md
schnuerle: Fixed, thank you.

jfh01:
What do you think about reordering the list of examples here? https://github.com/openmobilityfoundation/mobility-data-specification/blob/ms-requirements/policy/examples/requirements.md. Thinking from most->least common/complex. e.g. start with a simple MDS provider + GBFS and then get into some of the other options

policy/README.md
```diff
@@ -365,6 +389,280 @@ If a vehicle is matched with a rule, then it _will not_ be considered in the sub
...
+| `endpoint_name`      | Text  | Required | Name of the required endpoint under the API. At least one endpoint is required. E.g. for MDS 'provider': 'trips' | 
+
+**Provider Endpoints** - Specific to the `required_apis` array 
+
+| Name                 | Type  | Required/Optional | Description                | 
+| -------------------- | ----- | -------- | ----------------------------------- | 
+| `required_endpoints` | Array | Required | Array of optional endpoints required by the agency. At least one is required. Endpoints not listed will not be available to the agency. | 
+| `required_fields`    | Array | Optional | Array of optional field names required by the agency. Can be omitted if no optional fields are required. Use dot notation for nested fields. See **special notes** below. | 
+| `disallowed_fields`  | Array | Optional | Array of optional field names who's values are to be returned by the agency as `null`, even if required in MDS. Use dot notation for nested fields. See **special notes** below. |
```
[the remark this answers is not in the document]
schnuerle: My main concern with removing the required field itself is that any schema validation will then fail, since the field is required. 

Of course, if there is also validation on the value of that field, then that would fail as well. So maybe it's not an issue per se. Language may have to be added to say that schema validations may fail if required fields are removed.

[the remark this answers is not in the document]
quicklywilliam: Yep, I think that's a good path. I think maybe the validation _should_ fail - we're explicitly allowing Agencies to break the spec by disallowing required fields.

policy/examples/requirements.md
```diff
@@ -0,0 +1,717 @@
...
+
+Version 1.2.0 of MDS Policy and Geography for agencies to publish rules/fees/incentives and operating/equity/no-ride/slow speed/parking areas to all providers, and require GBFS's geofencing_zones.  
+
+```json
+{
+  "metadata": {
+    "mds_release": "1.2.0",
+    "file_version": "4",
+    "last_updated": "1611729218",
+    "max_update_interval": "T1M",
```
avatarneil, line 26: I'm pretty sure this should be `P1M` (if the update interval should be monthly), or `PT1M` (if the update interval should be minutely). I don't think that `T1M` is a valid ISO8601 duration.

Ditto for other occurrences of `T1M` in the examples.

policy/README.md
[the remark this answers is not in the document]
avatarneil: I think that generally in MDS we've considered `null` equivalent to `undefined` by convention (I swear that there was some issue for this at one point), so we should make sure that this aligns. In a lot of relational databases, `undefined` is _persisted_ as `null`, so the semantic difference between "this was never available" versus "this was omitted when it was sent" is lost anyways.

policy/examples/requirements.md
schnuerle: Yes it was correct in the spec definition example, but not correct in all the examples. Updated them all now to be P1M.

policy/README.md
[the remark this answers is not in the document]
schnuerle: I've made an update in this [commit](https://github.com/openmobilityfoundation/mobility-data-specification/commit/648b2192b4b6eee9ab9b013c1e2f0a23ba68e669#diff-5dbb467da8c3ceb340b4c471496e28b63ed45d1ee3e15ceed7c44c3b88340720). Let me know what you think.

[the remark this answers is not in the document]
quicklywilliam: lgtm!

schnuerle:
After lots of great review and feedback from dozens of people over the last few months, I believe this is ready to be merged to 'dev'! Thanks to @quicklywilliam for the original idea with #608, and this has grown to solve other issues raised by the community as well including #507 and #639.

If there are new suggested changes that come up, we can open an Issue or Discussion to resolve them before the 1.2.0 release candidate is made for approval, or even make minor updates once the RC is made.

Issue #645, opened 2021-05-11
MDS milliseconds timestamp constraints


tonial:
### Is your feature request related to a problem? Please describe.

As an agency, we often see second timestamps instead of millisecond timestamp when a provider uses MDS for the first time and tries to send us data.

### Describe the solution you'd like

In order to speed up bug detection on our side, we added a constraint that specifies that a timestamp should have 13 digits which allows dates between 2001-9-9 and 2286-11-20 (which is big enough).
Adding this into the Timestamp description and json schema could help to prevent this mistake.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

All timestamps in all APIs

schnuerle:
Thank you @tonial. A reasonable request that could be updated in the next release. As the schema expert, what do you think @thekaveman ?

thekaveman:
Yes, very reasonable request and should be as easy as modifying the common [timestamp definition](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/schema/templates/common.json#L378).

Right now, we have:

```json
"timestamp": {
  "$id": "#/definitions/timestamp",
  "type": "number",
  "description": "Integer milliseconds since Unix epoch",
  "multipleOf": 1.0,
  "minimum": 0
},
```

Because this is a number type, we can't do a (string) length check. But an easy modification would be to change the `minimum` to correspond to @tonial's 13-digit starting point of `1000000000000`. 

Alternatively, a more realistic minimum could correspond to (for example) Jan 1, 2018 (the year MDS was first released), in UTC `1514764800000`.

schnuerle:
Approved via PR #648. Thank you!

PR #648, opened 2021-05-27
Add minimal value to all timestamp json definition

tonial:
The goal is to prevent the easy to make mistake of using a second
timestamp instead of a millisecond timestamp.

This PR implements issue #645

### DIFF
--- a/agency/get_stops.json
+++ b/agency/get_stops.json
@@ -156,7 +156,7 @@
       "type": "number",
       "description": "Integer milliseconds since Unix epoch",
       "multipleOf": 1.0,
-      "minimum": 0
+      "minimum": 1514764800000
     },
     "uuid": {
       "$id": "#/definitions/uuid",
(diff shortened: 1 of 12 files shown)


schnuerle:
Hi @thekaveman could you take a look at this? Is it ok to require all of these values exist and be greater than 0? Thanks.

thekaveman (approved):
This looks good to me!

Important to note this change doesn't impact any _requirement_ for timestamps in any specific endpoint. It only changes the definition of a timestamp to be always greater than or equal to Jan 1, 2018 00:00:00 UTC.

A good example is in Provider `/trips`, the optional `publication_time` field. This change does not make the field required, it just dictates the lower bound if a value is given.

The change could impact testing scenarios or synthetic data generation that isn't using realistic time values, but IMHO is unlikely to impact real-world data producers in any meaningful way.

schnuerle (approved):
Thanks so much for this, looks good!

PR #649, opened 2021-06-03
Updated MDS URL for Superpedestrian

ezmckinn:
---
name: Emmett McKinney
about: providers.csv
title: Updated MDS URL for Superpedestrian

---

## Explain pull request

Updating base URL for Superpedestrian to `https://mds.linkyour.city/mds`. 

## Is this a breaking change

* Yes, breaking

## Impacted Spec

* `agency`
* `provider`

## Additional context

Our previous URL (https://wrangler-mds-production.herokuapp.com/mds) will remain active until July 6 at 12:00:00 UTC-4. So at the time of this PR, it is not a breaking change. But it will become a breaking change once we deprecate our previous service.


schnuerle (approved):
Thanks for the update

Issue #637, opened 2021-04-12
Provider Vehicles - Beta 0.4.1 - Feedback 


schnuerle:
Gather real world feedback to see how we can move the beta feature [Vehicles](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/provider#vehicles) out of beta. 

### Describe the solution you'd like

From the MDS Survey, about half of agencies are using Vehicles, most providers are, and almost all companies support it. Seems to have good real world usage. We need feedback here from those that are using it to know what may need to change before it becomes a stable part of MDS.

### Is this a breaking change

 * I'm not sure yet

### Impacted Spec

For which spec is this feature being requested?

 * [provider](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/provider#vehicles)

### Describe alternatives you've considered

Leave as a beta feature for longer until we have feedback. 

### Additional context

Reviewed as part of the MDS 1.2.0 release review Midway Checkpoint by the Working Group Steering Committees.

dirkdk:
I think soon Vehicles will be a core part of deployment cap enforcement so I think it is a good idea.

dirkdk:
Related, I am wondering if the requirement to also have a GBFS API should be dropped. I'm trying to think back why OMF made this part of the MDS spec https://github.com/openmobilityfoundation/mobility-data-specification#gbfs-requirement

If the reason was compliance checks, I would say the Vehicles endpoint replaces this need and we can drop it. We would like this very much as a provider. This would enable us to use different Quality of Service levels for MDS vs GBFS. For instance, as GBFS is public, we rate limit it but run into problems with data aggregators (@jiffyclub can attest to this) sometimes. If GBFS is not used for compliance anymore we can safely adjust our levels without running the risk of making compliance checks more fragile.

schnuerle:
Thank you for the feedback on Vehicles Dirk.

@dirkdk per your question about GBFS, could you please put that into a new issue for a deeper discussion there?

schnuerle:
See new issue #641 for discussion around GBFS and MDS, and this issue plus the new issue will be the topics at tomorrow's [working group meeting](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2021.04.29-(Joint-Working-Group)).

jiffyclub:
We at Populus are now consuming /vehicles and have found it useful in its current spec, so I think it's ready to move out of beta.

[...]

PR #665, opened 2021-07-23
Move Provider Vehicles Out of Beta

schnuerle:
## Explain pull request

PR to implement the discussions in #637 and move the /vehicles endpoint in Provider out of beta

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `provider`

## Additional context

N/A


schnuerle (commented):
Approved. Out of beta. 

- Defined jurisdiction
- updated for all vehicle states
- Note about reconciliation with status_changes and that status_changes is source of truth
- remove beta feature tag
- update schema

PR #677, opened 2021-08-04
Clarify policy rule min and max are inclusive

jiffyclub:
## Explain pull request

The policy docs don't say anywhere whether the minimum and maximum values specified on policy rules are to be interpreted as a closed or open interval (that is, whether the endpoints are included in the interval). Most real world policies seem to treat these min and maxes inclusively, e.g. parking for a max of 2 hours or a 70 MPG speed limit. You're not in violation until you go over those maximums. So this change clarifies that min and max values on policy rules are meant to define an inclusive (or closed) interval where values equal to the min and max match the rule.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`

## Additional context

N/A

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -10,7 +10,7 @@ This specification describes the digital relationship between _mobility as a ser
 
 - [General Information](#general-information)
   - [Versioning](#versioning)
-  - [Update Frequency](#update-frequency)  
+  - [Update Frequency](#update-frequency)
 - [Background](#background)
 - [Distribution](#distribution)
 - [REST Endpoints](#rest-endpoints)
@@ -37,7 +37,7 @@ This specification describes the digital relationship between _mobility as a ser
     - [Programs](#requirement-programs)
     - [Data Specs](#requirement-data-specs)
     - [APIs](#requirement-apis)
- 
+
 ## General information
 
 The following information applies to all `policy` API endpoints.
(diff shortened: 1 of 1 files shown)


jiffyclub:
Per discussion with @avatarneil on an [MDS working group call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2021.08.12-(Joint-Working-Group)) we may want to make the treatment of min and max a user configurable option. For some policies, especially those governing continuous values, it can make it easier to express a policy if min/max are _excluded_ from the range (making an open or half-open) an interval.

schnuerle:
Realize this clarification language could be part of 1.2.0. What do you think about how to make these a configurable option, or is this idea for a future release? 

Looks like there also may be a conflict with dev now to resolve.

jiffyclub:
See #689 for an alternative approach that lets users specify how `minimum` and `maximum` should be handled.

schnuerle:
Per our [working group meeting](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2021.09.09-(MDS-Working-Group)#action-items), PR #689 was selected as preferable to this solution. IF agreed, @jiffyclub you may close this PR.

jiffyclub:
Closing in favor of #689.

PR #686, opened 2021-09-01
Clarify language for Trips responses

bhandzo:
This clarifies what happens between when an hour passes, and a provider finishes processing data for that hour.

## Explain pull request

This clarifies how providers should respond to requests for Trips data between when the hour ends and when they finish processing data.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `provider`

## Additional context

...

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -185,11 +185,7 @@ The `/trips` API should allow querying trips with the following query parameters
 | --------------- | ------ | --------------- |
 | `end_time` | `YYYY-MM-DDTHH`, an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) extended datetime representing an UTC hour between 00 and 23. | All trips with an end time occurring within the hour. For example, requesting `end_time=2019-10-01T07` returns all trips where `2019-10-01T07:00:00 <= trip.end_time < 2019-10-01T08:00:00` UTC. |
 
-If the provider was operational during the requested hour the provider shall return
-a 200 response, even if there are no trips to report (in which case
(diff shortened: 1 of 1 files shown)


jiffyclub:
It looks like this PR was based off the `main` branch and not `dev` where this language has already been updated. See https://github.com/openmobilityfoundation/mobility-data-specification/tree/dev/provider#trips---responses and #437.

(As a side note, I'm kind of unclear on what the difference between `main` and `dev` is, the fact that we don't make PRs against the default branch is def a path to this kind of confusion.)

schnuerle:
I think you are right @jiffyclub - this seems to be addressed in the latest changes to the dev branch.  I didn't review fully and missed this when @bhandzo asked me offline about this language and I assumed it hadn't been changed. So Ben take a look and I think this work in the dev branch covers your concern.

For the main vs dev question, dev is where we do the work as a release happens, and then only when the release is approved it goes to main. The reason PRs don't go to the default branch main is we want people who visit the repo to see the latest approved spec by default. A release candidate for approval is made off of dev, and then a new release can start right after that, where new things could be pushed to dev. So it's really a way to manage the release approval process. There has been discussion about this process over the years for sure. The latest info can be seen in the [Release Guidance](https://github.com/openmobilityfoundation/governance/blob/main/technical/ReleaseGuidelines.md#branch-mechanics) doc.

bhandzo:
Yep! That language totally covers it.

schnuerle:
Ok so we won't use this PR since it's resolved in dev and will be in MDS 1.2.0.

Issue #666, opened 2021-07-24
Add a rate_applies_when field to policy rules


jiffyclub:
### Is your feature request related to a problem? Please describe.

As I've dug into policies and #658 especially I've discovered that there's no guidance about when a fee/subsidy associated with a policy should be applied to an event. Is it when the event is in violation of the policy, or when it's in compliance? And in the discussion in #658 I think we've made clear that for some rules one way makes more sense, and in others the other way, so I don't think there's one best way to go.

### Describe the solution you'd like

Instead of picking one convention or the other I think we could have a `rate_applies_when` field on [policy rules](https://github.com/openmobilityfoundation/mobility-data-specification/tree/7dca90a65a53491a4f5c06696ecfcb8fcbb1261b/policy#rules) that could have values of `violation` or `compliant` that would make this explicit on a per-rule basis.

### Is this a breaking change

A breaking change would require consumers or implementors of the API to modify their code for it to continue to function (ex: renaming of a required field or the change in data type of an existing field). A non-breaking change would allow existing code to continue to function (ex: addition of an optional field or the creation of a new optional endpoint). 

 * No, not breaking

### Impacted Spec

For which spec is this feature being requested?

 * `policy`

### Describe alternatives you've considered

The alternative I've considered is to pick either compliance or violation as the standard for applying a rate to an event. People would have to structure their rules such that applying the standard would have the desired outcome. I'm not a fan of that, though, because I think some fee rules have to be constructed in counter-intuitive ways when you have to choose one way or the other (see discussion in #658). By letting people specify how rates are meant to be interpreted alongside their rules they can write rules in the ways that make sense to them.

### Additional context

N/A.

avatarneil:
I think this is definitely an interesting idea (essentially allowing the flipping of positive vs negative association). I do think that if we introduce this concept (or something along these lines), we may want it to live at the Policy level as opposed to the Rule level. I can definitely imagine some weirdness if you have one Policy which has rules using different association tactics, specifically when thinking about `count` policies where you may have some overflow (e.g. https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/policy/examples/provider-cap.json). 

I'm sure I'll have more feedback after I think about this some more, I just noticed this issue a few minutes ago!

schnuerle:
Completed with #688. Thank you all!

PR #688, opened 2021-09-02
Add a rate_applies_when field to Policy Rules

jiffyclub:
## Explain pull request

This is implementing my proposal from #666. The `rate_applies_when` field allows people constructing policy rules with attached rates to specify when the rate is applicable: is it when something is within the rule bounds or when it is outside of them? The `rate_applies_when` when field may take on the values `in_bounds` and `out_of_bounds`. It is used in conjunction with the rule `minimum` and `maximum` fields. It defaults to `out_of_bounds`.

I also used this PR to bring the Policy JSON schema and an example policy up-to-date following the merger of #658.

## Is this a breaking change

* I'm not sure

This field is optional so it doesn't break in that sense, but having a defined notion of when rates apply could change the interpretation of existing policies.

## Impacted Spec

* `policy`

## Additional context

The MDS policy spec does not currently specify when rates apply. In #658 @avatarneil argued that they apply when an event falls _outside_ the rule bounds, but in the `rate` rule examples in the [example policies](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/policy/examples/README.md) the rates are applying to events/counts _within_ the rule bounds. With the merger of #658 we now have example rules that are not consistent in their interpretation of when rates apply. (The example rate policies do not specify `minimum` or `maximum` fields, so the are implicitly 0 and infinity, respectively, though perhaps it could be argued that `minimum` and `maximum` are not applicable to rate type rules.)

See also my proposal in #666.

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -10,7 +10,7 @@ This specification describes the digital relationship between _mobility as a ser
 
 - [General Information](#general-information)
   - [Versioning](#versioning)
-  - [Update Frequency](#update-frequency)  
+  - [Update Frequency](#update-frequency)
 - [Background](#background)
 - [Distribution](#distribution)
 - [REST Endpoints](#rest-endpoints)
@@ -37,7 +37,7 @@ This specification describes the digital relationship between _mobility as a ser
     - [Programs](#requirement-programs)
     - [Data Specs](#requirement-data-specs)
     - [APIs](#requirement-apis)
- 
+
 ## General information
 
 The following information applies to all `policy` API endpoints.
@@ -117,8 +117,8 @@ Authorization is not required. An agency may decide to make this endpoint unauth
 
 ### Policies
 
-Endpoint: `/policies/{id}`  
-Method: `GET`  
+Endpoint: `/policies/{id}`
+Method: `GET`
 `data` Payload: `{ "policies": [] }`, an array of objects with the structure [outlined below](#policy).
 
 #### Query Parameters
@@ -139,8 +139,8 @@ Policies will be returned in order of effective date (see schema below), with pa
 
 **Deprecated:** see the new [Geography API](/geography#transition-from-policy) to understand the transistion away from this endpoint, and how to support both in a MDS 1.x.0 release.
 
-Endpoint: `/geographies/{id}`  
-Method: `GET`  
+Endpoint: `/geographies/{id}`
+Method: `GET`
 `data` Payload: `{ geographies: [] }`, an array of GeoJSON `Feature` objects that follow the schema [outlined here](#geography) or in [Geography](/geography#general-information).
 
 #### Query Parameters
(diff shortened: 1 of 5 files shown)


schnuerle:
Another note from the meeting, when updating the schema, remember to update the files in the [schema templates folder](https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/schema/templates/policy/policy.json) too. That's what's used to generate the schema files.

jiffyclub:
I've added an example in d49ae26 and updated the schema template in f38cc96.

One question I had was whether I needed to add `rate_applies_when: "in_bounds"` to the [example rate policies](https://github.com/openmobilityfoundation/mobility-data-specification/tree/b8362b89d5df731174a1338e6ea5714507b4549c/policy/examples#metered-parking-fees). I can't really tell if `minimum`/`maximum` have meaning for those, but if they do they would need `rate_applies_when: "in_bounds"` since the range is 0 - infinity. Or maybe `rate_applies_when` doesn't apply when there's no bounds on the rule, the rate always applies.

schnuerle:
> I've added an example in [d49ae26](https://github.com/openmobilityfoundation/mobility-data-specification/commit/d49ae265d244a0e597acdc53d8a1b71cfd36b820) and updated the schema template in [f38cc96](https://github.com/openmobilityfoundation/mobility-data-specification/commit/f38cc964992e599f2cae951892c0ad31401b4bfd).
> 
> One question I had was whether I needed to add `rate_applies_when: "in_bounds"` to the [example rate policies](https://github.com/openmobilityfoundation/mobility-data-specification/tree/b8362b89d5df731174a1338e6ea5714507b4549c/policy/examples#metered-parking-fees). I can't really tell if `minimum`/`maximum` have meaning for those, but if they do they would need `rate_applies_when: "in_bounds"` since the range is 0 - infinity. Or maybe `rate_applies_when` doesn't apply when there's no bounds on the rule, the rate always applies.

@avatarneil @marie-x could you weigh in on this question so we can finish this PR for the release?

avatarneil:
I don't think we need to add it. The new (as of 1.2) rate policy examples use the default `out_of_bounds` methodology, and I think for the existing rate policies (as Matt mentioned) `rate_applies_when` doesn't apply when there are no bounds on the rule.

jiffyclub:
I would argue that if anything the existing rate examples are using `in_bounds`, since the bounds are infinite if unspecified. But I do think it's fair to say that if min and max are not specified a rate is clearly meant to apply in all cases.

schnuerle:
Look ready to approve.  I'm a little concerned that removing 'once' as an option completely, replacing it with 'once_on_unmatch' and 'once_on_match' may break some people's implementations if they are looking at 'once', but it would more likely just not have their code find these new options, which I think is ok as it's not breaking their processing. This change will be in the release notes and schema for 1.2 which should be adequate.

schnuerle (approved):

PR #689, opened 2021-09-02
Allow users to specify rule min/max treatment

jiffyclub:
## Explain pull request

MDS does not currently specify whether Policy Rule `minimum` and `maximum` values are considered within bounds when evaluating whether something matches the rule (in other words, whether the rule bounds are a open, closed, or half-open interval). This PR adds new fields to Policy Rules that allow users to individually specify the treatment of the `minimum` and `maximum` values.

The new fields are called `inclusive_minimum` and `inclusive_maximum`. They are boolean fields that both default to true.

## Is this a breaking change

* No, not breaking

I don't think this is breaking, but it does define that by default rule `minimum` and `maximum` are considered within the rule bounds, which could potentially break interpretation of existing policy rules.

## Impacted Spec

* `policy`

## Additional context

This is an alternative PR to #677, we should pick one or the other.

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -10,7 +10,7 @@ This specification describes the digital relationship between _mobility as a ser
 
 - [General Information](#general-information)
   - [Versioning](#versioning)
-  - [Update Frequency](#update-frequency)  
+  - [Update Frequency](#update-frequency)
 - [Background](#background)
 - [Distribution](#distribution)
 - [REST Endpoints](#rest-endpoints)
@@ -37,7 +37,7 @@ This specification describes the digital relationship between _mobility as a ser
     - [Programs](#requirement-programs)
     - [Data Specs](#requirement-data-specs)
     - [APIs](#requirement-apis)
- 
+
 ## General information
 
 The following information applies to all `policy` API endpoints.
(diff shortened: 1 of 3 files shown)


jiffyclub:
On the MDS call today Zack Bouz (LADOT) raised the point that rule `minimum` and `maximum` are expected to be integers, per the spec. With the assumption that those values will always be integers it makes it easier to do something like #677, pick inclusive, and let people construct rules under that. I'll try to think of an example where it still would be nice to allow one or the other to be exclusive.

jiffyclub:
However, the question is really not whether the rule `minimum` and `maximum` are integers, but whether the values we're checking against the rules are integers. If they are ever floats I would still want this option, or we would need to define rounding behavior for converting floats to integers for this purpose.

schnuerle:
Per the [working group action items](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2021.09.09-(MDS-Working-Group)#action-items), here is the [schema template location](https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/schema/templates/policy/policy.json) for policy which needs to be updated. The templates generate the files during the release creation.

jiffyclub:
449a94f updates the policy schema template. Additionally, #688 now includes an example of these in practice.

schnuerle (approved):

PR #693, opened 2021-09-08
Align Policy JSON Schema & Examples with spec

### LINKED ISSUE #599: [Policy API] Response when returning a single object is not clear
Ref: https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/policy/README.md#rest-endpoints

I think the specs should be more clear about how the exact response should be when it comes to returning a single object vs multiple. In fact, each of these cases should have their own little section:

- `/policies`
- `/policies/{id}`
- `/geographies`
- `/geographies/{id}`

When "bundling" the docs on single/multiple endpoints, confusion and conflicts arise. An example: `id` is listed as a query parameter along with `start_date` and `end_date`. Though this is not really a query parameter, but part of the endpoint.

I studied the code in *mds-core* and found some undocumented responses when it comes to single objects

[/policies](https://github.com/openmobilityfoundation/mds-core/blob/f2e8a81710ad4cd0f7ea59acd3ba0425bed512d4/packages/mds-policy/api.ts#L105)
```js
        res.status(200).send({ version: res.locals.version, data: { policies: active } })
```
=> Would produce expected result according to docs

[/policies/{id}](https://github.com/openmobilityfoundation/mds-core/blob/f2e8a81710ad4cd0f7ea59acd3ba0425bed512d4/packages/mds-policy/api.ts#L146)
```js
        res.status(200).send({ version: res.locals.version, data: { policy } })
```
=> Would produce something not expected or documentet, the data object would contain a **policy** property

[/geographies](https://github.com/openmobilityfoundation/mds-core/blob/f2e8a81710ad4cd0f7ea59acd3ba0425bed512d4/packages/mds-geography/api.ts#L75)
```js
        return res.status(200).send({ version: res.locals.version, data: { geographies } })
```
=> Would produce expected result according to docs

[/geographies/{id}](https://github.com/openmobilityfoundation/mds-core/blob/f2e8a81710ad4cd0f7ea59acd3ba0425bed512d4/packages/mds-geography/api.ts#L34)
```js
        return res.status(200).send({ version: res.locals.version, data: { geography } })
```
=> Would produce something not expected or documentet, the data object would contain a **geography** property

Thank you

avatarneil:
## Explain pull request

This PR is to resolve #599, which highlighted inconsistencies between the spec and JSONSchema definitions & examples. The approach taken was to leave the spec as-is (for now) in order to be non-breaking, and to update the JSONSchema definitions & examples to align with the spec in its current state.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`

## Additional context

For MDS 2.0, we may want to explore structuring this in accordance with the JSONAPI spec, e.g.
```
{
  "data": {
    "type": "policies",
    "attributes": {
     }
  }
},
```

### DIFF
--- a/policy/examples/idle-time.json
+++ b/policy/examples/idle-time.json
@@ -2,7 +2,7 @@
   "updated": 0,
   "version": "1.0.0",
   "data": {
-    "policy": [
+    "policies": [
       {
         "policy_id": "a2c9a65f-fd85-463e-9564-fc95ea473f7d",
         "name": "Idle Times",
--- a/policy/examples/metered-parking-fees.json
+++ b/policy/examples/metered-parking-fees.json
@@ -2,7 +2,7 @@
   "updated": 0,
   "version": "1.0.0",
   "data": {
-    "policy": [
+    "policies": [
       {
         "policy_id": "6a3dd008-836a-11ea-bc55-0242ac130003",
         "published_date": 1586736000000,
--- a/policy/examples/per-trip-fees.json
+++ b/policy/examples/per-trip-fees.json
@@ -2,7 +2,7 @@
   "updated": 0,
   "version": "1.0.0",
   "data": {
-    "policy": [
+    "policies": [
       {
         "policy_id": "d2567b3c-3071-48a6-bbeb-3424721dbd12",
         "published_date": 1586736000000,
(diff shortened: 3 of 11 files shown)


schnuerle:
Per our [working group meeting](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2021.09.09-(MDS-Working-Group)#action-items) this update was agreed upon. We will merge to dev by the end of the week.

PR #699, opened 2021-09-20
Support: Schema fixes for MDS 1.1.0

schnuerle:
## Explain pull request

Schema fixes for MDS 1.1.0. This will create a new release called 1.1.1.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `provider`


davenguyen (approved):
The only nitpicking I see is to update the `examples` right above the pattern. Otherwise, LGTM!

Issue #668, opened 2021-07-26
Policy API: clarify meaning of `start_date` and `end_date` parameters


jrheard:
### Is your feature request related to a problem? Please describe.

I'm confused by the wording of [the documentation of the /policies endpoint's query parameters](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy#policies). It says:

* that the meaning of `start_date` is "Earliest effective date; default is policies effective as of the request time"
* that the meaning of `end_date` is "Latest effective date; default is all policies effective in the future"

What's the meaning of "effective date" in this context? Is it the policy's start date, or is it referring to the "effective date _range_", i.e. the entire period of time during which the policy was active?

Let's say that a particular `/policies` endpoint is backed by a database containing just one policy, and that the policy has a `.start_date` of January 1 2021 and an `.end_date` of January 30 2021.

How would these `start_date` query parameter values affect the endpoint's response?

* December 1 2020
* January 5 2021
* February 10 2021

How would these `end_date` query parameter values affect the endpoint's repsonse?

* December 1 2020
* January 5 2021
* February 10 2021

I originally thought that the query parameters would be used to filter for all policies that were active during any point between the specified `start_date` and `end_date`, but on a closer read I'm beginning to think that the parameters are actually meant to filter for policies whose _start date_ is within the specified range, because [Geography.effective_date](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy#geography) explicitly refers to `Policy.start_date` (policies don't have a documented concept of an "effective date", Geography.effective_date was the closest thing I could find).

But then I go back to the wording of `end_date`, which is documented as "Latest effective date; default is all policies effective _in the future_", and it seems unlikely to me that the default behavior of the /policies endpoint is to only return policies whose `.start_date` is in the future, so I don't know. Now I think that the documentation probably means "effective date range" after all.

### Describe the solution you'd like

I think that the README should be reworded so that the behavior of these query parameters is clear.

### Is this a breaking change

No

### Impacted Spec

For which spec is this feature being requested?

 * `policy`

janedotx:
[...]

PR #742, opened 2022-01-20
Updating policy query parameter language to be clearer.

janedotx:
## Explain pull request

This is a response to this issue: https://github.com/openmobilityfoundation/mobility-data-specification/issues/668

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`


jrheard (approved):
This is great, thanks very much for doing this!

schnuerle:
@jean-populus what do you think of this suggested wording change?

jean-populus:
looks great @janedotx !

PR #744, opened 2022-02-16
Fixes to Requirements documentation

quicklywilliam:
Fixes metadata field which is not an array, removes required_data_specs and required_apis fields which are not at this level of the JSON.


schnuerle:
Thank you for these fixes.  I will review and merge to dev when ready.

schnuerle (approved):
Agree this change is good and a good catch/cleanup!

PR #745, opened 2022-02-17
Refactoring MDS in preparation for new modes

marie-x:
Refactoring MDS in preparation for new modes


modes/README.md
```diff
@@ -1 +1,88 @@
...
+* [Unique Mode Properties](#unique-mode-properties)
+
+## Modes
+
+The `mode` value is used to specify the applicable mobility category in MDS Policy, MDS Jurisdictions, and other parts of the spec. The only currently supported mode is `micromobility`, but others such as taxi and delivery robots are planned.
+
+A `mode` is defined as: A distinct regulatory framework for a type of mobility service, as distinguished by a combination of a) the data needed by regulators, b) the operating rules under which the service functions, c) the legal authority under which it is regulated, and the d) design and operating model of the service itself.
+
+There will be some gray areas and some differences from one jurisdiction to another (e.g. taxis and ridehail may be regulated under the same rules on one place, but different rules in another). We do not need to pre-define a complete taxonomy of modes, or identify every modal boundary upfront, but would instead add modes on an as-needed basis, maintaining as much consistency of naming as possible.
+
```
sanjivnanda, line 21: I think this was the source of my confusion on the call today. I was trying to think through whether there were rules that applied across modes. But, then I convinced myself that there are, and so it makes sense to me the way you are handling it in Policy.

schnuerle (approved):

PR #760, opened 2022-04-22
Add start_location and end_location to Trips

alexdemisch:
## Explain pull request

SFMTA collects MDS trip data from its E-Moped Parking Permit Program, but only requires the start and end points of each trip, and specifically disallows the intermediate points that describe the routing of each trip.  This is because we don't currently have a need for this data.  SFMTA expresses this requirement in its permit language by saying "the route field shall only contain the start and end points of the trip."  

The Policy Requirements endpoint allows agencies to indicate which fields are required and disallowed.  In the Trips endpoint, the start and end points/time are described by the start_time, end_time, and route fields.  If the route field is disallowed, the start_time and end_time fields still describe the times that trips started and ended, but the ability to communicate location is removed. 

Adding start_location and end_location as optional fields would preserve the ability to communicate start and end points/times without having the route field.

## Is this a breaking change

* No, not breaking

If the new fields are optional, then I don't think this is breaking.

## Impacted Spec

* `provider`


schnuerle:
This is something we discussed as a limitation with how Policy Requirements worked in 1.2 (because of how MDS fields are structured). An agency either had to specify that they wanted the start and end locations of a trip AND the entire trip route, or none of that. There is no way to eliminate the route data but still get start and end O/D points. 

I like this solution since it adds two fields that, while redundant with what is in the route array, clarifies what the start and end points are. This not only makes it easier to allow/disallow the data more granularly with Requirements (for privacy, data minimization, or certain program needs), it also makes data analysis easier since those special points are called out higher than the route level.

schnuerle:
I made some minor clarifications, and I think this is ready for WG review now. It will be important to add this to dev if ready to align with Agency/Provider Unification work. Thanks!

schnuerle:
Thank you for your work here! Due to a technical issue around the signing of our license agreement by a GitHub user who's account/email is changed, I am adding this change here to our latest feature branch with [this commit](https://github.com/openmobilityfoundation/mobility-data-specification/commit/2f9d58a50d99b992817a8d82bde9d42e6c5ff8b5).

Issue #748, opened 2022-02-25
Trip definitions


jean-populus:
### Is your feature request related to a problem? Please describe.

Trip counts are an important metric for cities but the data we receive through MDS doesn't reflect the reality on the ground. For example, short trips that are just a few meters, phantom trips, trips that leave the city boundary and disappear. We can't use the data as-is and have to make judgement call on which trips to count and which trips to filter.

### Describe the solution you'd like

It would be helpful if there was a shared understanding and recommendation on what counts as a Trip. Currently each city/provider/3rd party is a separate conversation and unique implementation.

For example, we could only count trips with an [actual_cost](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/provider/README.md#trips).

### Is this a breaking change

 * I'm not sure

### Impacted Spec

 * `policy`

### Describe alternatives you've considered

n/a

### Additional context

see [2022.02.17 working group discussion](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2022.02.17-(MDS-Working-Group)#general-discussion-on-trip-definitions)

ezmckinn:
Agreed that this is an important thing to clarify from an operator perspective, as well.   Some questions for us to consider on the next working group call: 

* What are valid reasons for a trip to be omitted (e.g. distance, time), vs. included and flagged as a valid edge case?  

* Do we omit linked `trip_start` and `trip_end` events, as well? 
    * I’d suggest “yes”, because this will keep trip counts aligned between /status_changes and /trips, and also avoid having to re-tool the state machine. However, this could be incompatible with Agency implementations that require trip start/end events to be reported within seconds of the user attempt to start.  
    
* Does MDS Policy need to be updated in any way to accommodate this definition? Might that even be the best place to define what will be counted?

schnuerle:
Discussion today about adding more clarifying information to the spec around trip definition and types of 'trips' in MDS.

Could go in [Definitions](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/general-information.md#definitions) and/or [provider trips](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/provider#trips).

Currently says: 

> A trip represents a journey taken by a mobility as a service customer with a geo-tagged start and stop point.

Could be more like:

[...]

PR #762, opened 2022-04-28
Define a "trip" for the purposes of analyses

jiffyclub:
## Explain pull request

This addresses #748 by adding a definition of a "trip" to the general-information.md file. The definition is meant for the purposes of analysis in particular, and not for filtering which data points are exchanged as part of MDS endpoints.

I wasn't sure on the exact right spot for this since I hadn't seen anything like this done before in MDS, but I'm open to suggestions on that and the wording of the definition itself.

## Is this a breaking change

* No, not breaking

I don't consider this a breaking change since it doesn't effect the data interchange at all, only downstream analytics, but I might be missing something.

## Impacted Spec

* `agency`
* `policy`
* `provider`

In a way, all of these and none of these. 🤔

### DIFF
--- a/general-information.md
+++ b/general-information.md
@@ -22,6 +22,7 @@ This document contains specifications that are shared between the various MDS [A
   - [GBFS Compatibility](#gbfs-compatibility)
 - [Timestamps](#timestamps)
 - [UUIDs](#uuids)
+- [Trips](#trips)
 - [Vehicle Characteristics](#vehicle-characteristics)
   - [Accessibility Options](#accessibility-options)
   - [Propulsion Types](#propulsion-types)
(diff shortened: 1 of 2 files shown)


general-information.md
```diff
@@ -177,6 +178,25 @@ Note if implementing the beta feature [Geography Driven Events](/general-informa
...
+as a user taking control of a vehicle for 60 seconds or more.
+If a user reserves a vehicle and cancels before taking control of the vehicle
+that reservation does not count as a trip, nor does a user taking control of
+the vehicle and canceling or ending the trip in under 60 seconds.
+
+Providers are still expected to report these events in MDS endpoints,
+but parties may use this definition as a shared reference at the
+recommendation of the MDS community.
+
+For reference on the development of this definition please refer to
```
schnuerle, line 196: I think you can cut the link to this Issue. Instead we may add a new Issue or Discussion area that can be referenced here for people to leave feedback.

jiffyclub:
Follow-up from the MDS working group meeting on 2022-04-28, I'll make the language about what providers should report more definitive and reference _all_ MDS endpoints. I'll also link from other places that discuss trips to this definition.

jean-populus (approved):
Looks good to me. Thanks @jiffyclub !

jiffyclub:
Made the update clarifying that providers should still send all info in all cases and made a link from provider trips to this definition.

PR #776, opened 2022-07-06
Update agencies.csv to include SFMTA requirements URL

alexdemisch:
## Explain pull request

Update agencies.csv to include link SFMTA requirements file and change the agency name from "San Francisco" to "San Francisco Municipal Transportation Agency.

## Is this a breaking change

* No, not breaking


schnuerle (approved):
Thanks and you are the second agency to publish their Requirements!

Issue #672, opened 2021-07-30
Provider Reports - Beta 1.1.0 - Feedback


schnuerle:
Gather feedback to see how we can move [Provider Reports](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/provider#reports) out of beta. 

### Describe the solution you'd like

Collect enough feedback, real-world usage, and requested updates/improvements to move this feature out of beta in a future release

### Is this a breaking change

 * Not sure yet

### Impacted Spec

For which spec is this feature being requested?

 * Provider

### Describe alternatives you've considered

Leave as a beta feature for longer until we have feedback. 

### Additional context

All beta features once part of an official release receive an Issue like this to gather feedback.

ViennaMike:
Likely minor: The current definition states "The authenticated reports are monthly, historic flat files that may be pre-generated by the provider."  Is there a reason these are specified as monthly? Could / should this not be variable, such as weekly, bi-monthly, etc.?

wellorder:
- It's a bit strange that the Accept header for this endpoint is still something like `Accept: application/vnd.mds+json;version=1.2.0`, since this endpoint does not return JSON.  If I understand correctly, `+csv` would not be valid since `csv` is not part of the [IANA list of acceptable suffixes](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xml), so I believe the "correct" thing would be to have no `+foo` suffix at all.  (Accept headers are not my area of expertise though!)  This would mean this is the only `/provider` endpoint that has a different media type, but that would reflect the actual response type better.
- `Counts are calculated based the agency's local time zone, and this time zone is returned within the StartDate value. For months where there is a Daylight Saving Time change, use the timezone that is in the majority of the month.` This seems a bit strange to me.  This would mean our counts would say timezones that did not apply to many of the rides aggregated.  It also means we'll have StartDates listed with timezones they didn't actually have (the first of the month often has a different timezone from the majority of the month when Daylight Savings occurs).  I guess the thinking is this is a double-check that we are returning results based on local timezones, but fundamentally there can be multiple timezones in a given month so a monthly report is going to have a hard time conveying that as a single piece of information.  It's not clear to me that it's worth the confusion, I doubt cities will be checking that we put the "right" timezone string in that field.  I'm not a city though, maybe I'm not thinking of some use case.
[...]

PR #804, opened 2022-11-18
Reports - Add provider_id and adaptive_scooter as a special group type

alexdemisch:
## Explain pull request

Following the discussion and [meeting earlier this year](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2022.07.07-(MDS-Working-Group)) on #672, add 'adaptive_scooter' as a special_group_type to reflect SFMTA's use of this in the Reports endpoint for its [scooter program](https://www.sfmta.com/blog/adaptive-scooters-coming-street-near-you).

Also, in our use of Reports, we've found that adding provider_id is necessary so we know which company the report is from.  It would be better if the providers included this directly. 

## Is this a breaking change

* Yes, breaking

## Impacted Spec

* `provider`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -365,6 +365,7 @@ The authenticated reports are monthly, historic flat files that may be pre-gener
 
 | Name               | Type                                      | Comments                                         |
 | ------------------ | ----------------------------------------- | ------------------------------------------------ |
+| provider_id      | UUID                                      | A UUID for the Provider, unique within MDS. See MDS [provider list](/providers.csv). |
 | StartDate          | date                                      | Start date of trip the data row, ISO 8601 format, local timezone |
 | Duration           | string                                    | Value is always `P1M` for monthly. Based on [ISO 8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations) |
 | Special Group Type | [Special Group Type](#special-group-type) | Type that applies to this row                    |
--- a/provider/README.md
+++ b/provider/README.md
@@ -450,6 +450,7 @@ Here are the possible values for the `special_group_type` dimension field:
 | Name       | Description                                                                                                           |	
 | ---------- | --------------------------------------------------------------------------------------------------------------------- |	
 | low_income | Trips where a low income discount is applied by the provider, e.g., a discount from a qualified provider equity plan. |	
+| adaptive_scooter | Trips taken on an adaptive scooter, e.g., scooter with a seat or wider base |
 | all_riders | All riders from any group                                                                                             |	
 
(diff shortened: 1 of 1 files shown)


schnuerle:
We had provider_id initially, then removed it thinking that you'd know the provider ID from the http link to the report itself coming from a specific provider! Are you saying you'd still like to see this in the CSV file because it's easier, or is there a scenario where one provider could be sending multiple provider_id in the same file (which may be the case with taxis or other modes/scenarios).

The adaptive_scooter option seems good. I'm wondering if there is a better defintion of this though or a link to some sort of canonical description/page since I think some people (in Europe in particular) aren't sure what this is. Is there something to note about this being 'handicap accessible'?

alexdemisch:
> We had provider_id initially, then removed it thinking that you'd know the provider ID from the http link to the report itself coming from a specific provider! Are you saying you'd still like to see this in the CSV file because it's easier, or is there a scenario where one provider could be sending multiple provider_id in the same file (which may be the case with taxis or other modes/scenarios).

Yes, you're right that we'd know which provider each report is coming from.  It's just easier to also have it in the data since it saves a step downstream when you're combining the data.  But we could leave it as it is.

> The adaptive_scooter option seems good. I'm wondering if there is a better defintion of this though or a link to some sort of canonical description/page since I think some people (in Europe in particular) aren't sure what this is. Is there something to note about this being 'handicap accessible'?

Hmm I don't know of anything canonical.  There's been some work done on the GBFS side to describe all the different types of vehicles based on their characteristics, hasn't there?  Maybe there's something we could use from that work?  Otherwise, maybe we could modify the description to say something like "Trips taken on a scooter with features to improve accessibility for people with disabilities, e.g., scooter with a seat or wider base"

schnuerle:
Those ideas make sense to me. I modified your PR to reflect them, and added the UUID to the examples.

schnuerle:
What do you think about #809 for removing both `start_date` and `duration` from reports?

alexdemisch:
> What do you think about #809 for removing both `start_date` and `duration` from reports?

I think we should keep both concepts somehow.  The current need for micromobility reports is monthly, but other frequencies could come up, especially for other modes as noted in the comments for #672.  I understand the complexity around UTC and timestamps for `start_date`... is there some other way to address it?  I know the date period is described in the path, but like we're doing with provider_id, it'd be good if the date range is also reflected in the data somehow.

wellorder:
I'm fine putting the the date string, like `2022-11-01`, but putting a time with offset is fundamentally misguided (there are months with more than one UTC offset).  Another possibility would be to include the time zone as its own column, rather than the offset.  (That was actually in the initial commit for that PR :) ) Then you have two columns with identical values in each report, and you're reporting to the city their own time zone (which, uh, they know), but I'm not against it.  (One could consider it evidence that the provider has followed the spec with regard to counting in the local time zone I suppose.) I _am_ against the UTC offset.

This discussion really should be a part of the other PR, so I will be including it there.

I have no strong feelings regarding including the `provider_id` or not.

PR #807, opened 2022-12-07
`/reports` column names

wellorder:
## Explain pull request

The `/reports` spec has column names in an inconsistent format.  `StartDate` is CamelCase, but all other column names are plain English.  We use snake case elsewhere, and even do so farther down in the `/reports` spec when discussing `special_group_type`, so I went with that.

## Is this a breaking change

Technically this is a breaking change, but the `/reports` endpoint is still beta so it's not clear how big of a deal that is.  That said, the goal would be for this change to be incorporated into 2.0, which allows for breaking changes anyway.

## Impacted Spec

* `provider`

## Additional context

N/A


schnuerle:
Love it, thank you!

PR #810, opened 2022-12-07
Reports - Accept header media-type

wellorder:
## Explain pull request

The spec currently requires the `/reports` endpoint to be queried with media-type `application/vnd.mds+json` in the `Accept` header.  But the endpoint returns a CSV, not JSON, so this is a bit confusing.  This PR suggests a different approach (though it has its own drawback since `csv` is not a standard media-type suffix).

## Is this a breaking change

This is a breaking change.  The intent is for this to be included in MDS 2.0.

## Impacted Spec

* `provider`

## Additional context

See the discussion from https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2022.07.07-(MDS-Working-Group) and the discussion at https://github.com/openmobilityfoundation/mobility-data-specification/issues/672

### DIFF
--- a/general-information.md
+++ b/general-information.md
@@ -351,7 +351,7 @@ The list of allowed `vehicle_type` values in MDS. Aligning with [GBFS vehicle ty
 
 MDS APIs must handle requests for specific versions of the specification from clients.
 
-Versioning must be implemented through the use of a custom media-type, `application/vnd.mds+json`, combined with a required `version` parameter.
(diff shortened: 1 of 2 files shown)


schnuerle:
I think the only reason not to do this is like you said:

> (`csv` is not a [standard media-type suffix](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xml)

Some including myself wanted to use this anyway, but there was some concern for not using a standard media-type suffix. I would be open to accepting your changes.

wellorder:
It had been long enough that I had forgotten parts of the [working group discussion](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2022.07.07-(MDS-Working-Group)) we had, but the notes have as an action item
```
add csv in reports header with note saying we know it’s not in the acceptable list
```
so I went with that here, and am fine with it as a solution.  I'd also be open to just not using a suffix, since `csv` is not on the list.  Using `json` for non-`json` data definitely sits poorly with me though.

schnuerle:
Can we just use `application/vnd.mds`? Is that valid?

schnuerle:
Agree and maybe @marie-x @avatarneil or @thekaveman could help with how to hand csv output.

PR #817, opened 2023-01-11
Remove 'rate' rule-type, since we can put rate_amount on any rule

### LINKED ISSUE #662: Remove "rate" as a rule type

I'm having trouble wrapping my head around the idea of "rate" rule. Rates don't exist in a vacuum, they apply to something else and that something else is more at the heart of the rule. And as we're seeing in #631/#658, we've ended up wanting rates in conjunction with other rule types, not as a standalone rule.

What if we got rid of the rate rule entirely and only had rule types based around the physical concept to which the rate applies? The policies in #658 are `time` rules. So is [this parking fee example](https://github.com/openmobilityfoundation/mobility-data-specification/blob/7dca90a65a53491a4f5c06696ecfcb8fcbb1261b/policy/examples/README.md#metered-parking-fees). I think a [per trip fee](https://github.com/openmobilityfoundation/mobility-data-specification/blob/7dca90a65a53491a4f5c06696ecfcb8fcbb1261b/policy/examples/README.md#per-trip-fees) could be a `count` rule, or maybe a new `per_unit` rule.

Rules could then have a `rate`/`fee`/`cost` (open to suggestions) section that describes anything monetary associated with the rule. This would be basically the same as the rate info that's currently in rules, but could also contain a field that describes whether the cost is meant to be applied to events in compliance with the rule or events in violation of the rule, clearing up some confusion we've discussed in relation to #658 (see #666). (I'd like this monetary section to be a sub-section of the rule, not at the top level as it is now, but that's a separate discussion I'll open an issue about. See #663.)

### Is this a breaking change

 * Yes, breaking

### Impacted Spec

`policy`

There are [example policies](https://github.com/openmobilityfoundation/mobility-data-specification/blob/7dca90a65a53491a4f5c06696ecfcb8fcbb1261b/policy/examples/README.md) and #658 that mix rates with other types of rules while using the `rate` rule type, so we could stick with this existing system where only rules of the `rate` type have monetary costs (or subsidies) associated. But, as outlined here and in #631, I don't think that's working well.

### Additional context

N/A

### LINKED ISSUE #674: Policy Rule Type: Rates - Beta 1.0.0 - Feedback
Gather feedback to see how we can move [Policy Rule Type: Rates](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy#rule-types) out of beta. 

Collect enough feedback, real-world usage, and requested updates/improvements to move this feature out of beta in a future release

### Is this a breaking change

 * Not sure yet

### Impacted Spec

 * Policy

Leave as a beta feature for longer until we have feedback. 

### Additional context

All beta features once part of an official release receive an Issue like this to gather feedback.

### LINKED ISSUE #785: Proposal: Adopt policy rate structure from CDS

In MDS at the moment, if you have a tiered parking fees policy (e.g. $10 / hour for the first hour, $5 / hour for hours 2 - 5, $3 / hour for hours 6 onward) you have to create multiple rules in your policy for each of those different time ranges (see [examples](https://github.com/openmobilityfoundation/mobility-data-specification/tree/26705816a4dd23d23a41cbd24d99948d2674b6f0/policy/examples#tiered-parking-fees-per-hour)). This has the weird of effect of making it so that different rules in the policy match the same time and space only for the purpose of describing what happens based on the duration of the event.

I think it's more understandable to be able to have a single rule that matches an event and to be able to reference the monetary consequence for that from the one rule instead of it depending on aggregating monetary consequences from multiple other rules. Indeed, that's what CDS does, allowing a [rule](https://github.com/openmobilityfoundation/curb-data-specification/tree/d0de87b1bd6c1848f6d5ca06a8686c116764141f/curbs#rule) to contain an array of [rates](https://github.com/openmobilityfoundation/curb-data-specification/tree/d0de87b1bd6c1848f6d5ca06a8686c116764141f/curbs#rate) so that it's the different rate objects that apply to different parts of the event instead of different rules. When evaluating the policy you can find that an event (or portion of an event) matches the rule and directly reference the rate structure to find the monetary consequence of that match.

My proposal is that MDS adopt the Rate object from CDS and allow rules of any type to contain an array of Rates, the same way that CDS does. Even outside of my specific complaint about tiered pricing structures in MDS, I think it'd be best if MDS and CDS had the same structures around this to make it easier to evaluate policies in both MDS and CDS. (This also ties into #662, I think we should remove the "rate" rule type and allow attaching rates to any rule.)

### Is this a breaking change

 * Yes, breaking

### Impacted Spec

 * `policy`

None

### Additional context

None

marie-x:
## Explain pull request

Remove 'rate' rule-type, since we can put rate_amount on any rule

## Is this a breaking change

* Yes, breaking

## Impacted Spec

* `policy`

## Additional context

I forget the ticket number

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -267,7 +267,7 @@ An individual `Policy` object is defined by the following fields:
 | `policy_id`      | UUID            | Required   | Unique ID of policy                                                                 |
 | `provider_ids`   | UUID[]          | Optional   | Providers for whom this policy is applicable; empty arrays and `null`/absent implies all Providers. See MDS [provider list](/providers.csv). |
 | `description`    | String          | Required   | Description of policy                                                               |
-| `currency`       | String          | Optional   | An ISO 4217 Alphabetic Currency Code representing the [currency](../general-information.md#costs-and-currencies) of all Rules of [type](#rule-types) `rate`.|
+| `currency`       | String          | Optional   | An ISO 4217 Alphabetic Currency Code representing the [currency](../general-information.md#costs-and-currencies) of all Rules with a `rate_amount`.|
 | `start_date`     | [timestamp][ts] | Required   | Beginning date/time of policy enforcement. In order to give providers sufficient time to poll, `start_date` must be at least 20 minutes after `published_date`.                                           |
 | `end_date`       | [timestamp][ts] | Optional    | End date/time of policy enforcement                                                 |
 | `published_date` | [timestamp][ts] | Required   | Timestamp that the policy was published                                             |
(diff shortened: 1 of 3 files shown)


schnuerle:
I think this looks good. The important file to look at is the [policy readme](https://github.com/openmobilityfoundation/mobility-data-specification/pull/817/files#diff-5dbb467da8c3ceb340b4c471496e28b63ed45d1ee3e15ceed7c44c3b88340720). Would like to have @jean-populus @S-eb take a look.

PR #878, opened 2023-09-08
Update "status changes" to "events" in provider

jiffyclub:
## Explain pull request

I noticed a few places in the provider README where it was still referring to "status changes" instead of the new term "events". There was also an inconsistency in the depiction of the response format.

## Is this a breaking change

* No, not breaking.

I think this is purely a documentation update to make it update the defined schema.

## Impacted Spec

* `provider`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -395,7 +395,7 @@ The `/events/historical` API uses the following query parameter:
 
 | Query Parameter    | Format | Expected Output |
 | ---------    | ------ | --------------- |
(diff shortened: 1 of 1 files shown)


schnuerle:
Thank you I think this is a good catch and something we can put in a patch release.

schnuerle:
We are looking at creating a 2.0.1 release soon for all open issues and PRs tagged with the [2.0.1 Milestone](https://github.com/openmobilityfoundation/mobility-data-specification/milestone/16). 

Any feedback from the community is welcome before merging to dev.

schnuerle (approved):

PR #880, opened 2023-10-16
Remove word "data" from provider response payload descriptions

jiffyclub:
# MDS Pull Request

## Explain pull request

In older versions of MDS data was nested under a `data` key in provider API responses, so trips data would e.g. be response['data']['trips']. This ended up reflected in the docs where each section would have a label like "`data` Payload" describing what was under the `data` key in the payload.

In MDS 2.0 this nesting was removed, `trips` is now a top level key in the response from provider trips endpoint. However, the docs were not updated to remove the word "data" from the payload descriptions, leading to potential confusion for readers of the documentation. This PR changes "`data` Payload" to "Payload" throughout the provider README to bring it in line with the spec.

## Is this a breaking change

* No, not breaking. This is a documentation only change that brings the docs more in line with the spec.

## Impacted Spec

* `provider`

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -189,7 +189,7 @@ The `/vehicles` endpoint returns the specified vehicle (if a device_id is provid
 **Method:** `GET`  
 **[Beta feature][beta]:** No (as of 1.2.0)  
 **Schema:** See [`mds-openapi`](https://github.com/openmobilityfoundation/mds-openapi) repository for schema.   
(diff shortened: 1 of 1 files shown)


schnuerle:
We are looking at creating a 2.0.1 release soon for all open issues and PRs tagged with the [2.0.1 Milestone](https://github.com/openmobilityfoundation/mobility-data-specification/milestone/16). 

Any feedback from the community is welcome before merging to dev.

Issue #868, opened 2023-05-22
Correction of description vehicle events (car sharing)


bergenklem:
Of what we can see, the table for [vehicle events in [car sharing mode](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/modes/car-share.md#vehicle-states-events) needs an update, as they right now points towards actions that are more in relation to taxi services (and/or autonomous vehicles) than a car sharing operation as of today.

![event types description](https://github.com/openmobilityfoundation/mobility-data-specification/assets/134277360/39cb85a2-e656-4052-8ff3-13296a4b02ec)

schnuerle:
Yes this looks like just a carry over typo. So this can be fixed in the next release, maybe a 2.0.1 [patch](https://github.com/openmobilityfoundation/governance/blob/main/technical/ReleaseGuidelines.md#versioning) release.

schnuerle:
If you can make a PR for this change, that would be helpful, and you'd be listed as a GitHub contributor.  We are looking at creating a 2.0.1 release soon for all open issues and PRs tagged with the [2.0.1 Milestone](https://github.com/openmobilityfoundation/mobility-data-specification/milestone/16). If not I will create the PR.

schnuerle:
Closed with #881

PR #881, opened 2023-10-17
Update to pairing list of vehicle_state and event_type for car sharing

bergenklem:
## Explain pull request

The pairing list for _vehicle_state_ and _event_types_ for car sharing has several descriptions more aligned with a taxi service than a car sharing operation. 

This draft PR is a follow-up of [Issue 868 ](https://github.com/openmobilityfoundation/mobility-data-specification/issues/868), and proposes minor updates to the descriptions so that they stay more in line with how car sharing works. 

This draft PR also includes two discussion points:

**1. Describing the event where a vehicle_state goes from _reserved_ to _stopped_.**

Original description is as follows: 

_| `reserved`               | `stopped`              | `reservation_stop`           | The vehicle has stopped to pick up the passenger                                                                |_

In a car-sharing operation, the car would already be parked - ready for the next customer to pick it up.

I've proposed to change the description to be "the customer has activated the vehicle". An activation is typically initiated through the providers app/RFID-chip/by provider (in cases where you've lost your key-card for example). 

**2. We should discuss the need for the "driver_cancellation" event.**

**Q: what is the difference between a driver and a customer for a car sharing-mode?**
As to my knowledge, in a car sharing operation, the driver = customer (the provider could let several customers/users be drivers in the same booking, but there's never a 3rd party driver involved).
If the driver is not the same as the customer(s), then I would say we are talking about a taxi service/ride hailing.

Hence, I don't think it make sense to distinguish between customer and driver in car sharing.

_my proposal would be to remove driver_cancellation from the list of event_types.

If so, this change would also cause need for an update of the state machine diagram.

## Is this a breaking change

Not breaking

## Impacted Spec

* `agency`
* `provider`


schnuerle:
I believe we left driver in there because there are scenarios where the company may deliver the car to the customer, e.g. have a driver employee/contractor drive the car to the customer's home. So we should leave this in for the 2.0.1 release. 

If we want to discuss and remove it in a future release, it would have to be the 3.0 release as this would be a breaking change. If anyone would like to do this, please open a separate issue.

PR #882, opened 2023-10-18
Clarify which vehicles to include in the `/vehicles` endpoint

### LINKED ISSUE #879: What vehicles are included in the MDS 2.0 /vehicles endpoint?
In MDS 2.0 the `/vehicles` endpoint contains static vehicle information that could be required for vehicles deployed years in the past. However, the [documentation](https://github.com/openmobilityfoundation/mobility-data-specification/tree/2.0.0/provider#vehicles) reads:
> The /vehicles is a near-realtime endpoint and returns the current status of vehicles in an agency's [Jurisdiction](https://github.com/openmobilityfoundation/mobility-data-specification/blob/2.0.0/general-information.md#definitions) and/or area of agency responsibility. All vehicles that are currently in any [vehicle_state](https://github.com/openmobilityfoundation/mobility-data-specification/blob/2.0.0/modes#vehicle-states) should be returned in this payload. Since all states are returned, care should be taken to filter out states not in the [PROW](https://github.com/openmobilityfoundation/mobility-data-specification/blob/2.0.0/general-information.md#definitions) if doing vehicle counts. For the states elsewhere and removed which include vehicles not in the [PROW](https://github.com/openmobilityfoundation/mobility-data-specification/blob/2.0.0/general-information.md#definitions) but provide some operational clarity for agencies, these must only persist in the feed for 90 minutes before being removed.

This is [carryover language from MDS 1.x](https://github.com/openmobilityfoundation/mobility-data-specification/tree/1.2.0/provider#vehicles) and I'm curious what the expectations are for the MDS 2.0 `/vehicles` endpoint (distinct from the `/vehicles/status` endpoint). Should it contain every vehicle ever deployed in a jurisdiction? If it is only expected to contain currently deployed vehicles (like the MDS 1.x `/vehicles` endpoint), what's the procedure for getting vehicle information for historically deployed vehicles? Request it one vehicle at a time via `/vehicles/{device-id}`?

And then thinking about the relationship between MDS 2.0 `/vehicles` and `/vehicles/status`, my instinct is that we should require that at a minimum the `/vehicles` endpoint should always return the same vehicles as would be returned by `/vehicles/status` and document that requirement. But that still leaves the question of how to get vehicle information for historically deployed vehicles.

jiffyclub:
# MDS Pull Request

## Explain pull request

The description of the Provider vehicles endpoints was never really updated in the MDS 1 -> 2 transition so this rearranges and clarifies some things now that there are two vehicle information endpoints, one for static information like vehicle/propulsion type and another for realtime status. These changes are mostly rearranging/clarifying, but there is one 
major addition to the description of the `/vehicles` endpoint describing that when called without a device ID it is expected to return every vehicle ever deployed in a jurisdiction. This closes #879.

## Is this a breaking change

* No, not breaking. This makes no changes to the spec format, only clarifies content expectations.

## Impacted Spec

* `provider`

## Additional context

See also #879.

### DIFF
--- a/provider/README.md
+++ b/provider/README.md
@@ -168,22 +168,19 @@ For Timestamps, Vehicle Types, Propulsion Types, UUIDs, Costs, and Currencies, r
 
 ## Vehicles
 
-The `/vehicles` is a near-realtime endpoint and returns the current status of vehicles in an agency's [Jurisdiction](/general-information.md#definitions) and/or area of agency responsibility. All vehicles that are currently in any [`vehicle_state`][vehicle-states] should be returned in this payload. Since all states are returned, care should be taken to filter out states not in the [PROW](/general-information.md#definitions) if doing vehicle counts. For the states `elsewhere` and `removed` which include vehicles not in the [PROW](/general-information.md#definitions) but provide some operational clarity for agencies, these must only persist in the feed for 90 minutes before being removed. 
+There are two vehicles related endpoints:
 
-As with other MDS APIs, `/vehicles` is intended for use by regulators, not by the general public. `/vehicles` can be deployed by providers as a standalone MDS endpoint for agencies without requiring the use of other endpoints, due to the [modularity](/README.md#modularity) of MDS. See our [MDS Vehicles Guide](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/MDS-Vehicles) for how this compares to GBFS `/free_bike_status`. Note that using authenticated `/vehicles` does not replace the role of a public [GBFS][gbfs] feed in enabling consumer-facing applications. If a provider is using both `/vehicles` and GBFS endpoints, the `/vehicles` endpoint should be considered source of truth regarding an agency's compliance checks.
+- `/vehicles` returns unchanging information about vehicles such as vehicle and propulsion type
+- `/vehicles/status` returns the current status of vehicles for real-time monitoring
 
-In addition to the standard [Provider payload wrapper](#response-format), responses from this endpoint should contain the last update timestamp and amount of time until the next update in accordance with the [Data Latency Requirements][data-latency]:
+As with other MDS APIs, the vehicles endpoints are intended for use by regulators, not by the general public. They can be deployed by providers as standalone MDS endpoints for agencies without requiring the use of other endpoints, due to the [modularity](/README.md#modularity) of MDS. See our [MDS Vehicles Guide](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/MDS-Vehicles) for how this compares to GBFS `/free_bike_status`. Note that using authenticated vehicles endpoints does not replace the role of a public [GBFS][gbfs] feed in enabling consumer-facing applications. If a provider is using both the vehicles endpoints and GBFS endpoints, the vehicles endpoints should be considered source of truth regarding an agency's compliance checks.
 
-```json
-{
-    "version": "x.y.z",
-    "last_updated": "12345",
-    "ttl": "12345",
-    "vehicles": []
-}
-```
+### Vehicle Information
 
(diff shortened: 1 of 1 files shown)


schnuerle:
@jiffyclub do you think you can present some slides to explain this change at this Thursday's working group meeting? If so it can be the main agenda item.

jiffyclub:
Yes, can do!

provider/README.md
```diff
@@ -228,7 +225,21 @@ See [Responses][responses], [Bulk Responses][bulk-responses], and [schema][schem
 
 ### Vehicle Status
 
-The `/vehicles/status` endpoint returns the specified vehicle (if a device_id is provided) or a list of known vehicles. Contains specific vehicle status records that are updated frequently.
+The `/vehicles/status` endpoint is a near-realtime endpoint and returns the current status of vehicles in an agency's [Jurisdiction](/general-information.md#definitions) and/or area of agency responsibility. All vehicles that are currently in any [PROW](/general-information.md#definitions) state [`vehicle_state`][vehicle-states] should be returned in this payload. Since all states are returned, care should be taken to filter out states not in the [PROW](/general-information.md#definitions) if doing vehicle counts. For the states `elsewhere`,  `removed`, and `missing`, which include vehicles not in the [PROW](/general-information.md#definitions) but provide some operational clarity for agencies, these vehicles must only persist in the feed for 90 minutes before being removed (and should persist in the feed for at least 90 minutes).
```
wellorder, line 228: "these vehicles must only persist in the feed for 90 minutes before being removed (and should persist in the feed for at least 90 minutes)" is the same as "these vehicles must persist in the feed for exactly 90 minutes", which sounds a bit strict, honestly.  (I suppose the existing language comes across this way as well, but less explicitly.)

I really don't want to be called out-of-compliance because a vehicle was present for 91 minutes because some backend process got delayed by 1 minute.  Technically enforcing such a TTL is not much more difficult than one with a grace period, but maybe allowing for things to be present for, say, up to 120 minutes will save some annoying discussions.  I can't imagine what logic a consumer might have that would be thrown off by such a grace period (maybe someone will think of something though).

provider/README.md
```diff
@@ -168,22 +168,19 @@ For Timestamps, Vehicle Types, Propulsion Types, UUIDs, Costs, and Currencies, r
...
-    "ttl": "12345",
-    "vehicles": []
-}
-```
+### Vehicle Information
 
-The `/vehicles` endpoint returns the specified vehicle (if a device_id is provided) or a list of known vehicles. Contains vehicle properties that do not change often.
+The `/vehicles` endpoint returns the specified vehicle (if a device_id is provided) or a list of known vehicles.
+It contains vehicle properties that do not change often.
+When `/vehicles` is called without specifying a device ID it should return every vehicle that has
```
wellorder, line 182: "every vehicle that has ever been deployed in an agency's jurisdiction" can be much larger than "every vehicle currently present (in or out of PROW) in an agency's jurisdiction".  Vehicles get lost, vehicles get transferred to other cities.  This is especially true for cities where vehicles are initially shipped to and they undergo some sort of QA before being moved along to their final destination.

The longer a provider has been present in a city the bigger a discrepancy this will be.  We currently return all vehicles that have been present in the last 21 days; I could see arguments for some longer period but having an unbounded period leads to unbounded amounts of data, and to what end?  How useful will a consumer find data about a vehicle that hasn't been seen in 2 years?  They almost certainly retrieved that data a year or two ago anyways.

jiffyclub:
Based on discussion on [today's MDS call](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2023.12.07-(MDS-Working-Group)) I'll update this to specify:
- `/vehicles` must return all vehicles deployed in a region in the last 30 days
- `/vehicles/{device_id}` must return vehicle information for all device IDs present in MDS data available from other endpoints regardless of when the device was deployed. If the vehicle information has changed over time the most recent information should be returned.

PR #918, opened 2024-08-30
Add CDS Rate object to MDS

### LINKED ISSUE #785: Proposal: Adopt policy rate structure from CDS

jiffyclub:
## Explain pull request

When developing CDS we came up with a more expansive definition of rates [1] to better capture real life scenarios. This commit brings that same definition to MDS, but keeps the MDS field rate_applies_when for specifying when the rate applies to an event.

As in CDS, a policy rule may now contain an array of rates specifying monetary charges or subsidies for events that match that rule. This removes the need to create multiple rules on a policy to capture different parts of the rate calculation, it can all be contained in the rate array.

[1]: https://github.com/openmobilityfoundation/curb-data-specification/tree/1.0.0/curbs#rate

## Is this a breaking change

* Yes, breaking

## Impacted Spec

* `policy`

## Additional context

Closes #785.

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -31,9 +31,6 @@ This specification describes the digital relationship between _mobility as a ser
   - [Rule Types](#rule-types)
   - [Rule Units](#rule-units)
   - [Rates](#rates)
-    - [Rate Amounts](#rate-amounts)
-    - [Rate Recurrences](#rate-recurrences)
-    - [Rate Applies When](#rate-applies-when)
   - [Messages](#messages)
   - [Value URL](#value-url)
   - [Order of Operations](#order-of-operations)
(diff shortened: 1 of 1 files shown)


schnuerle:
I like this proposal and think it's needed. But will have to wait for a 3.0 release since it's breaking. At the moment we are focused on 2.1.

PR #928, opened 2025-03-11
Real-time Policy Improvements and Clarifications

### LINKED ISSUE #906: Real-time support for MDS Policy

While MDS Policy [can be real-time](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy#update-frequency) if required by the city 1) the spec doesn't explicitly say this and 2) there is not a good way to get the latest policy because the entire policy has to be downloaded to be checked for an update.

Update the frequency language to state that real-time is possible, and set a recommended minimum threshold (eg, every 1 minute at most).

Add a new `last_updated` boolean parameter to the **/policies** [endpoint](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy#update-frequency) that returns only the `version`, `last_updated`, `end_date` fields, like so:

```
{
  "version": "2.0.0",
  "last_updated": 1570035222868,
  "end_date": 1570035222868
}
```

This creates a quick check option and small file size to allow an operator to check for any update in the file easily and regularly.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `policy`

There has also been talk of making this endpoint a push endpoint option, in addition to the existing pull option, like with do with Agency and Provider for operator data. But this complexity may not be needed if the recommended solution above is adequate.

### Additional context

This was discussed a few times at OMF member steering committee meetings, and at public working groups, in the context of supporting real-time emergency response using Policy for AV robotaxis to support #905

schnuerle:
## Explain pull request

Updating the MDS spec to make it clearer and easier to use Policy for real-time management of the public right of way. 

Currently includes the following changes:

1. Real-time language clarification and examples
2. Noting that agencies may manage their public Policy feeds, for example with an API key
3. Policy uniqueness note to prevent identical policy duplication
4. last_updated parameter to return only the last updated date
5. active_only parameter to return only current, active policies
6. A Policy object relationship diagram.
7. Additional relevant policy [examples](https://github.com/openmobilityfoundation/mobility-data-specification/blob/main/policy/examples/README.md) (moved these to an external document for ease of updates)

To possibly come in the future, in new Pull Requests:

1. A push endpoint option, like using POST in [MDS Agency](https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/agency), and related to [CDS push Events](https://github.com/openmobilityfoundation/curb-data-specification/pull/180) (PR #951)
2. A new Message/Notification/Communication endpoint/API (may come in a separate PR)

Likely not coming based on discussion in meeting:

1. Adding a `max_update_interval` field like in Requirements to note how frequently the file may update (some discussion in WG that this may not be a good idea)

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`

## Additional context

See Issue #906 for more history and context.


policy/README.md
```diff
@@ -78,14 +80,16 @@ See the [Policy Examples](./examples/README.md) for ways Policy can be implement
 
 ### Authorization
 
-The Policy endpoints should be made public. Authorization is not required.
+The Policy endpoints should be made public. Authorization is not required. Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain.
```
geneorama, line 83: I think it would make sense to suggest rate limiting and suggesting ways to query to limit bandwidth. 

Example language: "Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain, using rate limiting depending on the source of the request, and by publishing examples for queries that allow the agency to distinguish the audience of the request. "

schnuerle: Thanks @gene I made an update to the language incorporating what you said, and adding some more reasons that came up recently.

schnuerle:
See detailed notes, recording, slides, discussion from public [Working Group meeting](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2025.03.13-(MDS-Working-Group)) on March 13.  Leave your ideas and feedback here or inline with the changes as comments.

policy/README.md
```diff
@@ -57,35 +57,39 @@ The following information applies to all `policy` API endpoints.
...
 
-See the [Policy Examples](./examples/README.md) for ways Policy can be implemented.
+See the [MDS Policy Examples](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/MDS-Policy-Examples) wiki page for code examples of specific MDS Policy use cases, ideas on how Policy can be implemented.
 
 [Top][toc]
 
 ### Authorization
 
-The Policy endpoints should be made public. Authorization is not required.
+The Policy endpoints should be made public. Authorization is not required. Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain. Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain - this can be useful for rate limiting requests, ensure proper use, tracking access per requestor, and/or customization of the Policy tailored to the requestor.
```
TiphaineLAURENT, line 83: It seems the first sentence is repeated

policy/README.md
```diff
@@ -110,12 +114,15 @@ Policies shall be published by regulatory bodies or their authorized delegates a
 
 Policies typically refer to one or more associated geographies. Geographic information is obtained from the MDS [Geography](/geography) API.  Each policy and geography shall have a unique ID (UUID).
 
+Policies must be unique. A Policy may not have the exact same [Policy](#policy) object field values with a different `policy_id`. For example if the `mode_id`, `provider_ids`, `start_date`, `end_date`, and `rules` fields have the same values, that is the same Policy and must not be duplicated as two different policies in the Policy payload.
```
TiphaineLAURENT, line 117: I am not sure about this statement.

Policy may have the same listed fields values if a policy is updated thus replaced. That would mean 2 policies with possiblity all same values except description or currency and prev_policies but also different policy_id.

See https://github.com/openmobilityfoundation/mobility-data-specification/tree/main/policy#updating-or-ending-policies for reference

policy/README.md
```diff
@@ -159,6 +166,8 @@ _Query Parameters:_
...
 | `start_date` | [timestamp][ts] | Optional    | Beginning date of the queried time range; the default value is the request time |
 | `end_date`   | [timestamp][ts] | Optional    | Ending date of the queried time range; the default value is null, which captures all policies that are effective in the future|
+| `last_updated`   | Boolean | Optional    | If true, the endpoint only returns two fields: `version`, `last_updated`. Useful to quickly check when any data in the file has last been changed, without downloading the entire Policy payload. See the [Schema](#schema) section for an example. |
+| `active_only`   | Boolean | Optional    | If true, return only the current active and future policies, not the retired/previous policies. Any policy that is a prev_policies would not be returned. However, the array of prev_policies still will be returned for reference as part of any relevant active policy. Useful to reduce the Policy payload size for use cases where you do not need to know the previous policy details. |
```
TiphaineLAURENT, line 170: I would provide a default value of false in order to clarify the default behavior of fetching all declared policies

policy/README.md
schnuerle: Hi That's just showing what's changed. The first two sentences remain, but the new clarifying info is after that.

TiphaineLAURENT: Hi,
Yeah my formulation is misleading.
I am talking about the repeated "Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain." which is the first modified sentence.
Sorry about the confusion

schnuerle: Yes I see now. Fixing it.

policy/README.md
schnuerle: Maybe it should include all fields? But I don't think we want a situation where someone makes a second policy where everything is the same, except the `description` is "Description 2."

policy/README.md
schnuerle: Yes makes senses. I added default value of false if parameter is not provided.

policy/README.md
schnuerle: @TiphaineLAURENT Does [this change](https://github.com/openmobilityfoundation/mobility-data-specification/pull/928/commits/a3437521dfa7a0d94b62a565ef034c45c7963518) better reflect the concept and remain consistent with the rest of MDS? We will be talking about it at this week's [Working Group meeting](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2025.05.15-(MDS-Working-Group)).

Issue #873, opened 2023-08-04
Flag geographies that should be returned in the Reports endpoint


alexdemisch:
The Reports endpoint provides a useful way to aggregate information that otherwise shouldn’t be provided in a raw form.  This has been a helpful way for SFMTA to gather information about scooter trips taken by users on a low-income plan and trips on adaptive devices.  

To make the Reports endpoint data more useful, we added a set of neighborhoods to our Geographies endpoint that are helpful analytical zones for which we want to see aggregated data.  However, our Geographies endpoint also contains equity neighborhoods used for distribution requirements and a no parking zone.  We really don’t need aggregated data for these “operational” geographies, but the spec requires operators to aggregate data for all geographies anyway (even geographies that may no longer have any use because all data is immutable).

This can be overly burdensome for operators, and will only get more complex as more geographies are added, especially because the geographies are immutable.  Having a method to flag which geographies should be returned in a reports endpoint would be helpful.  I could see this going into Policy Requirements, Geographies, or Reports.  But Policy Requirements might make the most sense.

schnuerle:
Agree and this was a known complexity/limitation going into the launch of the Reports beta feature (we made [changes](https://github.com/openmobilityfoundation/mobility-data-specification/issues/672) and took it out of beta for 2.0). It's time we addressed this in the next release, and I also think Policy Requirements in the best place to enumerate the relevant geographies.

schnuerle:
Completed with #930

PR #930, opened 2025-04-08
Return only relevant geographies in Provider Reports

alexdemisch:
## Explain pull request

In response to Issue #873, adding an optional `geographies` field to endpoints in Policy Requirements to allow for a list of relevant geographies to that endpoint.   Issue #873 refers to adding relevant geographies to the reports endpoint only.  I drafted this PR so that it could apply to any endpoint, but not sure that's necessary.  If we want it to be more flexible, we may want to add it to Agency endpoints as well.  I haven't thought through potential use cases for this, though.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`


schnuerle:
I feel this should be a field called `report_geographies` to be clear, right? It just applies to Reports. Then the description should specify that the list of geographies here are the only geographies to return with Provider Reports.

alexdemisch:
Yes, that makes sense if the geographies array should only apply to the reports endpoint.  I wasn't sure whether or not to make it more flexible to potentially apply to other endpoints, but not sure why we would want to do that.

PR #931, opened 2025-04-08
Add service_start and service_end to Vehicle object in Provider and Agency APIs

### LINKED ISSUE #899: Add service_start_time and servict_end_time to VEHICLES endpoint
The only vehicle info in Trips/Events/Telemetry is DEVICE_ID - we need to join them with Vehicles to get more vehicle info e.g. PROPULSION_TYPE, ACCESSIBILITY_ATTRIBUTES. There is no datetime fields currenty to indicate when a device joins the retires from the fleet, which is critical in the cases of a device gets a new VEHICLE_ID (plate number) and stays in the fleet, or a device gets upgrades to a adaptive device, which will results in two records with same DEVICE_ID. 

Joining Trips/Events/Telemetry with Vehicles on DEVICE_ID will not get us the right vehicle attributes.

Add service_start_time and servict_end_time to the VEHICLES endpoint. 

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `provider`

Mu-yi-Zhou:
## Explain pull request

In response to Issue https://github.com/openmobilityfoundation/mobility-data-specification/issues/899, adding a required field service_start and required if available field service_end to Vehicle endpoint of Provider API.

* No, not breaking

## Impacted Spec

* `provider`

### DIFF
--- a/data-types.md
+++ b/data-types.md
@@ -34,6 +34,9 @@ A vehicle record is as follows:
 | `battery_capacity`   | Integer  | Required if Available | Capacity of battery expressed as milliamp hours (mAh) |
 | `fuel_capacity`      | Integer  | Required if Available | Capacity of fuel tank (liquid, solid, gaseous) expressed in liters |
 | `maximum_speed`      | Integer  | Required if Available | Maximum speed (kph) possible with vehicle under normal, flat incline, smooth surface conditions. Applicable if the device has a built-in or intelligent speed limiter/governor. |
+|`service_start`       |Timestamp | Required              | Date/time the vehicle starts providing service |
+|`service_end`         |Timestamp | Required if Available | Date/time the vehicle stops providing service. Required if the vehicle is retired |
+
 
 [Top][toc]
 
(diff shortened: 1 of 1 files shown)


schnuerle:
What about the timestamp you were asking about [here](https://github.com/openmobilityfoundation/mobility-data-specification/issues/899#issuecomment-2239799322)?

And should these fields be called `lifecycle_start` and `lifecycle_end` to more accurately describe their use?

schnuerle:
At the public working group meeting, we discussed this at length. Here are some of the key takeaways. See [notes](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2025.04.10-(MDS-Working-Group)) for full details and recording.

**Key Points**

- MDS Vehicles is a real time status of vehicles, not a history
- **Vehicle property change**: when part of a vehicle is changed: accessibility, batter, tires, IoT device, license plate, color, etc (custom for each Mode). Useful for knowing use and rides with certain equipment. This is really what SFMTA wants to see (going to rename Issue/PR). Is this history a new data structure in MDS? Or a new event type in Events with descriptive text?
   - Vehicle lifecycle: when a vehicle was added to service, then finally retired. Useful for carbon footprint discussions, lifetime operations, green goals, etc. MDS Events already has decommissioned state for when its removed from service.
   - Vehicle maintenance: When vehicles have maintenance done on or off street - already covered in MDS with vehicle states in Events. Last inspection date is in Vehicles now. Some public agencies get this in monthly reports, but MDS should be able to handle this instead.
- Need to hear move from operators, vendors, data aggregators. Comment on GitHub now with challenges or issues.

PR #943, opened 2025-07-04
Admin patch 2.0.2 dev

schnuerle:
Cleanup dev from patch


schnuerle (commented):
Moving patch details to dev.

PR #944, opened 2025-07-04
Add Incidents 

### LINKED ISSUE #613: [Provider API] - Include crash data as part of the `/status_changes` endpoint, `event_type` = `accident`

Several cities are inquiring about crash data. We have been supporting them in linking this data with street usage and existing infrastructure. Crash data are today issued from manual reporting at hospitals. It provides timestamp, location and type of vehicles. This data however is not of high quality, and it is difficult to make meaningful statistics out of it and hence support decision-making on infrastructure or policies. 

I believe it could be valuable to add this to the `/status_changes` endpoint. For example, I would see a transition from `on_trip` to `unavailable` with `event_type` = `accident/crash`. 
I don't know to which extent providers can obtain this information "live". Clearly a survey just after a crash is difficult. Some providers have reporting process through the website. I wonder whether the (next generation) vehicles are (will be) able to detect crash with accelerometer & speed data. A good question to operators. 
Ideally this would be part of the `/status_changes` endpoint. However we could also imagine a separate reporting but it may add complexity.   

Today, this would be very useful as historical data to get analytics on crash for shared micro-mobility and link it to street segment. However, tomorrow, in a broader context, I believe this is a great step for connected vehicles to report 'live' crash to roadway authorities and emergency services. 

### Is this a breaking change

I don't believe this is breaking. 

### Impacted Spec

Provider

An alternative could again be separate monthly reporting in csv by providers where we would see the number of crashes. However this would be left to goodwill reporting by users & providers.

schnuerle:
## Explain pull request

Creation of an incidents object and new incident endpoints in Provider and Agency to allow reporting of various incidents through MDS. 

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `agency`
* `provider`

## Additional context

Provides a solution for issue #613.

### DIFF
--- a/data-types.md
+++ b/data-types.md
@@ -249,7 +249,7 @@ A Trip is defined by the following structure:
 
 ## Incidents
 
- Incidents are used in both [Provider](/provider#stops) and [Agency](/agency#stops) telemetry data, whether on or off a Trip. 
+ Incidents are used in both [Provider](/provider#incidents) and [Agency](/agency#incidents) telemetry data, whether on or off a Trip. 
 
 | Field              | Type            | Required/Optional      | Comments |
 | ----               | ----            | ----                   | ----     |
--- a/data-types.md
+++ b/data-types.md
@@ -144,7 +144,7 @@ A standard point of vehicle telemetry. References to latitude and longitude impl
 | `fuel_percent`    | Integer         | [Required if Applicable](./general-information.md#required-if-applicable-fields)  | Percent fuel in vehicle, expressed between 0 and 100 |
 | `tipped_over`     | Boolean         | Required if Known      | If detectable and known, is the device tipped over or not? Default is 'false'. |
 | `gtfs_stop_id` | String | [Optional](./general-information.md#optional-fields) | A unique stop ID to be recorded when a vehicle makes a stop event at a location. Matches [GTFS](https://gtfs.org/documentation/schedule/reference/) `stop_id` |
-| `incident_ids`    | Array of UUIDs  | Optional               | Array of one or more [Incident](#incident) IDs that are connected to this telemetry data point. |
+| `incident_ids`    | UUID[]          | Optional               | Array of one or more [Incident](#incidents) IDs that are connected to this telemetry data point. |
 
 ### GPS Data
 
--- a/data-types.md
+++ b/data-types.md
@@ -15,6 +15,7 @@ This MDS data types page catalogs the objects (fields, types, requirements, desc
 - [Stops](#stops)
   - [Stop Status](#stop-status)
 - [Trips](#trips)
+- [Incidents](#incidents)
 - [Reports](#reports)
 
 ## Vehicles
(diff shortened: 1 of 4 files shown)


data-types.md
```diff
@@ -245,6 +246,25 @@ A Trip is defined by the following structure:
...
+| `description`      | String          | Optional               | Text description of the incident. |
+| `severity`         | String          | Optional               | Description of the severity of the incident. |
+| `report_id`        | String          | Optional               | Identifier of an external report, from a police report, citation, internal system, service request, etc. The report source is communicated by the operator to the agency outside of MDS. |
+| `report_type`      | String          | Optional               | Description of the type of report referenced by the `report_id`, eg. police, customer, remote operator, 311 call, etc. |
+| `contact_info`     | String          | Optional               | Description of any relevant contact information about the incident the operator can provide. |
+| `preliminary`      | Boolean         | Optional               | If `true`, then this information in this Incident is only preliminary, with more details and/or validation coming at a later date. If 'false`, the information provided here is deemed valed with no more updates expected. |
+
```
schnuerle, line 280: We may want to add more fields here, per this [comment](https://github.com/openmobilityfoundation/mobility-data-specification/issues/613#issuecomment-2942595074).

data-types.md
```diff
@@ -143,6 +144,7 @@ A standard point of vehicle telemetry. References to latitude and longitude impl
 | `fuel_percent`    | Integer         | [Required if Applicable](./general-information.md#required-if-applicable-fields)  | Percent fuel in vehicle, expressed between 0 and 100 |
 | `tipped_over`     | Boolean         | Required if Known      | If detectable and known, is the device tipped over or not? Default is 'false'. |
 | `gtfs_stop_id` | String | [Optional](./general-information.md#optional-fields) | A unique stop ID to be recorded when a vehicle makes a stop event at a location. Matches [GTFS](https://gtfs.org/documentation/schedule/reference/) `stop_id` |
+| `incident_ids`    | UUID[]          | Optional               | Array of one or more [Incident](#incidents) IDs that are connected to this telemetry data point. |
```
schnuerle, line 149: Do we need to add this same field to Events? I don't think so because Telemetry captures all timestamps and locations of vehicles. 

I also don't think we need any 'incident' type of events or event types, or vehicle states. Since incidents can be lots of different types of things, not all are applicable to event types or vehicle states.

schnuerle:
I think this proposal meets DDOT's requirements as Stephanie laid out in a [comment](https://github.com/openmobilityfoundation/mobility-data-specification/issues/613#issuecomment-2884535092), for the 12 hour notification window.

We could add more crash specific info in MDS. But I'm inclined to keep it simple like this, and instead reference an external report for more info, using the report_id provided here. That way MDS gives you basic notification of details of an incident or preliminary incident data, without trying to capture all data for all types of incidents, crashes, etc, or be the authoritative source of truth. There is an [inline comment](https://github.com/openmobilityfoundation/mobility-data-specification/pull/944#discussion_r2186237172) about this in the PR.

There is also a comment about if we need this in Events too. [Please read](https://github.com/openmobilityfoundation/mobility-data-specification/pull/944#discussion_r2187308502) and reply.

We could also capture more data in Telemetry GPS points for this, like accelerometer data that Cowboy captures and sends, per this [comment](https://github.com/openmobilityfoundation/mobility-data-specification/issues/613#issuecomment-3037452754).

Another question I have: what is the frequency of Telemetry updates? I don't see where MDS specifies this. Those of you working on the ground as vendors consuming MDS or operators sending MDS Telemetry, how often are these data pings?

schnuerle:
Based on the latest [working group meeting action items](https://github.com/openmobilityfoundation/mobility-data-specification/wiki/Web-conference-notes,-2025.07.10-(MDS-Working-Group)#action-items-and-decisions), this proposal has been updated:

- Add incident types: Disengagements, Theft ([done](https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-incident/data-types.md#incidents))
- Add yes/no fields: ambulance transport, medical dispatch ([done](https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-incident/data-types.md#incidents))
- Add to telemetry as optional: accelerometer data ([done](https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-incident/data-types.md#gps-data))
- Open question: how often should telemetry be updated/sent? ([updated](https://github.com/openmobilityfoundation/mobility-data-specification/blob/feature-incident/provider/README.md#telemetry))

Should there be other incident types? Traffic violation was brought up, where an operator knows that a violation occurred (and might have a good reason as to why).

data-types.md
schnuerle: Agreed in WG meeting that it is not needed in Events.

data-types.md
schnuerle: Added some more fields based on WG meeting.

schnuerle:
Besides other feedback, and feedback on incident types/violations (Should there be other incident types? Traffic violation was brought up), this may be ready for dev soon.

schnuerle (commented):
Possibility for other incident types (like Traffic violation) in a future PR, but for now this is ready for dev based on all the discussions.

PR #949, opened 2025-07-30
Update Policy Authorization description to allow authentication

### LINKED ISSUE #915: Clarify that Policy API can require Authorzation to access the published material

We could clarify the fact that Policy API would sometimes require Authorization to access the policies/geographies published on it.

When looking at the documentation of the Policy API, in the Authorization section, we could change the language there to say that the publication with our without Authorization of polices on the Policy API is at-will for the publishing user.

### Is this a breaking change

 * No, not breaking

### Impacted Spec

 * `policy`

N/A

### Additional context

We have on multiple occasions at Blue Systems had to publish the policies with Authorized access only to implement some of the use-cases requested by our client cities. So we know this is a minor change for the operators, since they would have credentials for other APIs anyway.

pierre-bouffort:
## Explain pull request

In the Policy API section, this PR updates the Authorization description to allow for authentication when use-cases justify it. 

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `policy`

## Additional context

Relates to #915

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -151,9 +151,9 @@ See the [Responses section][responses] for information on valid MDS response cod
 
 **Endpoint**: `/policies/{policy_id}`  
 **Method**: `GET`  
-**Schema:** See [`mds-openapi`](https://github.com/openmobilityfoundation/mds-openapi) repository for schema.    
-**Authorization**: public _or_ authenticated (see [Authorization](#authorization)) 
-**`data` Payload**: `{ "policies": [] }`, an array of objects with the structure [outlined below](#policy).
+**Schema:** See [`mds-openapi`](https://github.com/openmobilityfoundation/mds-openapi) repository for schema.  
+**Authorization**: public _or_ authenticated (see [Authorization](#authorization))  
+**`data` Payload**: `{ "policies": [] }`, an array of objects with the structure [outlined below](#policy).  
 
 _Path Parameters:_
 
@@ -200,8 +200,8 @@ See [Responses][responses], [Bulk Responses][bulk-responses], and [schema][schem
 
 **Endpoint**: `/requirements/`  
 **Method**: `GET`  
-**[Beta feature](/general-information.md#beta-features)**: *No (as of 2.0.0)*. 
-**Authorization**: public
+**[Beta feature](/general-information.md#beta-features)**: *No (as of 2.0.0)*.  
+**Authorization**: public  
 **Schema:** See [`mds-openapi`](https://github.com/openmobilityfoundation/mds-openapi) repository for schema.  
 **`data` Payload**: `{ requirements: [] }`, JSON objects that follow the schema [outlined here](#requirement).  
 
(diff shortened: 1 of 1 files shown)


policy/README.md
```diff
@@ -80,7 +80,9 @@ See the [MDS Policy Examples](https://github.com/openmobilityfoundation/mobility
 
 ### Authorization
 
-The Policy endpoints should be made public. Authorization is not required. Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain - this can be useful for rate limiting requests, ensure proper use, tracking access per requestor, and/or customization of the Policy tailored to the requestor.
+In most cases, the Policy endpoints should be made public. Authorization is not required in such cases. Agencies may make reasonable accommodations to manage their endpoints, for example, using an API key that has a clear, public way to obtain - this can be useful for rate limiting requests, ensure proper use, tracking access per requestor, and/or customization of the Policy tailored to the requestor.
+
+In some cases though, it can be justified to use Authorization for the Policy API (some agencies may decide to make it authenticated for privacy programs or functional purposes). Authorization may then be used for the Policy API. It should then rely on the standard [Authorization](../general-information.md#authorization) methods used in other MDS APIs.
```
schnuerle, line 87: FYI [draft language here for what is in CDS](https://github.com/openmobilityfoundation/curb-data-specification/tree/dev/curbs#authorization):

[Authorization](https://github.com/openmobilityfoundation/curb-data-specification/blob/dev/general-information.md#authorization) is not required for any of the Curbs endpoints, as this information should be made public and easily accessible. Free to acquire API keys or authorization may be utilized by the public agency to track usage, prevent abuse, or for certain mobility programs or pilots.

schnuerle:
Adding @geneorama's detailed and valid points from the Issue here:

Capturing some arguments against allowing authorization:

- Providing a provision for authorization risks casting it as "should" have authorization
- Adding authorization adds another layer of maintenance
- Adding authorization adds a barrier for would be civic developers in the public, who (assuming there is a process for obtaining keys) are faced with navigating a system that is opaque and bureaucratic, and may not be supported across administrations

If it's used during development a temporary measure, it is exceptionally risky that the service will never be open:

- A service can't be properly tested in development without authentication because it would require significant changes to the changing the code base
- Funding will almost certainly be reluctant to cover development work to remove authentication once a project is complete
- Fundamentally if a service is to be open, then it needs to be designed and tested in the open form.

There ways to address many, if not all, of the justifications for authentication

- Use IP whitelisting for pilots
- Require vendor identification for API calls
- Provide and require use private codes for requests to avoid spoofing, if that is a concern
- Employ security and load balancing systems for the open API just as you would any other open API

schnuerle (approved):
Approving based on majority of positive feedback and use of some API keys and authorization in the real world now for specific programs. If this is abused and MDS Policy feeds that by most account should be public are put behind an arduous key process or blanket authorization, we may reevaluate in the next release and make appropriate changes.

PR #955, opened 2025-09-19
Clarify relationship between /vehicles and /vehicles/status

jiffyclub:
## Explain pull request

I'm trying to make it explicit that when calling the /vehicles endpoint without specifying a vehicle it must include all vehicles returned by calling the /vehicles/status at the same time without specifying a vehicle. (It may also include additional vehicles as appropriate, but the response must include all vehicles from /vehicles/status.) I think the existing documentation says this implicitly, but I've encountered some confusion in the wild and would like to make it explicit.

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `provider`

## Additional context

None.


schnuerle (approved):
Thank you for this clarification

PR #982, opened 2026-06-02
Clarify pagination requirements and recommended page size (#981)

### LINKED ISSUE #981: Mimimum page size for pagination /vehicles and /vehicles/status endpoint

Currently, providers can choose their own page size for the `/vehicles` and `/vehicles/status` endpoints.

In practice, consumers usually need the complete vehicle list and status data for dashboards, monitoring, and synchronization. Very small page sizes create unnecessary API load. For example, with 4,000 vehicles and a page size of 20, a consumer must make:

`2 * (4000 / 20) = 400 API calls`

every polling interval. If polling every 30 seconds, this creates excessive traffic for both consumers and providers.

Either:

1. Remove pagination from `/vehicles` and `/vehicles/status`, since most use cases require the full dataset anyway.
2. Or define a minimum page size, for example `1000`, and preferably allow consumers to request a page size explicitly via a query parameter.

This would reduce unnecessary API calls while still allowing pagination for very large datasets if needed.

### Is this a breaking change?

Yes, for providers that currently use very small page sizes.

However, the expected implementation impact should be limited. In most cases, providers would only need to increase the configured page size or support a consumer-defined page size parameter.

### Impacted Spec

`provider`

Keeping pagination as-is is possible, but it allows inefficient implementations that put unnecessary load on both consumers and providers.

Another option is to add guidance instead of a hard requirement, but that may not solve the issue if providers continue using very small page sizes.

zeulus:
## Explain pull request

Moves the core pagination details into `general-information.md` and has the per-API specs link to that shared section. Documents the JSON:API `page[number]`/`page[size]` query parameters (previously undocumented in MDS) and recommends a minimum supported `page[size]` of 500 and a default of 1000, leaving the maximum for each provider/agency to document. The `links` keys and example move out of `provider/README.md` into the shared section, and Policy pagination is noted as optional.

## Is this a breaking change

* No, not breaking (all guidance is recommended, not required)

## Impacted Spec

* `general`
* `agency`
* `provider`
* `policy`

## Additional context

See #981

### DIFF
--- a/agency/README.md
+++ b/agency/README.md
@@ -161,7 +161,7 @@ _Path Parameters:_
 | ------------ | ---- | ----------------- | ------------------------------------------- |
 | `device_id`  | UUID | Optional          | If provided, retrieve the specified vehicle |
 
-If `device_id` is specified, `GET` will return an array with a single vehicle record, otherwise it will be a list of vehicle records with pagination details per the [JSON API](https://jsonapi.org/format/#fetching-pagination) spec:
+If `device_id` is specified, `GET` will return an array with a single vehicle record, otherwise it will be a list of vehicle records with pagination details (see [Pagination][pagination]):
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


jiffyclub:
Looks good to me!

schnuerle (approved):

schnuerle:
@zeulus can you click this and sign the SLA so this work can be accepted in the release? https://cla-assistant.io/openmobilityfoundation/mobility-data-specification?pullRequest=979

PR #983, opened 2026-06-10
Update providers.csv with Flexigo

MarcusVerdoza:
name: MarcusVerdoza
about: Suggest changes to MDS
title: Update providers.csv with Flexigo

---

## Explain pull request

Adding a new provider - Flexigo

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `provider`


schnuerle (approved):
Thank you for the update!

PR #989, opened 2026-08-12
Update providers.csv with Storer Transit Systems

MarcusVerdoza:
name: MarcusVerdoza
about: Suggest changes to MDS
title: Update providers.csv with Storer Transit Systems

---

## Explain pull request

Adding a new provider - Storer Transit Systems

## Is this a breaking change

* No, not breaking

## Impacted Spec

* `provider`


schnuerle:
Approved and ready for use now!

It will go into the next MDS version when we create that in the future. For now it's in the dev branch.

PR #992, opened 2026-09-15
fix: correct `system_start` typo to `service_start` in event types table

vbourgeois90-blue:
## Explanation

`modes/event_types.md` lists `system_start` ("Start of service"), but this value is not used by any mode state machine and is not present in the normative schema.

The value used everywhere else is `service_start`:

- **State machines:** `modes/car-share.md`, `modes/passenger-services.md` and `modes/delivery-robots.md` all list `service_start` and pair it with `service_end`.
- **Schema:** [`mds-openapi`](https://github.com/openmobilityfoundation/mds-openapi) `models/data-types/event-type.yaml` enumerates `service_end` and `service_start`, and has no `system_start`. Per-mode event schemas (e.g. `models/modes/car-share/event.yaml`) also use `service_start`.

This appears to date from a00eae6 ("chore: add missing event types"), which added `service_end` and `system_start` together — the second of the pair was mistyped.

Practical impact: implementers reading the event types table as the canonical list may emit `system_start`, which then fails validation against the published OpenAPI schema. We have seen providers send both values in the wild.

## Is this a breaking change?

No. Documentation-only correction; `system_start` was never a valid schema value.

## Impacted Spec

- `modes/event_types.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


schnuerle:
Thank you @vbourgeois90-blue !  Make sure to sign the CLA so we can accept your change:

https://cla-assistant.io/openmobilityfoundation/mobility-data-specification?pullRequest=992

